PR #17, opened 2023-08-22
Add help target

andrewg-xyz:


andrewg-xyz:
Just a suggestion, I've found the help target to be immensely helpful and think it'd be good to be used by default in all projects.

UncleGedd (approved):
@andrewg-xyz you glorious 🦄 , thanks for the contribution!

PR #19, opened 2023-08-23
Adds global vars to bundle

UncleGedd:


UncleGedd:
very stale PR, will re-attack in the future if it comes up again

PR #26, opened 2023-08-27
distinguish between zarf config and uds config

UncleGedd:


src/cmd/bundle.go
```diff
@@ -189,3 +207,38 @@ func init() {
...
 }
+
+// configureZarf copies configs from UDS-CLI to Zarf
+func configureZarf() {
+	zarfConfig.CommonOptions.Insecure = config.CommonOptions.Insecure
+	zarfConfig.CommonOptions.TempDirectory = config.CommonOptions.TempDirectory
+	zarfConfig.CommonOptions.OCIConcurrency = config.CommonOptions.OCIConcurrency
+	zarfConfig.CommonOptions.Confirm = config.CommonOptions.Confirm
+	zarfConfig.CommonOptions.CachePath = config.CommonOptions.CachePath
+}
```
Noxsios, line 218: I would make a PR on Zarf to add a `config.SetCommonOptions( opts config.CommonOptions )` function so that this operation is more strongly typed and options are directly tied 1:1 (as it stands someone could accidentally change a line to be like: `zarfConfig.CommonOptions.Confirm = config.CommonOptions.Insecure`)

UncleGedd: Love it, this was on my list to chat with you all about

PR #50, opened 2023-09-07
adds publish cmd

UncleGedd:
- Adds publish cmd to allow publishing local bundles
- Fixes some bugs/discrepancies between local and remote bundles

### DIFF
--- a/src/pkg/bundle/remote.go
+++ b/src/pkg/bundle/remote.go
@@ -7,6 +7,7 @@ package bundle
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"github.com/defenseunicorns/uds-cli/src/config"
 	"github.com/defenseunicorns/uds-cli/src/pkg/utils"
@@ -191,6 +192,7 @@ func (op *ociProvider) LoadBundle(concurrency int) (PathMap, error) {
 		return nil, err
 	}
 
+	spinner := message.NewProgressSpinner("Starting bundle pull")
 	loaded, err := op.LoadBundleMetadata() // todo: remove? this seems redundant, can we pass the "loaded" var in
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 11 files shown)


src/pkg/bundle/remote.go
```diff
@@ -191,6 +193,9 @@ func (op *ociProvider) LoadBundle(concurrency int) (PathMap, error) {
 		return nil, err
 	}
 
+	animatedText := pterm.DefaultSpinner
```
corang, line 196: Why aren't we using the `message` library?

corang (changes requested):

src/pkg/bundle/remote.go
UncleGedd: We actually want to move away from pterm and the Zarf TUI entirely. Right now a lot of TUI things are tightly coupled to the Zarf library so we can't. This was just me attempting some pterm though, happy to see if it's easier/cleaner to use the Zarf `message` library

corang: Why do we want to move away from zarf tui?

UncleGedd: Happy to give more context. UDS CLI as a product is distinct from Zarf. We should look at what worked well and what didn't and take those as lessons learned. In this case, @jeff-mccoy made the case that pterm was tough to work at a larger scale, and a more MVC-like or event-driven TUI like [BubbleTea](https://github.com/charmbracelet/bubbletea) is likely a better choice. 

When we actually have the ability to separate TUI things out of Zarf libs, I'll write an ADR for going one way or the other

corang: Gotcha, I think I'd make the argument that using the message library for now is better than pterm since as of now we still have to use it so it makes for better consistency and the tech debt stays the same if we're not going to stay with pterm or the message library long term.

corang: Also this is starting to feel like we should develop a separate tui library for all of defense unicorns to use for go terminal apps instead of redoing it "better" every time we make a new one.

UncleGedd: Enforcing lib usage, etc is tricky. Gotta balance freedom with opinionation. I'm all for a similar look and feel, but if we can get there with less complexity then I lean toward that

Nevertheless, updated to use `message` Zarf lib

corang (approved):

PR #64, opened 2023-09-14
Confirm when OptionalComponents are specified

### LINKED ISSUE #32: A user should never be able to select which components get deployed from a package in a bundle when `optional-components` is declared
Error log:
```
                                                                                                      
  📦 LOGGING COMPONENT                                                                                
                                                                                                      

     ERROR:  Failed to deploy bundle: unable to deploy all components in this Zarf Package: unable to deploy
             component logging: unable to deploy component logging: unable to push images to the registry: open
             /tmp/zarf-2357824383/images/blobs/sha256/0abe676a9b183901b27c89d50397611e820be40583fe02ef94e148fe33956f00:
             no such file or directory
```
Bundle spec:
```
zarf-packages:
  # Zarf init
  - name: init
    repository: ghcr.io/defenseunicorns/packages/init
    ref: v0.29.1
    optional-components:
      - git-server
```
Deploy without confirm and select logging as a component you want.

Bundles are meant to be declaritive, by allowing user choice of components in packages during deploy time you lose that declaritiveness. I think a user should never be allowed to select components but I'll settle for not allowing when optional-components is specified.

kirankt:
Added automatic confirmation when OptionalComponents are specified. 

Also cleaned up some of the import formatting of code in the same directory. Its desirable to first import std libs followed by external ones.

Closes #32


UncleGedd:
> One issue that still persists, is that when optional components are not specified, then the user will still be prompted to choose componentst that haven't been included in the bundle. Methinks, its probably a good idea to always confirm and not give a choice.

I agree @kirankt . I think that's in line with @corang 's thinking as well

UncleGedd:
Thanks @kirankt ! Appreciate you cleaning up the imports as well, need to automate that on my machine

UncleGedd (approved):

Issue #77, opened 2023-10-12
chore: setup OpenSSF scorecard


Racer159:
This repo should have OpenSSF scorecard setup on it to recommend best practices for security settings.

https://github.com/defenseunicorns/zarf/blob/main/.github/workflows/scorecard.yaml
https://github.com/defenseunicorns/zarf-ui/blob/main/.github/workflows/scorecard.yaml

PR #78, opened 2023-10-12
Add a scorecard.yaml workflow

Racer159:
Fixes: #77 

This adds a scorecard workflow to generate recommendations for security best practices.

:warning: As is, this is setup to use a scorecard read token named `SCORECARD_READ_TOKEN` (which is optional but gives more info).  See more here: https://github.com/ossf/scorecard-action#authentication-with-fine-grained-pat-optional


UncleGedd (approved):
thanks @Racer159 , really appreciate it!

EDIT: just noticed the E2E tests failing, going to re-run them to see if that was a fluke

UncleGedd (approved):

PR #79, opened 2023-10-12
update go version for ci

corang:


UncleGedd:
Hey thanks for this @corang , could you also make the update in `go.mod` (and any other references to that old version of Go that I'm not thinking of)?

corang:
tbh I don't really know how to do that, is it just updating the value in the `go.mod`?

UncleGedd:
> tbh I don't really know how to do that, is it just updating the value in the `go.mod`?

Yep exactly, just bump the version in the `go.mod` and run `go mod tidy` (not sure if the `go mod tidy` will actually change anything but we'll all need to do it anyway)

UncleGedd (approved):
thank you!!

PR #83, opened 2023-10-13
feature: setup renovate

TristanHoladay:


TristanHoladay:
basic setup. will need a separate PR for fine tuning based on feedback as well as for attempting to handle updating version comments next to digests, which Renovate finds automatically.

UncleGedd (approved):
Thanks @TristanHoladay ! Agreed we should further refine the config, I'll make a follow-up ticket

PR #97, opened 2023-10-13
Restore OCI concurrency

UncleGedd:
- Restores OCI concurrency by refactoring "pull" and "publish" bundle ops to use `oras.Copy()`
- Edits imports to have goimports style syntax

### DIFF
--- a/src/pkg/bundle/remote.go
+++ b/src/pkg/bundle/remote.go
@@ -12,18 +12,21 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
-	"github.com/defenseunicorns/uds-cli/src/config"
-	"github.com/defenseunicorns/uds-cli/src/pkg/utils"
-	"github.com/defenseunicorns/uds-cli/src/types"
 	"github.com/defenseunicorns/zarf/src/pkg/message"
 	"github.com/defenseunicorns/zarf/src/pkg/oci"
 	zarfUtils "github.com/defenseunicorns/zarf/src/pkg/utils"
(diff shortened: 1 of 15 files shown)


src/pkg/bundle/remote.go
```diff
@@ -252,35 +258,29 @@ func (op *ociProvider) LoadBundle(_ int) (PathMap, error) {
...
+	}
+
+	// Create a thread to update a progress bar as we save the package to disk
+	doneSaving := make(chan int)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go zarfUtils.RenderProgressBarForLocalDirWrite(op.dst, estimatedBytes, &wg, doneSaving, fmt.Sprintf("Pulling bundle: %s", bundle.Metadata.Name))
+	_, err = oras.Copy(op.ctx, op.Repo(), op.Repo().Reference.String(), store, op.Repo().Reference.String(), copyOpts)
+	if err != nil {
+		return nil, err
```
mikevanhemert, line 274: I am maybe a 2/10 in Golang these days, but will this leak goroutines when there are errors?

mikevanhemert (approved):
Looks cleaner, happy to see concurrency back! I asked one clarification question, but it can be visited later if my spidey senses are right.

src/pkg/bundle/remote.go
UncleGedd: Good call! I'll add the trigger to stop that goroutine if an error is detected

mikevanhemert (approved):

decleaver (approved):
lgtm

Issue #115, opened 2023-10-18
Use OCI concurrency and caching where possible


UncleGedd:
Caching and OCI concurrency have recently been implemented for `pull` and `publish` ops, this enhancement also needs to be implemented for bundle creation and any other places where we are pulling from OCI

UncleGedd:
[This PR](https://github.com/defenseunicorns/uds-cli/pull/137) implements OCI concurrency + caching when pulling from OCI during bundle deploy and create

UncleGedd:
Still need:
- Caching when pulling in local pkgs during bundle create
- Caching when pulling in local pkgs during bundle deploy
- Caching + OCI concurrency during bundle pull for both local and remote

UncleGedd:
completed with https://github.com/defenseunicorns/uds-cli/pull/751

PR #123, opened 2023-10-19
update Zarf to 0.30.x

UncleGedd:
Upgrades Zarf lib to the latest 0.30.x using the new packager API

Noting that this PR points to the latest commit in [this Zarf branch](https://github.com/defenseunicorns/zarf/pull/2087) to get around the [Docker refs issue](https://github.com/defenseunicorns/zarf/issues/2086). Will likely wait until that Zarf PR gets merged and Zarf cuts a new release to merge this UDS CLI PR (cc @Noxsios @Racer159 )

EDIT: merging!


src/pkg/sources/remote.go
```diff
@@ -0,0 +1,190 @@
...
+			}
+			if err := os.WriteFile(filepath.Join(dst.Base, config.ZarfYAML), zarfYamlBytes, 0644); err != nil {
+				return err
+			}
+			break
+		}
+	}
+
+	return nil
+}
```
Noxsios, line 126: This can be simplified to something like:

```
root, err := r.Remote.FetchRoot()
...
manifestDesc := root.Locate(r.PkgManifestSha)
...
if oci.IsEmptyDescriptor(manifestDesc) // handle
...
manifest, err := r.remote.FetchManifest(manifestDesc)
...
pkg, err := oci.FetchZarfYaml(manifest)
...
if err := utils.WriteYaml(pkg) // dunno if this exists 
...
```

src/pkg/sources/remote.go
```diff
@@ -0,0 +1,190 @@
...
+
+// Collect doesn't need to be implemented
+func (r *RemoteBundle) Collect(_ string) (string, error) {
+	return "", nil
```
Noxsios, line 130: if the method is not implemented, it is best to return an `fmt.Errorf("not implemented in %T", r)` so that if this is fed into `packager` and is tried to be called, it will not fail silently

src/pkg/sources/remote.go
```diff
@@ -0,0 +1,190 @@
...
+		lb, err := r.Remote.Repo().Fetch(ctx, layer)
+		if err != nil {
+			return err
+		}
+
+		err = store.Push(ctx, layer, lb)
+		if err != nil {
+			return err
+		}
+	}
```
Noxsios, line 186: this has no concurrency and should use `oras.Copy` if methods in `oci.NewOrasRemote` are insufficient

src/pkg/sources/remote.go
```diff
@@ -0,0 +1,190 @@
...
+
+// Collect doesn't need to be implemented
+func (r *RemoteBundle) Collect(_ string) (string, error) {
+	return "", nil
+}
+
+// downloadPkgFromRemoteBundle downloads a Zarf package from a remote bundle
+func (r *RemoteBundle) downloadPkgFromRemoteBundle() error {
+	// todo: use oras.Copy for faster downloads
+	ctx := context.TODO()
```
Noxsios, line 136: context should be initialized right before it is used, not at the top of a function to ensure it is not mutated before being used

src/pkg/sources/tarball.go
```diff
@@ -0,0 +1,243 @@
...
+		return err
+	}
+
+	err = sourceArchive.Close()
+	return err
+}
+
+// Collect doesn't need to be implemented
+func (t *TarballBundle) Collect(_ string) (string, error) {
+	return "", nil
```
Noxsios, line 163: return an error

src/pkg/sources/tarball.go
```diff
@@ -0,0 +1,243 @@
...
+		written, err := io.Copy(target, stream)
+		if err != nil {
+			return err
+		}
+		if written != size {
+			return fmt.Errorf("expected to write %d bytes to %s, wrote %d", size, path, written)
+		}
+
+		return nil
+	}
```
Noxsios, line 227: just like how `sources.TarballSource` works, this is where a filelist should be built, then later fed into `dst.SetFromPaths` as this is the true source of where a file placed on disk is from. Extracting each file by name, then doing a `filepath.Walk` to get the names again is a wasted operation.

src/test/e2e/bundle_test.go
```diff
@@ -30,7 +31,7 @@ func zarfPublish(t *testing.T, path string, reg string) {
 	require.NoError(t, err)
 }
 
-const zarfVersion = "v0.29.1"
+const zarfVersion = "v0.30.1"
```
Noxsios, line 34: this should prob be an env var that gets passed into `go test`

src/test/packages/01-uds-bundle/uds-bundle.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
 zarf-packages:
   - name: init
     repository: localhost:888/init
-    ref: v0.29.1
+    ref: v0.30.1
```
Noxsios, line 10: not to beat a dead horse, but this is why bundle YAML templating needs to be a feature, as this is now 3+ places that need to be updated all at the same time. There may be a way to configure renovate to do this.

src/pkg/sources/remote.go
```diff
@@ -0,0 +1,190 @@
...
+	})
+	if err != nil {
+		return err
+	}
+
+	if err = utils.ReadYaml(dst.ZarfYAML, &pkg); err != nil {
+		return err
+	}
+
+	dst.SetFromPaths(files)
```
Noxsios, line 65: this should be using `dst.SetFromLayers`, not `dst.SetFromPaths`

src/pkg/sources/remote.go
Noxsios, line 1: neither `sources.ValidatePackageIntegrity` nor `sources.ValidatePackageSignature` are called in these sources, which means that bundle is going to lose out on integrity+signature validation. If so I would say an ADR might be in order for what bundle will replace integrity+signature checks with.

Noxsios: This also means the `public-key` field in bundle's YAML is unused

src/pkg/bundle/common.go
```diff
@@ -181,9 +182,6 @@ func (b *Bundler) ValidateBundleResources(bundle *types.UDSBundle, spinner *mess
 			publicKeyPath = ""
 		}
 
-		if err := packager.ValidatePackageSignature(tmp, publicKeyPath); err != nil {
-			return err
-		}
```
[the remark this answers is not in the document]
Noxsios: this function now writes the `public-key`, but does not do anything with it

Noxsios (changes requested):
changes requested

src/pkg/bundle/common.go
```diff
@@ -71,7 +71,7 @@ func NewOrDie(cfg *types.BundlerConfig) *Bundler {
 // ClearPaths clears out the paths used by Bundler
 func (b *Bundler) ClearPaths() {
 	_ = os.RemoveAll(b.tmp)
-	_ = os.RemoveAll(zarfConfig.ZarfSBOMDir)
+	//_ = os.RemoveAll(zarfConfig.ZarfSBOMDir)
```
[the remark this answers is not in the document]
UncleGedd: Thanks, resolved!

src/test/e2e/bundle_test.go
UncleGedd: Agreed, made an issue to track: https://github.com/defenseunicorns/uds-cli/issues/126

src/pkg/sources/tarball.go
UncleGedd: Fixed

src/pkg/bundle/common.go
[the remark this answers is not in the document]
UncleGedd: Not supporting signing currently, recognize it as a need and made an issue: https://github.com/defenseunicorns/uds-cli/issues/127

src/pkg/bundle/remote.go
```diff
@@ -268,7 +201,7 @@ func (op *ociProvider) LoadBundle(_ int) (PathMap, error) {
 	doneSaving := make(chan int)
 	var wg sync.WaitGroup
 	wg.Add(1)
-	go zarfUtils.RenderProgressBarForLocalDirWrite(op.dst, estimatedBytes, &wg, doneSaving, fmt.Sprintf("Pulling bundle: %s", bundle.Metadata.Name))
+	go zarfUtils.RenderProgressBarForLocalDirWrite(op.dst, estimatedBytes, &wg, doneSaving, "update text goes here", fmt.Sprintf("Pulling bundle: %s", bundle.Metadata.Name))
```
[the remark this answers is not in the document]
UncleGedd: good catch, fixed

src/pkg/bundle/deploy.go
```diff
@@ -132,11 +123,32 @@ func (b *Bundler) Deploy() error {
...
+			if err != nil {
+				return err
+			}
+			source = &sources.RemoteBundle{
+				PkgName:        pkg.Name,
+				PkgOpts:        &opts,
+				PkgManifestSHA: sha,
+				TmpDir:         pkgTmp,
+				Remote:         remote,
+			}
```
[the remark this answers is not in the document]
UncleGedd: good idea, implemented

src/pkg/bundle/remove.go
```diff
@@ -40,21 +41,45 @@ func (b *Bundler) Remove() error {
...
+				return err
+			}
+			source = &sources.RemoteBundle{
+				PkgName:        pkg.Name,
+				PkgOpts:        &pkgCfg.PkgOpts,
+				PkgManifestSHA: sha,
+				TmpDir:         pkgTmp,
+				Remote:         remote,
+			}
+		}
```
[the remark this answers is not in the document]
UncleGedd: good idea, implemented

src/pkg/sources/remote.go
UncleGedd: Tracking, we restored OCI concurrency in push/pull ops but currently have a ticket to address it everywhere else: https://github.com/defenseunicorns/uds-cli/issues/115

src/pkg/sources/remote.go
UncleGedd: fixed

src/pkg/sources/tarball.go
UncleGedd: Fo sho, good call, removed Walk

src/pkg/sources/remote.go
UncleGedd: Good call, removed the Walk function and returned the proper layers from `downloadPkgFromRemoteBundle`

src/pkg/sources/remote.go
UncleGedd: We currently don't "officially" support signing bundles. I think some of the functionality is still there from your initial implementation but there wasn't a need for it at the time so we didn't continue to support. Agreed it does need to be implemented though, it's just not currently prioritized. Made an issue to track: https://github.com/defenseunicorns/uds-cli/issues/127

As for package integrity, good call and resolved!

mikevanhemert (approved):
lgtm

PR #132, opened 2023-10-24
adds id to goreleaser

UncleGedd:


decleaver (approved):
Looks good with the VIP go over

PR #137, opened 2023-10-26
feat: caching and OCI concurrency for remote packages during bundle create and deploy

UncleGedd:
- Implements caching and OCI concurrency for pulling remote Zarf packages during bundle create and deploy operations
- Refactors some really common OCI logic into functions now in `utils/oci.go`

### DIFF
--- a/src/pkg/bundler/remote.go
+++ b/src/pkg/bundler/remote.go
@@ -5,9 +5,7 @@
 package bundler
 
 import (
-	"bytes"
 	"context"
-	"encoding/json"
 	"fmt"
 	"io"
 	"path/filepath"
@@ -20,7 +18,6 @@ import (
 	zarfTypes "github.com/defenseunicorns/zarf/src/types"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"oras.land/oras-go/v2"
-	"oras.land/oras-go/v2/content"
 	ocistore "oras.land/oras-go/v2/content/oci"
 
 	"github.com/defenseunicorns/uds-cli/src/config"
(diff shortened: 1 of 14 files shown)


mikevanhemert (dismissed):
Looks great! Same potential concurrency issue in [src/pkg/sources/remote.go:171](https://github.com/defenseunicorns/uds-cli/pull/137/files#diff-cc6b07ec48af179448d51a500be72923462835fd2e22b0f5d6b49667c5be4ac9R171) - returning during a goroutine can result in [leaky goroutines](https://medium.com/codezillas/golang-leaky-goroutines-and-how-to-clean-them-30b505417028). I'll check issues and if I don't see one throw one in there

src/pkg/bundler/remote.go
```diff
@@ -31,10 +34,11 @@ type RemoteBundler struct {
 	RemoteSrc       *oci.OrasRemote
 	RemoteDst       *oci.OrasRemote
 	localDst        *ocistore.Store
+	tmpDir          string
 }
 
 // NewRemoteBundler creates a bundler to pull remote Zarf pkgs
-func NewRemoteBundler(pkg types.BundleZarfPackage, url string, localDst *ocistore.Store, remoteDst *oci.OrasRemote) (RemoteBundler, error) {
+func NewRemoteBundler(pkg types.BundleZarfPackage, url string, localDst *ocistore.Store, remoteDst *oci.OrasRemote, tmpDir string) (RemoteBundler, error) {
```
mikevanhemert, line 41: The logic here is hidden and not documented, and it is used three different ways. For code cleanup / simplicity, I recommend breaking this into three methods or providing some usage notes

mikevanhemert (approved):

src/pkg/bundler/remote.go
UncleGedd: 100% ! I think the `bundler` pkg itself is probably wrong. There's a matrix of operations to consider when doing local + remote bundles that interact with local + remote packages. TLDR: yes, def needs a refactor. In the interim I'll make a note to add comments indicating what happens when you pass the various args

PR #140, opened 2023-10-31
adds uds run command

UncleGedd:
Adds a `uds run` cmd to run an `tasks.yaml` that define reusable workflows. Needs docs, following up with this ticket: https://github.com/defenseunicorns/uds-cli/issues/141

### DIFF
--- a/src/cmd/internal.go
+++ b/src/cmd/internal.go
@@ -23,8 +23,8 @@ var internalCmd = &cobra.Command{
 	Short:   lang.CmdInternalShort,
 }
 
-var configSchemaCmd = &cobra.Command{
-	Use:     "config-schema",
+var configUDSSchemaCmd = &cobra.Command{
+	Use:     "config-uds-schema",
 	Aliases: []string{"c"},
 	Short:   lang.CmdInternalConfigSchemaShort,
 	Run: func(cmd *cobra.Command, args []string) {
@@ -37,8 +37,23 @@ var configSchemaCmd = &cobra.Command{
 	},
 }
 
+var configTasksSchemaCmd = &cobra.Command{
+	Use:     "config-tasks-schema",
+	Aliases: []string{"c"},
+	Short:   lang.CmdInternalConfigSchemaShort,
+	Run: func(cmd *cobra.Command, args []string) {
+		schema := jsonschema.Reflect(&types.TasksFile{})
+		output, err := json.MarshalIndent(schema, "", "  ")
+		if err != nil {
+			message.Fatal(err, lang.CmdInternalConfigSchemaErr)
+		}
+		fmt.Print(string(output) + "\n")
+	},
+}
+
 func init() {
 	rootCmd.AddCommand(internalCmd)
 
-	internalCmd.AddCommand(configSchemaCmd)
+	internalCmd.AddCommand(configUDSSchemaCmd)
+	internalCmd.AddCommand(configTasksSchemaCmd)
 }
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -39,5 +39,5 @@ var runCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(runCmd)
 	runFlags := runCmd.Flags()
-	runFlags.StringVar(&config.TaskFileLocation, "file", config.TasksYAML, lang.CmdRunFlag)
+	runFlags.StringVarP(&config.TaskFileLocation, "file", "f", config.TasksYAML, lang.CmdRunFlag)
 }
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023-Present The UDS Authors
+
+package cmd
+
+import (
+	"github.com/defenseunicorns/zarf/src/pkg/message"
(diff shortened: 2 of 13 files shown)


src/cmd/internal.go
```diff
@@ -37,8 +37,23 @@ var configSchemaCmd = &cobra.Command{
 	},
 }
 
+var configActionsSchemaCmd = &cobra.Command{
+	Use:     "config-actions-schema",
+	Aliases: []string{"c"},
+	Short:   lang.CmdInternalConfigSchemaShort,
+	Run: func(cmd *cobra.Command, args []string) {
+		schema := jsonschema.Reflect(&types.ActionsFile{})
```
corang, line 45: I opted to use comment descriptions instead of shoving more things into the struct tag, to get those comments in the schema you either need to use the same reflector as wren or move the comments back into the struct tags, I'd prefer to leave then as comments since it's much more readable

src/types/actions.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Package types contains all the types used by UDS.
+package types
+
+import (
+	zarfTypes "github.com/defenseunicorns/zarf/src/types"
+)
+
+type ActionsFile struct {
+	// Definitions and default values for variables used in run.yaml
```
corang, line 11: How descriptions are added to the structs

src/cmd/run.go
```diff
@@ -0,0 +1,42 @@
...
+	Use:   "run [ TASK NAME ]",
+	Short: "run a task",
+	Long:  `run a task from an actions file`,
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		var actionsFile types.ActionsFile
+
+		err := utils.ReadYaml(actionsFileLocation, &actionsFile)
+		if err != nil {
+			panic(err)
```
corang, line 29: I used panic for speed with wren, it should be changed to follow the pattern of uds-cli. Only putting the comment here but there are many places

src/test/base
```diff
@@ -0,0 +1 @@
+im a dude
```
corang, line 1: ??

corang (changes requested):

src/test/base
UncleGedd: she's a dude
he's a dude
cuz we're all dudes YEAH

plz tell me you know this [reference](https://www.youtube.com/watch?v=OnxWtp1FP6c) @corang !

src/cmd/run.go
UncleGedd: Screenshot 2023-10-31 at 6 29 11 PM
Good catch! I wasn't paying attention and my IDE automatically hid those error checks

src/cmd/internal.go
UncleGedd: I think it's def a stylistic thing, I actually prefer the struct tags because:
- it's one less reflective operation
- the comments look like comments, it's not clear that they serve another purpose without looking at another part of the codebase

Thoughts @decleaver ?

src/cmd/run.go
UncleGedd: fixed

mikevanhemert (approved):
Awesome to see this coming in! Great work @corang @Michael-Kruggel @UncleGedd @decleaver

zachariahmiller (approved):
🚀

corang:
@UncleGedd with actions already bring an explicit thing in the file what do you think of tasks.yaml instead?

src/cmd/internal.go
decleaver: Yeah, having comments that are more than just comments (aka actually matter) makes me nervous

jeff-mccoy: I avoid reflecting as much as possible.

UncleGedd:
> ady being an explicit thing in the file

Great idea, "actions is super overloaded as it is, changed to `tasks.yaml` and updated in code

src/cmd/run.go
```diff
@@ -0,0 +1,43 @@
...
+		if err := runner.Run(tasksFile, taskName); err != nil {
+			message.Fatalf(err, "Failed to run action: %s", err)
+		}
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(runCmd)
+	runFlags := runCmd.Flags()
+	runFlags.StringVar(&config.TaskFileLocation, "file", config.TasksYAML, lang.CmdRunFlag)
```
jeff-mccoy, line 42: would love to see shorthand too.

```suggestion
	runFlags.StringVarP(&config.TaskFileLocation, "file", "f", config.TasksYAML, lang.CmdRunFlag)
```

src/pkg/runner/runner.go
```diff
@@ -0,0 +1,435 @@
...
+}
+
+func (r *Runner) executeTask(task types.Task) error {
+	if len(task.Files) > 0 {
+		if err := r.placeFiles(task.Files); err != nil {
+			return err
+		}
+	}
+
+	if len(task.Actions) > 0 {
```
jeff-mccoy, line 71: I'd drop the `len()` checks here

UncleGedd: dropped!

jeff-mccoy (approved):
overall lgtm, there a few lint things that could improve over time, I'm a fan of more vs less code comments in non-trivial functions and I expect runner.go to slim up as we refactor zarf pieces.

PR #150, opened 2023-11-02
flattens task yaml structure

UncleGedd:
Removes `name` key under an `[]actions.task`, new syntax is:
```
actions:
  - task: name-of-task
```

### DIFF
--- a/.github/workflows/test-schema.yaml
+++ b/.github/workflows/test-schema.yaml
@@ -16,7 +16,7 @@ jobs:
         uses: ./.github/actions/golang
 
       - name: Docs and schemas
-        run: "make test-schema"
+        run: "make schema test-schema"
 
       - name: Save logs
         if: always()
--- a/hack/test-generate-schema.sh
+++ b/hack/test-generate-schema.sh
@@ -1,17 +1,16 @@
 #!/usr/bin/env sh
 
-if [ -z "$(git status -s uds.schema.json)" ]; then
-    echo "Success!"
-    exit 0
-else
-    git status uds.schema.json
-    exit 1
-fi
+check_git_status() {
+    if [ -z "$(git status -s "$1")" ]; then
+        echo "Success!"
+    else
+        echo "Schema changes found, please regenerate $1"
+        git status "$1"
+        exit 1
+    fi
+}
 
-if [ -z "$(git status -s tasks.schema.json)" ]; then
-    echo "Success!"
-    exit 0
-else
-    git status tasks.schema.json
-    exit 1
-fi
+check_git_status uds.schema.json
+check_git_status tasks.schema.json
+
+exit 0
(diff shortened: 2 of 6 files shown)


mjnagel (approved):

mikevanhemert:
How does this support or impact the ability to reference tasks that are remote?

For instance, I may define a set of tasks that are universally use across a suite of repositories (stand up a cluster, tear down a cluster, etc) and instead of copy-pasting those tasks reop-to-repo, prefer to keep things DRY. It seems to me the older schema may have afforded that evolution better this update.

jeff-mccoy:
I love the shorthand, but would be good to understand our intentions with remote tasks as I don't fully know what they would be yet.

mikevanhemert:
> I love the shorthand, but would be good to understand our intentions with remote tasks as I don't fully know what they would be yet.

Entirely possible that I'm behind in my understanding of things. There were some early discussions surrounding DRY with whatever task running we do. Could I create a handful of tasks that are useful to many repos, stick them in one location, _and_ use them for both CI and local development? That would be very powerful

mikevanhemert:
With the current PR, it might be this easy

```
#local task
- name: a-task
  actions:
    - task: local

- name: local
  actions:
    - cmd: echo "hello local"
...
# remote task
actions:
  - task: remote
    ref: https://raw.githubusercontent.com/defenseunicorns/uds-cli/b923cba25730bb487f2e619a0b58d4b9141adecd/src/test/tasks.yaml
```

UncleGedd:
I created a [follow up ticket](https://github.com/defenseunicorns/uds-cli/issues/151) to support remote task files, if that's something we think we need today I can create a follow-on PR or we can let that issue simmer for a bit

zachariahmiller:
> With the current PR, it might be this easy
> 
> ```
> #local task
> - name: a-task
>   actions:
>     - task: local
> 
> - name: local
>   actions:
>     - cmd: echo "hello local"
> 
> ...
> 
> # remote task
> actions:
>   - task: remote
>     ref: https://raw.githubusercontent.com/defenseunicorns/uds-cli/b923cba25730bb487f2e619a0b58d4b9141adecd/src/test/tasks.yaml
> ```
> 
> So the flattened syntax remains, but there's a way to reference remote tasks if desired

While i agree this works, it becomes pretty verbose is a remote file of (common) tasks or even just another file with common tasks in a different directory is pulled in.

What seems much nicer is this sort of syntax with
```yaml
include:
   common: ./extra-tasks.yaml
tasks:
  - task: local-task
  - task: common:remote-taske
``` 

see https://github.com/defenseunicorns/JosueAvena/blob/b318483bd1364540747e8b519fcf8c12158a9750/Taskfile.yml#L40

UncleGedd:
^^ added @zachariahmiller 's suggested syntax to that follow-on ticket as well

decleaver (approved):
Do it. (looks good)

PR #166, opened 2023-11-06
fix: use system Zarf so that init works

UncleGedd:
When the Zarf init package is creating gitea users, it users a Zarf `internal` command that is not build into the UDS CLI binary. This means that users don't get created unless we configure the Zarf lib to use the system Zarf


decleaver (approved):
LGTM

PR #168, opened 2023-11-06
docs: uds runner

UncleGedd:
Adds `docs/runner.md` for the UDS runner feature

### DIFF
--- a/docs/runner.md
+++ b/docs/runner.md
@@ -1,0 +1,238 @@
+# UDS Runner
+
+UDS runner enables UDS Bundle developers to automate UDS builds and perform common shell tasks. It
+uses [Zarf](https://zarf.dev/) under the hood to perform tasks and shares a syntax similar to `zarf.yaml` manifests.
+Many [Zarf Actions features](https://docs.zarf.dev/docs/create-a-zarf-package/component-actions) are also available in
+UDS runner.
+
+## Table of Contents
+
+1. [Quickstart](#quickstart)
+2. [Key Concepts](#key-concepts)
+    - [Tasks](#tasks)
+    - [Actions](#actions)
+        - [Task](#task)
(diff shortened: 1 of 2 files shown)


docs/runner.md
```diff
@@ -0,0 +1,250 @@
...
+
+Actions can run arbitrary bash commands including in-line scripts, and the output of a command can be placed in a
+variable using the `setVariables` key
+
+```yaml
+tasks:
+  - name: foo
+    actions:
+      - cmd: |
+          ARCH=$(uname -m)
```
zachariahmiller, line 138: if you indent the actual shell commands to the same indentation as the |, does the syntax issue go away?

Also maybe you can update this to be `cmd: uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/'`

UncleGedd: good call, I didn't see that it was jacked, fixing now

UncleGedd:
TIL Github hates <style>

docs/runner.md
```diff
@@ -0,0 +1,239 @@
...
+#### Cmd
+
+Actions can run arbitrary bash commands including in-line scripts, and the output of a command can be placed in a
+variable using the `setVariables` key
+
+```yaml
+tasks:
+  - name: foo
+    actions:
+      - cmd: uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/'
```
jeff-mccoy, line 136: now that #160 is merged, maybe we should do some different example so ppl aren't confused and try to use that somewhere vs the built-in env variable?

UncleGedd: Updated with `echo -n 'dHdvIHdlZWtzIG5vIHByb2JsZW0=' | base64 -d`

mikevanhemert (approved):

PR #178, opened 2023-11-07
fix: hotfix for broken pull operation

UncleGedd:
Removes annotation from root manifest layers to fix broken pull. Other logic in the codebase depends on those annotations and adding an additional one jacked up that logic


decleaver (approved):
LGTM

PR #181, opened 2023-11-08
feat: import local tasks into runner

UncleGedd:
Adds ability to import local task files into the runner using syntax:
```yaml
includes:
  - foo: ./tasks-to-import.yaml
tasks:
  - name: the-task
    actions:
      - task: foo:imported-task
```
Where `imported-task` exists in the `tasks-to-import.yaml`

### DIFF
--- a/src/pkg/runner/runner.go
+++ b/src/pkg/runner/runner.go
@@ -17,7 +17,7 @@ import (
 
 	"github.com/defenseunicorns/zarf/src/config/lang"
 	"github.com/defenseunicorns/zarf/src/pkg/message"
-	"github.com/defenseunicorns/zarf/src/pkg/utils"
+	zarfUtils "github.com/defenseunicorns/zarf/src/pkg/utils"
 	"github.com/defenseunicorns/zarf/src/pkg/utils/helpers"
 	zarfTypes "github.com/defenseunicorns/zarf/src/types"
 	"github.com/mholt/archiver/v3"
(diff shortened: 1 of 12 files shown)


mikevanhemert (commented):
I like that the [included task files' paths are relative to the location of the parent task ](https://github.com/defenseunicorns/uds-cli/pull/181/files#diff-6553d584900e07480c3dbfce800c02fc84c8fe97202b4a8523b86e69c200c741R63) file. This should allow for really clean execution and nesting without any strange env variables or extra path params.

src/pkg/runner/runner.go
```diff
@@ -52,6 +57,23 @@ func Run(tasksFile types.TasksFile, taskName string, setVariables map[string]str
...
+		taskFileDir := filepath.Dir(config.TaskFileLocation)
+		includePath := filepath.Join(taskFileDir, include)
+
+		var tasksFile types.TasksFile
+		if err := zarfUtils.ReadYaml(includePath, &tasksFile); err != nil {
+			return err
+		}
+
+		// add the tasks to the runner
+		r.TasksFile.Tasks = append(r.TasksFile.Tasks, tasksFile.Tasks...)
```
mikevanhemert, line 72: Should the included task files also be sent to `importTasks`? The pattern should be fairly straightforward and it would provide a lot of flexibility to UDS CLI users

mikevanhemert (changes requested):

src/pkg/runner/runner.go
UncleGedd: doneskies

mikevanhemert (approved):

PR #190, opened 2023-11-13
feat: adds simple Helm overrides

jeff-mccoy:
## Description
Enables simple Helm overrides in `uds-bundle.yaml`. Note that overwritten values cannot be objects or lists, only simple values like strings, numbers, etc

## Related Issue

Fixes # https://github.com/defenseunicorns/uds-cli/issues/192

## Type of change

## Checklist before merging


decleaver (approved):
shipit!

PR #191, opened 2023-11-13
fix: updates release CI to write to GHCR for e2e tests

UncleGedd:
## Description
Fixes bug in release CI where an e2e test was unable to access GHCR


decleaver (approved):
Shame! ...on me... didn't even think of it. Looks good!

PR #202, opened 2023-11-15
Spike: Charm with pterm

UncleGedd:
## Description
Spike to determine if we can use Charm/Bubbletea simultaneously with pterm

## Notes
- disable pterm output when program begins but set it's default output to be a buffer we create
- right before calling deploy(), enable pterm output
- grab that pterm output and put it in a Bubbletea viewport


UncleGedd:
Would like to opt out of [printing this table](https://github.com/defenseunicorns/zarf/blob/8763b6b2bf47ded67fb33130161b46c85086ce32/src/pkg/packager/deploy.go#L84) in Zarf

^this is an issue because this function eventually calls [this](https://github.com/defenseunicorns/zarf/blob/8763b6b2bf47ded67fb33130161b46c85086ce32/src/pkg/message/credentials.go#L35) which overwrites my pterm config to send output to the buffer

UncleGedd:
Another looong running PR, going to close the PR but keep the branch

PR #270, opened 2023-12-13
fix: nested includes of uds tasks do not resolve paths as expected

### LINKED ISSUE #250: Nested includes of uds tasks do not resolve paths as expected
### Environment

Device and OS: MBP, macOS
App version: 0.4.1

### Steps to reproduce

File/folder structure:
```
tasks.yaml
tasks/foo.yaml
tasks/bar.yaml
```

```
# tasks.yaml

includes:
  - foo: ./tasks/foo.yaml

tasks:
  - name: foobar
    actions:
      - task: foo:foobar
```

```
# foo.yaml

includes:
  - bar: ./bar.yaml

tasks:
  - name: foobar
    actions:
      - cmd: "echo foo"
      - task: bar:bar
```

```
# bar.yaml

tasks:
  - name: bar
    actions:
      - cmd: "echo bar"
```

With this setup I would expect to be able to run two tasks with the same results:
```
uds run foobar
uds run -f tasks/foo.yaml foobar
```

### Expected result

Both would have the echos of foo and bar.

### Actual Result

- `uds run foobar`: Errors on the includes: `ERROR:  Failed to run action: unable to read included file bar.yaml: open bar.yaml: no such file or directory`.

- `uds run -f tasks/foo.yaml foobar`: succeeds and logs foo/bar.

It appears as if paths for includes are always resolved relative to the top level task file (what is passed into `uds run`) rather than relative to the place that it was included from. Worth noting that adding `./tasks/` to the inclusion of `bar` will fix `uds run foobar` but break `uds run -f tasks/foo.yaml foobar`. I would have expected that all includes are resolved relative to the direct parent that is including them.

decleaver:
## Description
Fixes how uds tasks resolve includes paths relative to the file

## Related Issue

Fixes #250 

## Type of change

## Checklist before merging


UncleGedd (commented):
Awesome work @decleaver ! Could you also add a line in the docs describing how `includes` paths are relative to the `tasks.yaml` that `includes` them?

decleaver:
added "first class" ticket documentation as part of this PR as well

UncleGedd (approved):

PR #271, opened 2023-12-14
docs: shared vars + first class UDS

UncleGedd:
## Description
- Adds docs on how to share the same variable between multiple packages with minimal syntax
- Adds more docs + examples on UDS first class things


decleaver (approved):
Similar to the words I had, but better and more clear lol

Issue #276, opened 2023-12-15
uds-config parsing failures


mjnagel:
### Environment

Device and OS: macOS
App version: 0.5.0 (but bug is also present on 0.4.1)
Kubernetes distro being used: k3d/rke2, but also not relevant

### Steps to reproduce

Two ways to view the bug:

Option 1:
1. Create a bundle of any sort.
2. Create an empty config file, i.e. `touch uds-config.yaml`
3. Attempt to deploy the bundle.
4. See error: `ERROR:  Failed to load uds-config: node not found`

Option 2:
1. Create a bundle of any sort.
2. Create a config file without `variables` specified:
    ```yaml
    options:
      log-level: debug
    ```
4. Attempt to deploy the bundle.
5. See error: `ERROR:  Failed to load uds-config: node not found`

### Expected result

uds-cli should honor my options or an empty config file.

### Actual Result

The CLI appears to error whenever `variables` is not present (and on previous versions whenever `bundle.deploy.zarf-packages` was not present).

### Severity/Priority

Workaround is available by adding an empty `variables: {}`.

PR #277, opened 2023-12-15
fix: properly parse config files missing `variables`

mjnagel:
## Description

Switches to unmashal with goyaml + `omitempty` on `variables`.

Also added a sample uds-config that just uses options (no variables) on the 01 test to hopefully catch this in the future, can turn it into a separate test case if desired.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-cli/issues/276

## Type of change

## Checklist before merging

- [x] [Contributor Guide Steps](https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md)(https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md#submitting-a-pull-request) followed

### DIFF
--- a/src/cmd/uds.go
+++ b/src/cmd/uds.go
@@ -5,7 +5,6 @@
 package cmd
 
 import (
-	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


UncleGedd (approved):
awesome work, thanks again!

PR #281, opened 2023-12-15
feat: don't require "oci://" when performing bundle operations

### LINKED ISSUE #278: Don't require "oci://" when performing bundle operations
Today, some operations require the `oci://` prefix with interacting with remote bundles. To avoid confusion, ensure all bundle commands that interact with remote registries work both with or without the `oci://` prefix. For example, both of these `deploy` commands should work:
- `uds deploy ghcr.io/my-bundle:0.1.0`
- `uds deploy oci://ghcr.io/my-bunde:0.1.0`

decleaver:
## Description

Today, some operations require the oci:// prefix with interacting with remote bundles. To avoid confusion, ensure all bundle commands that interact with remote registries work both with or without the oci:// prefix. For example, both of these deploy commands should work:

uds deploy ghcr.io/my-bundle:0.1.0
uds deploy oci://ghcr.io/my-bunde:0.1.0

## Related Issue

Fixes #278 

## Type of change

## Checklist before merging


src/pkg/bundle/remote.go
```diff
@@ -316,3 +318,12 @@ func CheckOCISourcePath(source string) string {
 	}
 	return source
 }
+
+// EnsureOCIPrefix ensures oci prefix is part of provided remote source path, and adds it if it's not
+func EnsureOCIPrefix(source string) string {
+	var ociPrefix = "oci://"
+	if source[:6] != ociPrefix {
```
UncleGedd, line 325: instead of the magic `6`, can we do it by the length of `oci://`?

UncleGedd: ie. something like ` ociPrefix = len("oci://")` and then `if source [:ociPrefix] ...`

decleaver: unless you prefer creating a new variable, vs using the function result directly... i can update if so

UncleGedd: nah all good! just wanted to refactor the magic number

UncleGedd (approved):
🚀

PR #282, opened 2023-12-15
chore: only process includes that are needed by task

### LINKED ISSUE #280: Don't process all includes if not needed to run a task
If you are running a task that references an include, uds-cli processes all of the includes, even if it doesn't need all of them. Issue was found with the following test :

```
	t.Run("test includes paths", func(t *testing.T) {
		t.Parallel()

		// NOTE: Even though the foobar task doesn't use remote include, since it does reference an "include", all the "includes" get processed,
		// which means we must set "GIT_REVISION" for the remote include
		// get current git revision
		gitRev, err := e2e.GetGitRevision()
		if err != nil {
			return
		}
		setVar := fmt.Sprintf("GIT_REVISION=%s", gitRev)

		stdOut, stdErr, err := e2e.RunTasksWithFile("run", "foobar", "--set", setVar)
		require.NoError(t, err, stdOut, stdErr)
		require.Contains(t, stdErr, "echo foo")
		require.Contains(t, stdErr, "echo bar")
	})
```

Given the following tasks.yaml content:

```
includes:
  - local: ./tasks-to-import.yaml
  - foo: ./more-tasks/foo.yaml
  - remote: https://raw.githubusercontent.com/defenseunicorns/uds-cli/${GIT_REVISION}/src/test/tasks/remote-import-tasks.yaml

- name: foobar
    actions:
      - task: foo:foobar
```

We had to set the GIT_REVISION var because "remote" needs it, even though it does not get used by the foobar task.

decleaver:
## Description

If you are running a task that references an include, uds-cli processes all of the includes, even if it doesn't need all of them. This PR makes it so only the includes that are needed for a task are processed

## Related Issue

Fixes #280 

## Type of change

## Checklist before merging


src/pkg/runner/runner.go
```diff
@@ -50,11 +50,17 @@ func Run(tasksFile types.TasksFile, taskName string, setVariables map[string]str
 	// only process includes if the task requires them
 	for _, a := range task.Actions {
 		if strings.Contains(a.TaskReference, ":") {
-			err = runner.importTasks(tasksFile.Includes,config.TaskFileLocation)
-			if err != nil {
-				return err
+			taskReferenceName := strings.Split(a.TaskReference, ":")[0]
+			for _, b:= range tasksFile.Includes{
```
UncleGedd, line 54: What does `b` refer to in this case?

UncleGedd:
Great work! Love the simple fix, one quick Q about variable naming

src/pkg/runner/runner.go
decleaver: not a good sign if you have to ask lol, updated var name to "include"

UncleGedd (approved):
🚀

PR #290, opened 2023-12-18
fix: handle multiple nested includes

### LINKED ISSUE #283: Nested tasks with Includes
### Environment

Device and OS: Ubuntu
App version: 0.5.1 & 0.3.1 & 0.4.1

### Steps to reproduce
Screenshot from 2023-12-15 14-49-50

In the screenshot refer to the cluster-full task definition on line 22. It refers to the cluster-create task definition which refers to an included task from the `setup.yaml` tasks directory. 

When trying to run a `uds run cluster-full` it fails saying that it cannot find the `setup:k3d-cluster task`. @UncleGedd & @zachariahmiller and I all looked at this and the file naming and paths were all correct, so not a naming problem. To further prove that theory, if i were to un-comment the cluster-full definition on line 8 of the screen shot, and run `uds run cluster-full` it would work as expected. 

Also tested this across several uds versions, originally when i encountered this error i was on v0.3.1, tried upgrading to v0.4.1, and then also v0.5.1. Same problem persisted across versions.

### Expected result
Being able to reference an included task in a nested task reference.

### Actual Result
Unable to reference the included task.

### Severity/Priority
Low Priority

decleaver:
## Description

Issue was found where if you reference a task with multiple includes from another task, it errors out not being able to pull in those includes. Not able to process includes of "sibling"

## Related Issue

Fixes #283 

## Type of change

## Checklist before merging

### DIFF
--- a/src/test/tasks/tasks.yaml
+++ b/src/test/tasks/tasks.yaml
@@ -96,3 +96,6 @@ tasks:
     actions:
       - task: foo:foobar
       - task: remote:echo-var
+  - name: extra-foobar
+    actions:
+      - task: more-foobar
--- a/src/pkg/runner/runner.go
+++ b/src/pkg/runner/runner.go
@@ -62,9 +62,9 @@ func (r *Runner) processIncludes(task types.Task, tasksFile types.TasksFile) err
 	for _, a := range task.Actions {
 		if strings.Contains(a.TaskReference, ":") {
 			taskReferenceName := strings.Split(a.TaskReference, ":")[0]
-			for _, b := range tasksFile.Includes {
-				if b[taskReferenceName] != "" {
-					referencedIncludes := []map[string]string{b}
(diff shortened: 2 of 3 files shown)


src/test/tasks/tasks.yaml
```diff
@@ -96,3 +96,6 @@ tasks:
     actions:
       - task: foo:foobar
       - task: remote:echo-var
+  - name: extra-foobar
```
UncleGedd, line 99: Does any test use this new `extra-foobar` task?

decleaver: lol my local debugger, let me create that test

decleaver: pushed a test for the rest of the world lol

UncleGedd (approved):
🚀

Issue #296, opened 2024-01-01
Zarf v0.32 break UDS CLI (oci change)


jeff-mccoy:
In [v0.32](https://github.com/defenseunicorns/zarf/releases/tag/v0.32.0) the OCI refs have changed and no longer publish with the arch in the tag. Need to update the UDS CLI lib to support this (better) publishing strategy.

<img width="458" alt="Screenshot 2024-01-01 at 5 29 56 PM" src="https://github.com/defenseunicorns/uds-cli/assets/882485/2cba8bf0-befe-44a8-ad24-b7b51c4e6b99">
<img width="824" alt="Screenshot 2024-01-01 at 5 29 51 PM" src="https://github.com/defenseunicorns/uds-cli/assets/882485/0fe71f4f-060d-4b51-8198-b51a93e052b1">

decleaver:
updates in https://github.com/defenseunicorns/uds-cli/pull/300

PR #300, opened 2024-01-04
chore: update to zarf version 0.32.1

decleaver:
## Description

update uds-cli zarf version to v0.32.1 and supports zarf multi arch

## Related Issue

Relates to #296 

## Type of change

## Checklist before merging


src/pkg/utils/sbom.go
```diff
@@ -47,23 +47,26 @@ func SBOMExtractor(dst string, SBOMArtifactPathMap map[string]string) func(ctx c
...
+			_, err = open.Read(buffer)
+			if err != nil {
+				return err
+			}
+			err = open.Close()
+			if err != nil {
+				return err
+			}
+			path := filepath.Join(dst, config.BundleSBOM, f.NameInArchive)
+			// todo: handle collisions? especially for zarf-component SBOM files?
```
Racer159, line 62: if you more simply wrap the tarballs from each package that will naturally separate them and may simplify the logic for matching OCI descriptors and extracting things from them.

Also note how `zarf package inspect <package> --sbom-out directory` works (it makes sub dirs for the package by name).

Racer159: This logic also clobbers the compare.html which will cause weirdness in the SBOM selector when going to the compare screen.

Racer159: (it may be good to have a tech debt issue / discussion on how the Zarf team can help - it may be good for y'all to make your own high level compare file and such too).

decleaver: Thanks @Racer159 , created the following issue to address the above: https://github.com/defenseunicorns/uds-cli/issues/305

mikevanhemert (approved):
Ran well locally

Issue #268, opened 2023-12-13
Accepts uds-bundle.yaml and uds-bundle.yml


UncleGedd:
To prevent users from seeing silly errors, accept both `.yaml` and `yml`

anthonywendt:
Was going to use this issue to dig in and get more familiar with uds-cli on the side this week. If you want to boot me from it I wont be offended.

PR #310, opened 2024-01-11
feat: handle bundle yaml and yml files

anthonywendt:
## Description
Updated to handle both uds-bundle.yaml and uds-bundle.yml

## Related Issue

Fixes https://github.com/defenseunicorns/uds-cli/issues/268

## Type of change

## Checklist before merging

- [x] [Contributor Guide Steps](https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md)(https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md#submitting-a-pull-request) followed

### DIFF
--- a/src/cmd/uds.go
+++ b/src/cmd/uds.go
@@ -32,15 +32,16 @@ var createCmd = &cobra.Command{
 			if !zarfUtils.IsDir(args[0]) {
 				message.Fatalf(nil, "(%q) is not a valid path to a directory", args[0])
 			}
-			pathToBundleFile = args[0] + "/"
+			pathToBundleFile = filepath.Join(args[0])
 		}
 		// Handle .yaml or .yml
-		if _, yamlErr := os.Stat(pathToBundleFile + config.BundleYAML); yamlErr == nil {
-			bundleCfg.CreateOpts.UserProvidedBundleYamlFile = config.BundleYAML
-		} else if _, ymlErr := os.Stat(pathToBundleFile + strings.Replace(config.BundleYAML, ".yaml", ".yml", 1)); ymlErr == nil {
-			bundleCfg.CreateOpts.UserProvidedBundleYamlFile = strings.Replace(config.BundleYAML, ".yaml", ".yml", 1)
+		bundleYml := strings.Replace(config.BundleYAML, ".yaml", ".yml", 1)
+		if _, err := os.Stat(filepath.Join(pathToBundleFile, config.BundleYAML)); err == nil {
+			bundleCfg.CreateOpts.BundleFile = config.BundleYAML
+		} else if _, err = os.Stat(filepath.Join(pathToBundleFile, bundleYml)); err == nil {
+			bundleCfg.CreateOpts.BundleFile = bundleYml
 		} else {
-			message.Fatalf(ymlErr, "Neither %s or %s found", config.BundleYAML, strings.Replace(config.BundleYAML, ".yaml", ".yml", 1))
+			message.Fatalf(err, "Neither %s or %s found", config.BundleYAML, bundleYml)
 		}
 	},
 	Run: func(cmd *cobra.Command, args []string) {
--- a/src/cmd/uds.go
+++ b/src/cmd/uds.go
@@ -27,11 +27,20 @@ var createCmd = &cobra.Command{
 	Args:    cobra.MaximumNArgs(1),
 	Short:   lang.CmdBundleCreateShort,
 	PreRun: func(cmd *cobra.Command, args []string) {
-		if len(args) > 0 && !zarfUtils.IsDir(args[0]) {
-			message.Fatalf(nil, "(%q) is not a valid path to a directory", args[0])
+		pathToBundleFile := ""
+		if len(args) > 0 {
+			if !zarfUtils.IsDir(args[0]) {
+				message.Fatalf(nil, "(%q) is not a valid path to a directory", args[0])
+			}
+			pathToBundleFile = args[0] + "/"
 		}
-		if _, err := os.Stat(config.BundleYAML); len(args) == 0 && err != nil {
-			message.Fatalf(err, "%s not found in directory", config.BundleYAML)
+		// Handle .yaml or .yml
+		if _, yamlErr := os.Stat(pathToBundleFile + config.BundleYAML); yamlErr == nil {
+			bundleCfg.CreateOpts.UserProvidedBundleYamlFile = config.BundleYAML
+		} else if _, ymlErr := os.Stat(pathToBundleFile + strings.Replace(config.BundleYAML, ".yaml", ".yml", 1)); ymlErr == nil {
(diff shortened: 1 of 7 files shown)


decleaver (approved):
pulled in locally and ran some extra checks with the uds-config.yml, looks good!

decleaver (approved):
pulled in locally and ran some extra checks with the uds-config.yml! 
Reviewed with anthony, looks good

src/cmd/uds.go
```diff
@@ -27,11 +27,20 @@ var createCmd = &cobra.Command{
...
+		if len(args) > 0 {
+			if !zarfUtils.IsDir(args[0]) {
+				message.Fatalf(nil, "(%q) is not a valid path to a directory", args[0])
+			}
+			pathToBundleFile = args[0] + "/"
 		}
-		if _, err := os.Stat(config.BundleYAML); len(args) == 0 && err != nil {
-			message.Fatalf(err, "%s not found in directory", config.BundleYAML)
+		// Handle .yaml or .yml
+		if _, yamlErr := os.Stat(pathToBundleFile + config.BundleYAML); yamlErr == nil {
```
UncleGedd, line 38: I see the intent here, but `yamlErr` seems misleading as var name; it's not actually a YAML error

src/cmd/uds.go
```diff
@@ -27,11 +27,20 @@ var createCmd = &cobra.Command{
...
+			bundleCfg.CreateOpts.UserProvidedBundleYamlFile = config.BundleYAML
+		} else if _, ymlErr := os.Stat(pathToBundleFile + strings.Replace(config.BundleYAML, ".yaml", ".yml", 1)); ymlErr == nil {
```
UncleGedd, line 40: Referring to `os.Stat(pathToBundleFile + strings.Replace(config.BundleYAML, ".yaml", ".yml", 1))`, can we DRY this up a bit? Repeating the `strings.Replace` bit makes for some long lines

src/config/config.go
```diff
@@ -20,7 +20,7 @@ const (
 	// BlobsDir is the string for the blobs/sha256 dir in an OCI artifact
 	BlobsDir = "blobs/sha256"
 
-	// BundleYAML is the string for zarf.yaml
+	// BundleYAML is the string for uds-bundle.yaml
```
UncleGedd, line 23: good catch!

src/pkg/bundle/create.go
```diff
@@ -39,7 +39,7 @@ func (b *Bundler) Create() error {
 	defer os.Chdir(cwd)
 
 	// read the bundle's metadata into memory
-	if err := utils.ReadYaml(config.BundleYAML, &b.bundle); err != nil {
+	if err := utils.ReadYaml(b.cfg.CreateOpts.UserProvidedBundleYamlFile, &b.bundle); err != nil {
```
UncleGedd, line 42: `UserProvidedBundleYamlFile` feels a bit verbose as a name. All bundle files are user provided, recommended shortening to something like `BundleFile`

src/cmd/uds.go
```diff
@@ -27,11 +27,20 @@ var createCmd = &cobra.Command{
...
 	Short:   lang.CmdBundleCreateShort,
 	PreRun: func(cmd *cobra.Command, args []string) {
-		if len(args) > 0 && !zarfUtils.IsDir(args[0]) {
-			message.Fatalf(nil, "(%q) is not a valid path to a directory", args[0])
+		pathToBundleFile := ""
+		if len(args) > 0 {
+			if !zarfUtils.IsDir(args[0]) {
+				message.Fatalf(nil, "(%q) is not a valid path to a directory", args[0])
+			}
+			pathToBundleFile = args[0] + "/"
```
UncleGedd, line 35: Instead of manually building out paths, we prefer to use `filepath.Join` in Go

src/test/e2e/bundle_test.go
```diff
@@ -272,3 +272,32 @@ func TestBundleWithGitRepo(t *testing.T) {
...
+	zarfPublish(t, pkg, "localhost:889")
+
+	bundleDir := "src/test/bundles/09-uds-bundle-yml"
+	bundlePath := filepath.Join(bundleDir, fmt.Sprintf("uds-bundle-yml-example-%s-0.0.1.tar.zst", e2e.Arch))
+
+	create(t, bundleDir) // todo: allow creating from both the folder containing and direct reference to uds-bundle.yaml
+	inspect(t, bundlePath)
+	inspectAndSBOMExtract(t, bundlePath)
+	// Test with an "options only" config file
+	os.Setenv("UDS_CONFIG", filepath.Join("src/test/bundles/09-uds-bundle-yml", "uds-config.yml"))
```
UncleGedd, line 300: nice!

UncleGedd (changes requested):

src/pkg/bundle/create.go
anthonywendt: I have irrational flashbacks digging through old fortran code with cryptic single letter variables that I over compensate for. Sorry about that.

src/cmd/uds.go
anthonywendt: Updated

src/cmd/uds.go
anthonywendt: Updated

src/cmd/uds.go
anthonywendt: Updated

src/cmd/uds.go
```diff
@@ -27,11 +27,21 @@ var createCmd = &cobra.Command{
...
+		if len(args) > 0 {
+			if !zarfUtils.IsDir(args[0]) {
+				message.Fatalf(nil, "(%q) is not a valid path to a directory", args[0])
+			}
+			pathToBundleFile = filepath.Join(args[0])
 		}
-		if _, err := os.Stat(config.BundleYAML); len(args) == 0 && err != nil {
-			message.Fatalf(err, "%s not found in directory", config.BundleYAML)
+		// Handle .yaml or .yml
+		bundleYml := strings.Replace(config.BundleYAML, ".yaml", ".yml", 1)
```
UncleGedd, line 38: Thanks for that refactor. One more thing, looks like you can remove the `pathToBundleFile` variable entirely and replace it in those `if` statements with `args[0]`

anthonywendt: I don't believe I can do that. The way I wrote this code, the handle .yaml or .yml section can handle either file type, whether or not the user provided a directory to the create command. If I make the change you suggested and the user doesn't provide a directory, args[0] will error with index out of range. This is why, while checking if the user provided a directory and checking if its a valid directory I join it to the `pathToBundleFile` string. Using the `pathToBundleFile` string allows the following .yaml vs .yml checks to work in either situation without issue. Unless I'm overlooking something or misunderstanding the suggestion.

UncleGedd: You are def correct, thanks for that clarification!

UncleGedd (approved):

PR #329, opened 2024-01-17
chore: refactor Annotations and SBOM logic

UncleGedd:
## Description
- refactors instances of `desc.Annotations == nil` to have more specific checks; this will allow us to add annotations (not title!) to the root manifest of bundles
- adds test for plain `uds` cmd
- refactors SBOM logic and provides better warnings
- docs fix


decleaver (approved):
lgtm, went over in gather

PR #341, opened 2024-01-22
bug: ensure remote bundles have an index.json

UncleGedd:
## Description
In preparation for multi-arch bundles, ensure remote bundle OCI artifacts have an `index.json` that includes a Platform


decleaver (approved):
walked through all the changes in gather

Issue #306, opened 2024-01-09
`uds remove` uses the wrong zarf package name during removal


mjnagel:
### Environment

Device and OS: macOS
App version: 0.5.2
Kubernetes distro being used: k3d/k3s 1.27 (issue also observed on rke2 1.27)

### Steps to reproduce

1. Deploy a bundle containing a single package. The package should be named differently from the zarf package, i.e.:
    ```yaml
    packages:
      - name: foo
        repository: ghcr.io/packages/bar # Zarf package named bar
    ```
2. Attempt to remove the bundle with `uds remove`.

### Expected result

Bundle is removed.

### Actual Result

```
      WARNING  Skipping removal of foo. Package not deployed
```

### Severity/Priority

Mid

### Additional Context

It appears that uds-cli is using the name specified in the bundle when trying to do a removal, rather than the underlying zarf package name.

PR #342, opened 2024-01-22
feat: allow custom zarf package names for remote package refs

decleaver:
## Description

Allows you to use custom names when referencing a remote zarf package in your uds-bundle.yaml. This does not apply when referencing local zarf packages. When referencing local zarf packages, the bundle package name must still match the actual zarf package name

## Related Issue

Relates to #306 

## Type of change

## Checklist before merging

### DIFF
--- a/src/config/config.go
+++ b/src/config/config.go
@@ -52,6 +52,10 @@ const (
 
 	// EnvVarPrefix is the prefix for environment variables to override bundle helm variables
 	EnvVarPrefix = "UDS_"
+
+	ZarfPackageNameAnnotation = "zarf.package.name"
+
+	UDSPackageNameAnnotation = "uds.package.name"
 )
 
 var (
--- a/src/config/config.go
+++ b/src/config/config.go
@@ -53,8 +53,10 @@ const (
 	// EnvVarPrefix is the prefix for environment variables to override bundle helm variables
 	EnvVarPrefix = "UDS_"
 
+	// ZarfPackageNameAnnotation is the annotation key for the value that specifies the zarf package name
 	ZarfPackageNameAnnotation = "zarf.package.name"
 
+	// UDSPackageNameAnnotation is the annotation key for the value that specifies the name given to a zarf package in the uds-bundle.yaml
 	UDSPackageNameAnnotation = "uds.package.name"
 )
 
--- a/src/pkg/bundle/bundle.go
+++ b/src/pkg/bundle/bundle.go
@@ -53,7 +53,7 @@ func Create(b *Bundler, signature []byte) error {
 	for i, pkg := range bundle.Packages {
 		fetchSpinner := message.NewProgressSpinner("Fetching package %s", pkg.Name)
 		zarfPackageName := ""
-		zarfLayerAdded := false
+		zarfRootLayerAdded := false
 		defer fetchSpinner.Stop()
 
 		if pkg.Repository != "" {
@@ -69,7 +69,7 @@ func Create(b *Bundler, signature []byte) error {
 			}
 
 			// grab layers for archiving
-			for i, layerDesc := range layerDescs {
+			for _, layerDesc := range layerDescs {
 				if layerDesc.MediaType == ocispec.MediaTypeImageManifest {
 					// rewrite the Zarf image manifest to have media type of Zarf blob
 					err = os.Remove(filepath.Join(b.tmp, config.BlobsDir, layerDesc.Digest.Encoded()))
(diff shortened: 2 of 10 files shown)


src/pkg/bundle/bundle.go
```diff
@@ -81,7 +82,37 @@ func Create(b *Bundler, signature []byte) error {
...
 					// ensure media type is Zarf blob for layers in the bundle's root manifest
 					layerDesc.MediaType = oci.ZarfLayerMediaTypeBlob
+
+					// add package name annotations
+					annotations := make(map[string]string)
+					layerDesc.Annotations = annotations
+					layerDesc.Annotations[config.UDSPackageNameAnnotation] = pkg.Name
+
+					// If zarf package name has been obtained from zarf config, set the zarf package name annotation
+					if zarfPackageName != "" {
```
UncleGedd, line 92: Won't `zarfPackageName` always be `""`? I see on line 55 where we define it to be an empty string, but I don't see where that would change before it hits line 92

src/config/config.go
```diff
@@ -52,6 +52,10 @@ const (
 
 	// EnvVarPrefix is the prefix for environment variables to override bundle helm variables
 	EnvVarPrefix = "UDS_"
+
+	ZarfPackageNameAnnotation = "zarf.package.name"
+
+	UDSPackageNameAnnotation = "uds.package.name"
```
UncleGedd, line 58: linter things

src/pkg/bundle/bundle.go
```diff
@@ -81,7 +82,37 @@ func Create(b *Bundler, signature []byte) error {
...
+						layerDesc.Annotations[config.ZarfPackageNameAnnotation] = zarfPackageName
+					}
+
 					rootManifest.Layers = append(rootManifest.Layers, layerDesc)
+					zarfLayerAdded = true
+				}
+				if layerDesc.MediaType == "application/vnd.zarf.config.v1+json" {
```
UncleGedd, line 99: This block can be refactored in a couple ways:
- Zarf has a constant for that zarf.config media type
- have a more specific variable name than `data`
- we can use Zarf's ConfigPartial on `data` to avoid the generic types

src/pkg/bundle/bundle.go
```diff
@@ -52,7 +52,8 @@ func Create(b *Bundler, signature []byte) error {
 	// grab all Zarf pkgs from OCI and put blobs in OCI store
 	for i, pkg := range bundle.Packages {
 		fetchSpinner := message.NewProgressSpinner("Fetching package %s", pkg.Name)
-
+		zarfPackageName := ""
+		zarfLayerAdded := false
```
UncleGedd, line 56: Let's be more specific with this var name, yes it's a Zarf layer, but more specifically it's a descriptor pointing to the Zarf package's root manifest

src/pkg/bundle/bundle.go
```diff
@@ -110,6 +141,11 @@ func Create(b *Bundler, signature []byte) error {
 
 			zarfPkgDesc, err := localBundler.ToBundle(store, zarfPkg, artifactPathMap, b.tmp, pkgTmp)
 
+			// add package name annotations, for local zarf packages, these names will be the same
```
UncleGedd, line 144: What do you think of making a follow-up ticket for handling UDS pkg name vs Zarf pkg name for local bundle?

src/pkg/bundle/bundle.go
```diff
@@ -221,6 +257,16 @@ func CreateAndPublish(remoteDst *oci.OrasRemote, bundle *types.UDSBundle, signat
 		// ensure media type is a Zarf blob and append to bundle root manifest
 		zarfManifestDesc.MediaType = oci.ZarfLayerMediaTypeBlob
 		message.Debugf("Pushed %s sub-manifest into %s: %s", url, dstRef, message.JSONValue(zarfManifestDesc))
+
+		// add package name annotations to zarf manifest
+		zarfYamlPackage, err := remoteBundler.RemoteSrc.FetchZarfYAML()
```
UncleGedd, line 262: Based on the var name, I'm not sure what `zarfYamlPackage` is. Is it the `zarf.yaml`?

src/pkg/bundle/tarball.go
```diff
@@ -346,3 +346,22 @@ func (tp *tarballBundleProvider) PublishBundle(bundle types.UDSBundle, remote *o
...
+func (tp *tarballBundleProvider) ZarfPackageNameMap() (map[string]string, error) {
+	if err := tp.getBundleManifest(); err != nil {
+		return nil, err
+	}
+
+	nameMap := make(map[string]string)
+	for _, layer := range tp.manifest.Layers {
+		if layer.MediaType == oci.ZarfLayerMediaTypeBlob {
+			// only the uds bundle layer will have AnnotationTitle set
+			rel := layer.Annotations[ocispec.AnnotationTitle]
```
UncleGedd, line 360: Maybe think through refactoring this logic to more specific, something like:
`if layer.Annotations[ocispec.AnnotationTile] != config.BundleYAML`

IMO this makes it clear that we're dealing with Zarf image manifest descs, not the `uds-bundle.yaml`

src/pkg/bundle/remote.go
```diff
@@ -336,3 +336,33 @@ func EnsureOCIPrefix(source string) string {
 	}
 	return source
 }
+
+// get zarf package name mappings from oci provider
+func (op *ociProvider) ZarfPackageNameMap() (map[string]string, error) {
```
UncleGedd, line 341: nice work with this!

src/pkg/bundle/remote.go
```diff
@@ -336,3 +336,33 @@ func EnsureOCIPrefix(source string) string {
 	}
 	return source
 }
+
+// get zarf package name mappings from oci provider
```
UncleGedd, line 340: nit: need to include `ZarfPackageNameMap` in the function's comment (our linter should really be catching this stuff)

UncleGedd:
Great work on this PR! I think at this point we are ~90% there. One thing I didn't see is the same logic applied to the `uds create -o` path?

src/pkg/bundle/bundle.go
decleaver: going with "zarfRootLayerAdded"

src/pkg/bundle/bundle.go
decleaver: yup, updated to "zarfYamlFile"

src/pkg/bundle/bundle.go
```diff
@@ -81,7 +82,37 @@ func Create(b *Bundler, signature []byte) error {
...
+					}
+					var data map[string]interface{}
+					err = json.Unmarshal(jsonData, &data)
+					if err != nil {
+						return err
+					}
+					zarfPackageName = data["annotations"].(map[string]interface{})["org.opencontainers.image.title"].(string)
+					// Check if zarf image manifest has been added to root manifest already, if so add zarfPackageName annotation
+					if zarfLayerAdded {
+						rootManifest.Layers = append(rootManifest.Layers, layerDesc)
```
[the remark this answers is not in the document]
decleaver: comments added to explain the logic due to not being able to guarantee the order the zarf layers get processed.

src/pkg/bundle/tarball.go
decleaver: agree. updated

decleaver:
I believe this block of [code](https://github.com/defenseunicorns/uds-cli/pull/342/files#diff-43aff4de6dc735f8fb96fc0c013e6a8196df5f397bad0d23fcf9348f9229f492R263)  will handle create -o but will test out to confirm

decleaver:
confirmed and added additional testing

UncleGedd (approved):
Great work on this! I also pulled it down ran it through some scenarious where the pkg name in the `uds-bundle.yaml` changes multiple times, works great!

Issue #344, opened 2024-01-22
Excessive Copy Operations for Variables


Racer159:
### Describe what should be investigated or refactored

There seems to be an excessive number of copy operations for variables being placed into the package variable map in the bundler's loadVariables function.

### Links to any relevant code

https://github.com/defenseunicorns/uds-cli/blob/edf67623f7513cddfb010340e399bacc9bfb5fd8/src/pkg/bundle/deploy.go#L197

### Additional context

This should be able to be a single map that is copied into without all of the intermediate maps. This isn't a _huge_ problem but there could be cases when larger blocks of text are duplicated in memory that could cost a few unnecessary megabytes.

PR #347, opened 2024-01-23
feat: actionable tasks

### LINKED ISSUE #218: Tab complete

Would be nice to be able to use the tab key to complete a task name when using `uds run`. 

- **Given** a task named `something-unique`
- **When** I press the "tab" key on the keyboard when I type `someth`
- **Then** the taskname will be completed in my terminal

Noxsios:
## Description

Implements a similar behavior as `with` and `inputs` from GitHub actions.

Closes #218 

## Type of change

## Checklist before merging

- [x] [Contributor Guide Steps](https://github.com/defenseunicorns/uds-cli/blob/main/CONTRIBUTING.md)(https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md#submitting-a-pull-request) followed


corang:
closes #218

src/cmd/run.go
```diff
@@ -25,6 +25,24 @@ var runCmd = &cobra.Command{
 	Use:   "run [ TASK NAME ]",
 	Short: "run a task",
 	Long:  `run a task from an tasks file`,
+	ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
UncleGedd, line 28: nice!

src/pkg/runner/runner.go
```diff
@@ -168,14 +186,49 @@ func (r *Runner) getTask(taskName string) (types.Task, error) {
 	return types.Task{}, fmt.Errorf("task name %s not found", taskName)
 }
 
+func mergeEnv(env1, env2 []string) []string {
```
UncleGedd, line 189: Can you toss a comment on this fn indicating that it merges `env1` into`env2`?

src/pkg/runner/runner.go
```diff
@@ -168,14 +186,49 @@ func (r *Runner) getTask(taskName string) (types.Task, error) {
...
+
 func (r *Runner) executeTask(task types.Task) error {
 	if len(task.Files) > 0 {
 		if err := r.placeFiles(task.Files); err != nil {
 			return err
 		}
 	}
 
+	defaultEnv := []string{}
+	for name, ip := range task.Inputs {
```
UncleGedd, line 221: nit: `ip` is Internet Protocol, rename to something that isn't so overloaded (`p` for param, spell out `param` or `inputParam`, etc)

src/pkg/runner/runner.go
```diff
@@ -168,14 +186,49 @@ func (r *Runner) getTask(taskName string) (types.Task, error) {
...
+		d := ip.Default
+		if d == "" {
+			continue
+		}
+		defaultEnv = append(defaultEnv, formatEnvVar(name, d))
+	}
+
 	for _, action := range task.Actions {
+		action.Env = mergeEnv(action.Env, defaultEnv)
```
UncleGedd, line 230: No need to do this if `defaultEnv` is empty

src/test/tasks/composable-tasks.yaml
```diff
@@ -0,0 +1,96 @@
+tasks:
```
UncleGedd, line 1: I see in the tests how we're checking for the various success vs error outputs. I think these tests would give me more confidence for future extensions if they checked for the value of the input being echo'd instead of just "success"

UncleGedd:
Very nice work fellas! I left a few comments and I'll also go through this with @decleaver , but overall I like the design and am excited to see what reusable tasks are created. Couple more thoughts:

1. At this point go ahead and add some docs, the `INPUT_` syntax + `env:` block will be confusing to newcomers writing composable actions. Recommend writing multiple examples and discuss the limitations of the inputs (ie. can't be complex values)

1. Have you tested this with remote composable tasks? Recommend chatting with @mjnagel and getting a PR up for [uds-common-tasks](https://github.com/defenseunicorns/uds-common-tasks)

src/pkg/runner/runner.go
Noxsios: was just about to do that as this function is very ambiguous

src/pkg/runner/runner.go
UncleGedd: Is it a noop? I read this as `mergeEnv` will loop through `action.Env` copy it into `defaultEnv` and then return it
```
func mergeEnv(env1, env2 []string) []string {
	for _, s1 := range env1 {
		replaced := false
		for j, s2 := range env2 {
			if strings.Split(s1, "=")[0] == strings.Split(s2, "=")[0] {
				env2[j] = s1
				replaced = true
			}
		}
		if !replaced {
			env2 = append(env2, s1)
		}
	}
	return env2
}
```

Noxsios: true, noop was not the correct verbiage.

Yes, if `defaultEnv` is empty then this action boils down to `action.Env = action.Env` (after looping through `defaultEnv`, which is `nil` and creating a new slice in memory containing `action.Env`).

The speed cost of this operation is mainly negligible. So imo there is no need to add an additional logic statement to constrict the conditions it runs under.

```
if len(defaultEnv) > 0 {
 action.Env = mergeEnv(action.Env, defaultEnv)
}
```

does not actually perform any (logic wise) operations differently than straight calling:

```
action.Env = mergeEnv(action.Env, defaultEnv)
```

UncleGedd: > The speed cost of this operation is mainly negligible

I'm slightly trolling bc of @Racer159 's issue https://github.com/defenseunicorns/uds-cli/issues/344 😆. No need to change though, I also believe it's negligible

src/test/common.go
```diff
@@ -55,20 +55,6 @@ func (e2e *UDSE2ETest) UDS(args ...string) (string, string, error) {
 	return exec.CmdWithContext(context.TODO(), exec.PrintCfg(), e2e.UDSBinPath, args...)
 }
 
-// RunTasksWithFile executes a UDS run command. with the --file flag set to the test/tasks.yaml file.
```
UncleGedd, line 58: YASSSS

src/pkg/runner/runner.go
```diff
@@ -30,29 +32,69 @@ type Runner struct {
...
+		}
+
+		task.Actions, err = templateTaskActionsWithInputs(task, withMap)
+		if err != nil {
+			return err
+		}
+
+		if err := validateActionableTaskCall(task.Name, task.Inputs, withMap); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
decleaver: good catch

Noxsios:
This change will also need an ADR.

UncleGedd:
> This change will also need an ADR.

Tracking an overall runner ADR with https://github.com/defenseunicorns/uds-cli/issues/345. Curious though, why do you think this feature in particular needs an ADR?

Noxsios:
> > This change will also need an ADR.
> 
> Tracking an overall runner ADR with #345. Curious though, why do you think this feature in particular needs an ADR?

While this change itself does not need an entire ADR created for it, imo if there _was_ an ADR for runner than this PR would required an amendment to that ADR as this adds in a lot of further behavior than simple actions:

- `$UDS_ENV`
- Go templating actions
- `inputs/with` + interactions w/ env variables + templating
- Input validation + deprecated messaging

An initial ADR for runner should have been made when the initial runner boilerplate was brought over from `wren`.

UncleGedd (approved):
Reviewed and tested, docs to come in a subsequent PR. Thanks @Noxsios and @corang for your work on this!

decleaver (approved):
Went through with Gedd, added some comments and tests to get those warm fuzzy feelings

PR #349, opened 2024-01-24
fix: allow comma separated string values for helm overrides

### LINKED ISSUE #348: Error passing variable where value is a string of comma separated values to helm chart
### Environment

uds 0.6.2

### Steps to reproduce
Can reproduce locally by updating the ui_color variable in https://github.com/defenseunicorns/uds-cli/blob/main/src/test/bundles/07-helm-overrides/uds-config.yaml to be "green, yellow" and running the TestBundleWithHelmOverrides [test](https://github.com/defenseunicorns/uds-cli/blob/main/src/test/e2e/variable_test.go#L50)

### Expected result
Doesn't fail

### Actual Result
  ERROR:  Failed to deploy bundle: failed parsing --set data: key " yellow" has no value
  
### Additional Context
@mjnagel was seeing this error with following:

```
variables:
  uds-core:
    passthrough_eip_allocations: "eipalloc-0aac7783b81d08b1d,eipalloc-076ea2ba6d5465012,eipalloc-02a25ee2cf1255996"
```
I have a config file that looks like this ^...but on deploy am seeing this error:

```
    ERROR:  Failed to deploy bundle: failed parsing --set data: key "eipalloc-076ea2ba6d5465012" has no value                                                                                                             
                  (cannot end with ,)                                                                                                                                                                                           
     ERROR:  Failed to run action: command "Deploy the base UDS bundle" timed out after 0 seconds    
```

decleaver:
## Description

Fixes error passing variable where value is a string of comma separated values to helm chart

## Related Issue

Fixes #348 

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/bundle/deploy.go
+++ b/src/pkg/bundle/deploy.go
@@ -289,6 +289,10 @@ func (b *Bundler) loadChartOverrides(pkg types.Package) (ZarfOverrideMap, error)
 
 		// Loop through each chart in the component
 		for chartName, chart := range component {
+			//escape commas (with \\) in values so helm v3 can process them
+			for i, value := range chart.Values {
+				chart.Values[i] = strings.ReplaceAll(value, ",", "\\,")
+			}
 			// Merge the chart values with Helm
 			data, err := chart.MergeValues(getter.Providers{})
 			if err != nil {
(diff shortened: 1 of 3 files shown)


UncleGedd (approved):
Nice work! I think this work may need to be extended in the future. As in, today we're dealing with commas, but tomorrow we need to deal with other special chars. No need to address that today though, but something to keep in mind

PR #354, opened 2024-01-24
feat: uds deploy --set flag for package variables

### LINKED ISSUE #54: Support setting package variables
### Currently I believe there are two ways to set variables for zarf packages included in a bundle:

1.  Via the uds-config.yaml 
```
bundle:
  deploy:
    zarf-packages:
      dubbd:
        set:
          FOO: "bar"
```

2. [Via imports/exports](https://github.com/defenseunicorns/uds-cli/blob/main/src/test/packages/02-simple-vars/uds-bundle.yaml#L7-L18)

### I would propose that we support two additional methods of specifying variables
* Accept `--set` argument to bundle deploy to such that you can do something like  `uds bundle deploy --set dubbd.FOO=bar`
* Accept defining variables in the uds-bundle.yaml similar to what is possible with a zarf.yaml:
```
kind: UDSBundle
metadata:
  name: uds-core
  description: uds core
  version: 0.0.1

zarf-packages:
  - name: dubbd
    repository: oci://ghcr.io/defenseunicorns/packages/dubbd
    variables:
      - name: FOO
        default: bar
```

decleaver:
## Description

Allows you to set zarf package variables by using the `--set` flag when running the `uds deploy` command

## Related Issue

Fixes #54 

## Type of change

## Checklist before merging


src/test/e2e/commands_test.go
```diff
@@ -94,6 +94,13 @@ func deploy(t *testing.T, tarballPath string) (stdout string, stderr string) {
 	return stdout, stderr
 }
 
+func deployCustom(t *testing.T, input string) (stdout string, stderr string) {
```
UncleGedd, line 97: Nice! I like this pattern, it can extended beyond just deploying. What do you think of renaming it to something like `runCmd`?

src/test/e2e/variable_test.go
```diff
@@ -45,6 +45,10 @@ func TestBundleVariables(t *testing.T) {
 	require.Contains(t, stderr, "This fun-fact demonstrates precedence: The Red Dragon is the national symbol of Wales")
 	require.Contains(t, stderr, "shared var in output-var pkg: burning.boats")
 	require.Contains(t, stderr, "shared var in receive-var pkg: burning.boats")
+
+	_, stderr = deployCustom(t, "deploy "+bundlePath+" --set ANIMAL=Longhorns --set COUNTRY=Texas --confirm -l=debug")
```
UncleGedd, line 49: What is `+bundlePath+"?

UncleGedd:
Curious what @zachariahmiller @rjferguson21 @mjnagel (and others!) think. The way this feature is currently implemented, you can do `uds deploy --set FOO=bar` where `FOO` is a non-specific Zarf pkg variable. Do any of you have thoughts on explicitly identifying the Zarf pkg that the variable belongs to (ie. something like `uds deploy --set myzarfpkg.FOO=bar`) ?

zachariahmiller:
I'm primarily using --set right now in uds run vs deploy. https://github.com/defenseunicorns/uds-core/blob/main/.github%2Fworkflows%2Ftest.yaml#L63

For deploy I think the above sounds fine as long as it isn't confusing and the existing functionality still works. If FOO without a path is a variable at the bundle level (overrides or bundle level) and with a path (package-a. FOO) it's scoped only to the package I think that's fine in terms of functionality. 

I haven't really thought it through from a ux perspective though so I would like to hear from others if that's confusing somehow or if there is a better way.

src/cmd/uds.go
```diff
@@ -221,6 +222,7 @@ func init() {
 
 	// deploy cmd flags
 	rootCmd.AddCommand(deployCmd)
+	deployCmd.Flags().StringToStringVar(&bundleCfg.DeployOpts.SetVariables, "set", v.GetStringMapString(common.VPkgDeploySet), lang.CmdBundleDeployFlagSet)
```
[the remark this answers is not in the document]
decleaver: yeah, i need to dig in a little more and see/understand what viper is doing with this

src/test/e2e/variable_test.go
decleaver: the bundle tarball path, updated the name to that (bundleTarballPath) for clarity

UncleGedd (approved):
🚀

PR #386, opened 2024-02-02
feat: add envPath key to runner

UncleGedd:
## Description
- Adds an `envPath` key to the UDS Runner to load env vars
- Removes the concept of `UDS_ENV` pointing to a file because it potentially leaves that file on the user's system (potentially leaving secrets)


decleaver (approved):
walked through together, this is a lot more clear

PR #387, opened 2024-02-02
chore: only use camelCase in yaml

UncleGedd:
## Description
Swap `optional-components` to `optionalComponents`. Ensure we are only using camelCase in the `uds-bundle.yaml`


decleaver (approved):
camelCase! lgtm

PR #393, opened 2024-02-05
chore: refactors bundle create path

UncleGedd:
## Description

- Refactors bundle create path
- Adds arch diagram


decleaver (changes requested):
looks good, heading in the right direction! Looks like the GH validation picked up some minor things, and just need to add that architecture diagram.

decleaver (approved):

Issue #397, opened 2024-02-06
Bundles that deploy a cluster no longer work as of 0.8.0


mjnagel:
### Environment

Device and OS: MBP
App version: 0.8.0
Kubernetes distro being used: k3d
Other:

### Steps to reproduce

1. `uds deploy oci://defenseunicorns/uds/bundles/k3d-core-istio-dev:0.10.0-amd64`

### Expected result

Bundle deploys as previously.

### Actual Result

Bundle errors with:
```
     ERROR:  Failed to deploy bundle: failed to connect to k8s cluster: invalid configuration: no configuration
             has been provided, try setting KUBERNETES_MASTER environment variable
```

### Severity/Priority

Medium/High, this is currently blocking updates to 0.8.0 for test cluster deploys on a few repos.

### Additional Context

I believe the addition of [returning the error here](https://github.com/defenseunicorns/uds-cli/blob/v0.8.0/src/pkg/bundle/remote.go#L309-L311) is the origin of this new behavior, previously the error wasn't returned.

PR #398, opened 2024-02-06
fix: don't error out when a cluster isn't available

UncleGedd:
## Description
Fixes bug that prevented a UDS Bundle from deploying without a cluster

## Related Issue

Fixes https://github.com/defenseunicorns/uds-cli/issues/397

## Type of change


mikevanhemert (approved):
🚀

PR #405, opened 2024-02-07
chore: ignore new hidden folder used to store release assets downloaded using the gh cli

timblaktu:
ignore new hidden folder used to store release assets downloaded using the gh cli inside a local repo. This is just a convenience for devs that use gh cli in this way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 .env.*
 .idea/
 .image-cache/
+.release-downloads/
 .scratch/
 .vscode/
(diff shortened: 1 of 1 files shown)


zachariahmiller:
fixed the title check failure so ci would run... Not that you changed any code.

UncleGedd (approved):
Lots of complexity with this PR, recommend breaking into smaller chuncks next time.

JK thanks @timblaktu !

PR #406, opened 2024-02-07
fix: honors UDS_ env vars

### LINKED ISSUE #403: `UDS_ARCHITECTURE` flag no longer honored as of 0.7.0
### Environment

Device and OS: MBP
App version: 0.7.0+
Kubernetes distro being used: k3d

### Steps to reproduce

The below uses UDS Core as an example:
1. Clone uds-core
2. Navigate to the k3d-istio bundle - https://github.com/defenseunicorns/uds-core/tree/main/bundles/k3d-istio
3. Run `UDS_ARCHITECTURE=amd64 uds create --confirm`
4. Run `UDS_ARCHITECTURE=arm64 uds create --confirm`

### Expected result

The bundle creation should be expected to fail because we did not build the zarf package, but the `UDS_ARCHITECTURE` should be honored and would be looking for the appropriate zarf package based on architecture.

### Actual Result

The bundle creation will fail, looking for the zarf package matching your localhost architecture (in my case this is arm64, even when using the amd64 `UDS_ARCHITECTURE` env).

### Severity/Priority

Low/Mid - the `-a` flag works in most scenarios, but using the env was beneficial for calling bundle creates inside of UDS tasks. EDIT: Validated that `-a` should work inside of UDS tasks as well.

UncleGedd:
## Related Issue

Fixes #403 

Relates to #

## Type of change

### DIFF
--- a/src/cmd/viper.go
+++ b/src/cmd/viper.go
@@ -74,9 +74,10 @@ func initViper() {
 		v.SetConfigName("uds-config")
 	}
 
-	// E.g. UDS_LOG_LEVEL=debug
+	// we replace 'OPTIONS.' because in a uds-config.yaml, the key is options.<opt>, but in the environment, it's UDS_<OPT>
+	// e.g. UDS_LOG_LEVEL=debug
 	v.SetEnvPrefix("uds")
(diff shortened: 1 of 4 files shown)


decleaver (approved):
paired on it. good to go

PR #415, opened 2024-02-09
feat: removes --with, expands variables for inputs

UncleGedd:
## Description

The `--with` flag was meant to be used for testing tasks with `inputs` from the CLI. This feature ended up being confusing for users so we'd like to remove it. We believe this will encourage better testing patterns because it forces users to test their `inputs` with actual tasks instead of through the CLI.

Furthermore, we expanded the use of runner variables to ensure they can be passed into tasks using `with`

### DIFF
--- a/src/pkg/runner/runner.go
+++ b/src/pkg/runner/runner.go
@@ -50,6 +50,7 @@ func Run(tasksFile types.TasksFile, taskName string, setVariables map[string]str
 		return err
 	}
 
+	// can't call a task directly from the CLI if it has inputs
 	if task.Inputs != nil {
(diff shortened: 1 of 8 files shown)


src/pkg/runner/runner.go
```diff
@@ -397,6 +379,11 @@ func (r *Runner) performAction(action types.Action) error {
 			return err
 		}
 
+		// template the withs with variables
```
decleaver, line 382: this is a new "feat"? assuming this means we weren't allowing  variables w/ "withs"?

UncleGedd: yes!

decleaver (approved):
looks good

PR #418, opened 2024-02-12
feat: adds Zarf schema for version of Zarf used in UDS CLI

### LINKED ISSUE #416: Include zarf schema as part of uds-cli repo/releases

In uds-package-* repos we are trying to drop the need entirely for zarf and leverage vendored zarf for all zarf calls. By doing this we hope to drop any versioning weirdness between uds and zarf. One piece that we are currently still tied to for zarf is the versioned zarf schema to add for `.vscode` folders, etc.

Ideally I would like uds-cli to include a copy of the zarf schema, matching the version of zarf that is vendored in the given uds-cli release. This way I can track uds-cli's versioning exclusively, even for the zarf schema.

Maintain zarf schema separately and maybe do a lookup with `uds zarf version` to ensure we are using the correct version.

UncleGedd:
## Description

- Adds Zarf schema to this repo using the version of Zarf that is vendored in UDS CLI

## Related Issue

Fixes #416

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-* @UncleGedd @mikevanhemert @jeff-mccoy @decleaver
+* @UncleGedd @jeff-mccoy @decleaver @mjnagel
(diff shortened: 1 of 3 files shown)


decleaver (dismissed):
:shipit:

decleaver (approved):
🚀

Issue #370, opened 2024-01-30
Ability to use package variables within a uds-bundle


brianrexrode:
### Is your feature request related to a problem? Please describe
It would nice to be able to pass variables from one zarf package to another within a uds-bundle whether they are expected ZARF_VARs or not.

I would like to use uds-core packages within my AWS EKS environments and configure components like `loki` to use AWS resources. The resources would need to be created prior to deploying uds-core and in most cases require outputs to be used as values inputs to the helm charts. What if the uds-core `loki` values do not expose ZARF_VARs needed to properly configure my deployment on AWS?

Ex. 
I've created a zarf package that deploys S3 buckets, creates associated IRSA roles and policies, as well as defines the AWS Region I'm deploying to. Those outputs need to be passed along to the uds-core `loki` component's helm values. Each AWS environment I deploy to will/may have different values for each resource. 

I can export these ZARF_VARs from my custom s3-irsa-iac zarf package within my uds-bundle and import them to the uds-core package but it's my understanding they are useless if the uds-core package is not expecting the same ZARF_VARs
```
    exports:
      - name: LOKI_S3_BUCKET
      - name: LOKI_S3_IRSA_ROLE_ARN
      - name: LOKI_S3_AWS_REGION

    imports:
      - name: LOKI_S3_BUCKET
        package: s3-irsa-iac
      - name: LOKI_S3_IRSA_ROLE_ARN
        package: s3-irsa-iac
      - name: LOKI_S3_AWS_REGION
        package: s3-irsa-iac
```
I
### Describe the solution you'd like
It would be nice to have the ability to use those exported ZARF_VARs within the uds-bundle as regular VARs.
Ex.
```
      overrides:
          loki:
          values:
            - path: objectStorage.endpoint
              value: https://s3.amazonaws.com
          variables:
            - path: objectStorage.region
              value: ${LOKI_S3_AWS_REGION}
            - path: objectStorage.bucketName
              value: ${LOKI_S3_BUCKET}
```
- **Given** a state
- **When** an action is taken
- **Then** something happens

### Describe alternatives you've considered
Hardcoding the values I'm trying to override
(optional) A clear and concise description of any alternative solutions or features you've considered.

PR #423, opened 2024-02-13
feat: imported vars as override values

decleaver:
## Description

This pr allows you to use imported variables as override values in your uds-bundle.yaml

## Related Issue

Relates to #370 

## Type of change

## Checklist before merging


src/pkg/bundle/deploy.go
```diff
@@ -409,6 +410,21 @@ func addOverrideValue(overrides map[string]map[string]*values.Options, component
 		val := fmt.Sprintf("%s=%s", valuePath, j)
 		overrides[component][chart].JSONValues = append(overrides[component][chart].JSONValues, val)
 	default:
+		// Check for any templated variables if pkgVars set
```
naveensrinivasan, line 413: 1. Optimize Regular Expression Compilation:
Currently, the regular expression is compiled within the executed block of code. This is efficient for occasional use, but if this function is called frequently, it might lead to unnecessary overhead. A common practice is to compile regular expressions once and reuse them throughout the application. This can be achieved by declaring the compiled regex outside of the function scope. This change could improve the performance of your code.

```go
// Consider moving this outside the function to compile it only once
   var templatedVarRegex = regexp.MustCompile(`\${([^}]+)}`)
 ```
 
 2. Enhance Handling of Multiple Templated Variables:
The current implementation replaces the first occurrence of a templated variable. If a string contains multiple templated variables, it would be beneficial to replace all of them. This adjustment ensures that all templated variables are processed, making the function more versatile.

```go
   // Use ReplaceAllStringFunc to handle all occurrences of templated variables
   replacedValue := templatedVarRegex.ReplaceAllStringFunc(templatedVariable, func(match string) string {
       variableName := templatedVarRegex.FindStringSubmatch(match)[1]
       if varValue, ok := pkgVars[variableName]; ok {
           return varValue
       }
       // Consider handling missing variables more explicitly
       return fmt.Sprintf("${%s_not_found}", variableName)
   })
   value = replacedValue
  ```

File: docs/overrides.md [hunk 2]
```diff
@@ -130,6 +130,7 @@ The `value` is the value to set at the `path`. Values can be simple values such
               value:
                 customAnnotation: "customValue"
 ```
+If [importing](../README.md#importingexporting-variables) a variable from another package, that variable can also be used to set a value, using the template syntax `${...}`
```
UncleGedd, line 133: Can we add an explicit example of what this looks like?

src/test/bundles/12-exported-pkg-vars/uds-bundle.yaml
```diff
@@ -0,0 +1,28 @@
...
+        package: output-var
+
+    overrides:
+      podinfo-component:
+        unicorn-podinfo:
+          values:
+            - path: "podinfo.replicaCount"
+              value: 1
+            - path: "podinfo.ui.color"
+              value: ${COLOR}
```
UncleGedd, line 28: Can we also test the edge case of `${FOO} - ${BAR}` <- or something to that effect

mjnagel, line 28: Is there a reason this wouldn't be a `variable` with name `color` instead of `value`? This definitely works but feels a bit weird to me since I would typically think of `values` being "hardcoded" in a bundle and `variables` being dynamic.

UncleGedd: Good catch @mjnagel ! The vars are set at deploy time so the template probably doesn't make sense under `values`, only under `variables`

decleaver: I can switch it if that makes more sense, I just went with `values` because under `variables` it would get set under `default` since there is no value field for variables.

UncleGedd: Good point @decleaver , looks like the issue is to have this under `values` 🤔

mjnagel: After our synchronous chat I'm on board with `values`. As @UncleGedd explained it `variables` is more for things that the bundle deployer can configure. This is more in line with `values` that aren't configurable by the bundle deployer.

decleaver: After discussing, going to stick with setting in `values` vs `variables`

docs/overrides.md
decleaver: added in latest commit.

src/test/bundles/12-exported-pkg-vars/uds-bundle.yaml
decleaver: added test case for this, and additional test cases for checking object and list override values.

docs/overrides.md
```diff
@@ -130,6 +130,37 @@ The `value` is the value to set at the `path`. Values can be simple values such
               value:
                 customAnnotation: "customValue"
 ```
+If [importing](../README.md#importingexporting-variables) a variable from another package, that variable can also be used to set a value, using the template syntax `${...}`. In the example below the `COLOR` variable is being used to set the `podinfo.ui.color` value.
```
UncleGedd, line 133:
```suggestion
If using a variable that has been [exported](../README.md#importingexporting-variables) from another package, that variable can also be used to set a value, using the syntax `${...}`. In the example below the `COLOR` variable is being used to set the `podinfo.ui.color` value.
```

docs/overrides.md
```diff
@@ -130,6 +130,37 @@ The `value` is the value to set at the `path`. Values can be simple values such
...
+    ref: 0.0.1
+    exports:
+      - name: COLOR
+
+  - name: helm-overrides
+    path: "../../packages/helm"
+    ref: 0.0.1
+    imports:
+      - name: COLOR
+        package: output-var
```
UncleGedd, line 153:
```suggestion
```

UncleGedd (approved):
🚀

Issue #422, opened 2024-02-13
Import multiple variables from the same package without defining the pkg each time


brianrexrode:
### Is your feature request related to a problem? Please describe

It would be nice to have the ability to import multiple variables from the same package without having to define the package for each variable.

### Describe the solution you'd like
```
packages:
  - name: s3-irsa-iac
    exports:
      - name: LOKI_S3_BUCKET
      - name: LOKI_S3_IRSA_ROLE_ARN
      - name: LOKI_S3_AWS_REGION
      - name: VELERO_S3_BUCKET
      - name: VELERO_S3_IRSA_ROLE_ARN
      - name: VELERO_S3_AWS_REGION
  - name: uds-core
    imports:
      - package: s3-irsa-iac
        name: LOKI_S3_BUCKET
        name: LOKI_S3_IRSA_ROLE_ARN
        name: LOKI_S3_AWS_REGION
        name: VELERO_S3_BUCKET
        name: VELERO_S3_IRSA_ROLE_ARN
        name: VELERO_S3_AWS_REGION
```

### Describe alternatives you've considered

(optional) A clear and concise description of any alternative solutions or features you've considered.

UncleGedd:
@brianrexrode in the latest release of UDS CLI we changed the behavior of `exports` such that all `export`ed variables are global by default. Does that change resolve this issue?

PR #428, opened 2024-02-14
feat: import all vars exported from package

decleaver:
## Description

This pr allows you to import all the exported vars from a package by just providing the package name, so you don't need to manually import each individual exported var.

## Related Issue

Relates to #422 

## Type of change

## Checklist before merging

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ packages:
         package: output-var
 ```
 
-Variables that you want to make available to other package are in the `export` block of the Zarf package to export a variable from. By default, all exported variables are available to all of the packages in a bundle. To have another package ingest a specific exported variable, use the `imports` key to name both the `variable` and `package` that the variable is exported from, like in the example above.
+Variables that you want to make available to other packages are in the `export` block of the Zarf package to export a variable from. By default, all exported variables are available to all of the packages in a bundle. To have another package ingest a specific exported variable, like in the case of variable name collisions, use the `imports` key to name both the `variable` and `package` that the variable is exported from, like in the example above.
 
 In the example above, the `OUTPUT` variable is created as part of a Zarf Action in the [output-var](src/test/packages/zarf/no-cluster/output-var) package, and the [receive-var](src/test/packages/zarf/no-cluster/receive-var) package expects a variable called `OUTPUT`.
 
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ packages:
         package: output-var
 ```
 
-Variables that you want to make available to other package are in the `export` block of the Zarf package to export a variable from. To have another package ingest an exported variable, use the `imports` key to name both the `variable` and `package` that the variable is exported from. You can also import all of the variables for a package by omitting the variable name and just providing the package name.
+Variables that you want to make available to other package are in the `export` block of the Zarf package to export a variable from. By default, all exported variables are available to all of the packages in a bundle. To have another package ingest a specific exported variable, use the `imports` key to name both the `variable` and `package` that the variable is exported from, like in the example above.
 
 In the example above, the `OUTPUT` variable is created as part of a Zarf Action in the [output-var](src/test/packages/zarf/no-cluster/output-var) package, and the [receive-var](src/test/packages/zarf/no-cluster/receive-var) package expects a variable called `OUTPUT`.
 
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ packages:
         package: output-var
 ```
 
-Variables that you want to make available to other package are in the `export` block of the Zarf package to export a variable from. To have another package ingest an exported variable, use the `imports` key to name both the `variable` and `package` that the variable is exported from.
+Variables that you want to make available to other package are in the `export` block of the Zarf package to export a variable from. To have another package ingest an exported variable, use the `imports` key to name both the `variable` and `package` that the variable is exported from. You can also import all of the variables for a package by omitting the variable name and just providing the package name.
 
 In the example above, the `OUTPUT` variable is created as part of a Zarf Action in the [output-var](src/test/packages/zarf/no-cluster/output-var) package, and the [receive-var](src/test/packages/zarf/no-cluster/receive-var) package expects a variable called `OUTPUT`.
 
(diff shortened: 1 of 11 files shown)


decleaver:
Moving back to draft, looking into making package exports globally available by default, but brainstorming on how to handle potential name collisions.

src/pkg/bundle/common.go
```diff
@@ -327,10 +327,22 @@ func validateBundleVars(packages []types.Package) error {
 		// ensure imports have a matching export
 		if pkg.Imports != nil {
 			for _, v := range pkg.Imports {
-				if _, ok := exports[v.Name]; ok && v.Package == exports[v.Name] {
-					continue
+				// check import names if they are set, otherwise importing all the variables for the given package
```
naveensrinivasan, line 330: As a newcomer to the project, I've been familiarizing myself with the codebase and noticed a section in the import package validation logic that caught my attention. Specifically, it pertains to the logic block that validates import packages without specified variable names against exported packages. My understanding is that the current implementation might prematurely flag an import package as invalid if the first exported package name checked does not match the import package name, without considering the rest of the exported packages.

### Suggested Improvement:
I believe there might be an opportunity to enhance the validation process's accuracy by ensuring all exported package names are checked against the import package name before determining a mismatch. However, I acknowledge that I might not have full context or might be overlooking other considerations, so I'd appreciate any feedback or corrections.

### Proposed Changes:
```go
// check import names if they are set, otherwise importing all the variables for the given package
if v.Name != "" {
    if _, ok := exports[v.Name]; ok && v.Package == exports[v.Name] {
        continue
    }
    return fmt.Errorf("import var %s does not have a matching export", v.Name)
    // Check that the import package is valid
} else {
    // Initialize a variable to track if the package is found
    var packageFound bool
    for _, pkgName := range exports {
        if pkgName == v.Package {
            packageFound = true
            break
        }
    }
    // If the package is not found in exports, return an error
    if !packageFound {
        return fmt.Errorf("import package %s does not match any exporting package", v.Package)
    }
}
```

### Rationale:
Implementing this change could potentially make the validation logic more robust by ensuring that all exported packages are considered before concluding that an import package does not match. This could help in avoiding false negatives in the validation process and improve the overall reliability of the package validation mechanism.

### Impact:
This proposed change is intended to refine the validation logic without adversely affecting existing functionalities. It aims to enhance the accuracy of the package validation process.

As I'm new to this project, I'm aware that there might be aspects I'm not fully aware of or additional context that I might be missing. Therefore, I'm very open to feedback, corrections, or any insights that could help me understand this better. Thank you for considering my suggestion, and I look forward to any discussions this might prompt.

decleaver: Good catch! will be updated in next commit. Thanks @naveensrinivasan !

decleaver: So we actually got rid of the above code. We decided not to support importing variables by just specifying the package name they are being exported from. We are going to make all exported variables available by default, and in the case of name collisions, you can import variables from specific packages like before.

README.md
```diff
@@ -191,7 +191,7 @@ packages:
         package: output-var
 ```
 
-Variables that you want to make available to other package are in the `export` block of the Zarf package to export a variable from. To have another package ingest an exported variable, use the `imports` key to name both the `variable` and `package` that the variable is exported from.
+Variables that you want to make available to other package are in the `export` block of the Zarf package to export a variable from. By default, all exported variables are available to all of the packages in a bundle. To have another package ingest a specific exported variable, use the `imports` key to name both the `variable` and `package` that the variable is exported from, like in the example above.
```
UncleGedd, line 194:
```suggestion
Variables that you want to make available to other packages are in the `export` block of the Zarf package to export a variable from. By default, all exported variables are available to all of the packages in a bundle. To have another package ingest a specific exported variable, like in the case of variable name collisions, use the `imports` key to name both the `variable` and `package` that the variable is exported from, like in the example above.
```

src/test/bundles/02-simple-vars/import-all/uds-bundle.yaml
```diff
@@ -0,0 +1,17 @@
+kind: UDSBundle
+metadata:
+  name: import-all
+  description: show how import all works
```
UncleGedd, line 4:
```suggestion
  description: show how global exports work
```

src/test/e2e/variable_test.go
```diff
@@ -39,7 +39,39 @@ func TestBundleVariables(t *testing.T) {
 	os.Setenv("UDS_CONFIG", filepath.Join("src/test/bundles/02-simple-vars", "uds-config.yaml"))
 
 	_, stderr := deploy(t, bundleTarballPath)
+	bundleVariablesTestChecks(t, stderr, bundleTarballPath)
+	remove(t, bundleTarballPath)
 
+	// Run same test checks but with package that isn't explicitly importing vars
```
UncleGedd, line 45: Does it make sense to put this in a new test instead of extending this one?

src/test/e2e/variable_test.go
```diff
@@ -39,7 +39,39 @@ func TestBundleVariables(t *testing.T) {
...
+	bundleDir = "src/test/bundles/02-simple-vars/import-all"
+	bundleTarballPath = filepath.Join(bundleDir, fmt.Sprintf("uds-bundle-import-all-%s-0.0.1.tar.zst", e2e.Arch))
+	createLocal(t, bundleDir, e2e.Arch)
+	createRemoteInsecure(t, bundleDir, "localhost:888", e2e.Arch)
```
UncleGedd, line 49: Is `createRemoteInsecure` necessary? Do we test that creating remotely triggers the same error?

UncleGedd:
Great work on this! Appreciate the thorough testing of all the edge cases. We should think through ways to simplify some of these new e2e tests with unit tests, or just making the e2e tests smaller and more explicit

src/test/e2e/variable_test.go
UncleGedd: will resolve in https://github.com/defenseunicorns/uds-cli/issues/449

src/test/e2e/variable_test.go
decleaver: removed in latest commit, good catch!

UncleGedd (approved):

PR #431, opened 2024-02-15
fix: use tmpdir if provided

### LINKED ISSUE #430: if creating a bundle while setting `--tmpdir <dir>`, /tmp is still used
### Environment

Device and OS: rhel-like
App version:
Kubernetes distro being used: rke2
Other:

### Steps to reproduce

1. try to create a bundle specifying `--tmpdir <dir>`
2. see it still use /tmp

### Expected result
uses the passed in directory

### Actual Result
uses /tmp

### Additional Context

From discussion with @Noxsios :

>  we need to standardize creating tmp dirs across the board. rn any caller of Zarf's utils.MakeTempDirectory is going to use the systems tmp dirs unless you pass a dir string. (There's a bunch of bad behavior w/ this in Zarf as well thinking more on it)

@justin-o12  FYSA i created this issue to track what you ran into. 

@UncleGedd @decleaver I will link more details internally.

decleaver:
## Description

This PR fixes --tmpDir being ignored on create 

## Related Issue

Fixes #430 

## Type of change

## Checklist before merging

### DIFF
--- a/src/test/e2e/bundle_test.go
+++ b/src/test/e2e/bundle_test.go
@@ -115,6 +115,10 @@ func TestBundle(t *testing.T) {
 	os.Setenv("UDS_CONFIG", filepath.Join("src/test/bundles/01-uds-bundle", "uds-config.yaml"))
 	deploy(t, bundlePath)
 	remove(t, bundlePath)
+
+	//Test create using custom tmpDir
+	runCmd(t, "create "+bundleDir+" --tmpdir ./customtmp --confirm --insecure")
+
 }
 
 func TestPackagesFlag(t *testing.T) {
--- a/src/test/e2e/bundle_test.go
+++ b/src/test/e2e/bundle_test.go
@@ -10,7 +10,9 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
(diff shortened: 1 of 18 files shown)


src/test/e2e/bundle_test.go
```diff
@@ -115,6 +115,10 @@ func TestBundle(t *testing.T) {
 	os.Setenv("UDS_CONFIG", filepath.Join("src/test/bundles/01-uds-bundle", "uds-config.yaml"))
 	deploy(t, bundlePath)
 	remove(t, bundlePath)
+
+	//Test create using custom tmpDir
+	runCmd(t, "create "+bundleDir+" --tmpdir ./customtmp --confirm --insecure")
```
UncleGedd, line 120: I may be missing it, but are we testing that `./customtmp` is actually being populated?

decleaver: negative. wasnt 100% sure on how to check that, so just testing that it doesn't blow up

decleaver: pushed less lazy testing :)

UncleGedd (approved):
🚀

mjnagel (approved):

Issue #440, opened 2024-02-19
Potential TOCTOU Vulnerability in getBundleManifest() File Handling


naveensrinivasan:
A potential Time-Of-Check to Time-Of-Use (TOCTOU) vulnerability has been identified in the `getBundleManifest` function within the tarballBundleProvider, specifically within the `tarball.go` file. This vulnerability arises because the files `indexPath` and `manifestPath` are marked for deletion immediately after their creation and before the processing of their contents, creating an opportunity for an attacker to replace these files with malicious versions between their verification and use.

https://research.nccgroup.com/wp-content/uploads/2021/09/TOCTOU_whitepaper.pdf
### Attack Scenario
Initial State: The application creates temporary files (indexPath and manifestPath) to store data temporarily. These files are marked for deletion (using defer os.Remove(filePath)) after their contents are processed.

Exploitation Window: Between the creation (and subsequent check) of these files and their actual usage by the application, there's a brief time window. This is due to the deferred deletion, which does not occur until the function returns. During this window, the files are vulnerable to tampering.

### Attack Execution:

An attacker, with access to the file system where the application is running, monitors the creation of these temporary files.
Once the application creates indexPath or manifestPath, the attacker quickly replaces one or both of these files with malicious versions. The content of these malicious files could be crafted to exploit vulnerabilities in the application's file processing logic or to execute arbitrary code.
Impact:

The application proceeds to use the tampered files, under the assumption that they have not been altered since their creation and initial check.
Depending on the nature of the tampered content and the application's functionality, this could lead to various adverse outcomes, such as data corruption, unauthorized data disclosure, execution of arbitrary code, or complete system compromise.

```go
indexPath := filepath.Join(tp.dst, "index.json")
defer os.Remove(indexPath)
...
manifestPath := filepath.Join(tp.dst, manifestRelativePath)
defer os.Remove(manifestPath)
```

The potential exploitation of this vulnerability could allow for the compromise of the application's integrity or enable the execution of arbitrary code if an attacker has the capability to manipulate the file system.

Mitigation strategies may include:
1. Assessing the necessity of deleting these files immediately after processing. If immediate deletion is not critical, it might be safer to remove the files only after their contents have been securely processed.
2. Implementing integrity checks for the files just before they are used. This could be achieved by verifying the files against pre-determined good signatures or checksums.
3. Employing atomic operations for file handling to minimize the risk associated with this vulnerability.

[...]

PR #443, opened 2024-02-21
fix: toctou for files

naveensrinivasan:
- Fixes https://github.com/defenseunicorns/uds-cli/issues/440

## Description

toctou for files

## Related Issue

Fixes #
 https://github.com/defenseunicorns/uds-cli/issues/440

## Type of change

## Checklist before merging


naveensrinivasan:
> I'm not sure this actually prevents any attacks on the files in the temp dir, it just makes it much harder. The attacker would have to maliciously change the file after verification, and restore it before the second verification.
> 
> There are even some ways the attacker could win these races pretty easily. Like, do an early replacement of index.json with a symlink (before any reads or verification of it). And if they pointed into an attacker-controlled file on a FUSE volume, they could control what individual reads return.
> 
> I think the right way to do tmpdir security is probably just UNIX file permissions? Create a single randomly named subdirectory of /tmp (or wherever the OS specifies) and restrict permissions to something like 0700. If the user uses --tmpdir, it's on them to make sure that's a secure location.

I appreciate the detailed feedback and wasn't unaware of the FUSE volume issue. Thank you! I have implemented the changes. 

Please look at it and let me know.

bburky (changes requested):
uds-cli has a `--tmpdir` flag, this path should be used if configured. Only use `os.TempDir()` if it is not set.

I still don't think checksums add any security. An attacker modifying files be prevented by file permissions. Checksums at best can only detect changes, but are still prone to race conditions and can be bypassed by an attacker. Secure permissions should prevent the attacker entirely.

bburky (approved):
LGTM

UncleGedd (approved):
Looks great. Appreciate the additional feedback @bburky 🚀

PR #444, opened 2024-02-21
fix: gosec lint issues for the pkg

naveensrinivasan:
## Description

- Fixes gosec lint issues for the pkg

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/bundler/fetcher/local.go
+++ b/src/pkg/bundler/fetcher/local.go
@@ -68,12 +68,7 @@ func (f *localFetcher) GetPkgMetadata() (zarfTypes.ZarfPackage, error) {
 	if err != nil {
 		return zarfTypes.ZarfPackage{}, err
 	}
-	defer func(path string) {
-		err := os.RemoveAll(path)
-		if err != nil {
-
-		}
-	}(tmpDir)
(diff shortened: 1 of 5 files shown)


src/pkg/bundler/fetcher/local.go
```diff
@@ -69,10 +69,7 @@ func (f *localFetcher) GetPkgMetadata() (zarfTypes.ZarfPackage, error) {
 		return zarfTypes.ZarfPackage{}, err
 	}
 	defer func(path string) {
-		err := os.RemoveAll(path)
-		if err != nil {
-
-		}
+		os.RemoveAll(path) //nolint:errcheck
```
UncleGedd, line 72: The purpose of wrapping `os.RemoveAll` in `defer` was to do the error check as identified by the linter. But I agree it's unnecessary! What do you think of removing the outer wrapping entirely and just doing:
```go
defer os.RemoveAll(path) //nolint:errcheck
```

naveensrinivasan: I agree. Took care of it. Thanks

UncleGedd (approved):
Thanks for this!

PR #453, opened 2024-02-22
fix: relative paths for bundle create

decleaver:
## Description

fixes zarf package paths to be relative to the bundle directory vs the working directory

## Related Issue

Fixes #

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/bundle/common.go
+++ b/src/pkg/bundle/common.go
@@ -104,6 +104,11 @@ func (b *Bundle) ValidateBundleResources(bundle *types.UDSBundle, spinner *messa
 
 	// validate access to packages as well as components referenced in the package
 	for idx, pkg := range bundle.Packages {
+		// if package path is set, make it relative to source directory
+		if pkg.Path != "" {
+			pkg.Path = fmt.Sprintf("%s/%s", b.cfg.CreateOpts.SourceDirectory, pkg.Path)
+		}
+
 		spinner.Updatef("Validating Bundle Package: %s", pkg.Name)
 		if pkg.Name == "" {
 			return fmt.Errorf("%s is missing required field: name", pkg)
--- a/src/pkg/bundle/common.go
+++ b/src/pkg/bundle/common.go
@@ -106,7 +106,7 @@ func (b *Bundle) ValidateBundleResources(bundle *types.UDSBundle, spinner *messa
 	for idx, pkg := range bundle.Packages {
 		// if package path is set, make it relative to source directory
 		if pkg.Path != "" {
-			pkg.Path = fmt.Sprintf("%s/%s", b.cfg.CreateOpts.SourceDirectory, pkg.Path)
+			pkg.Path = filepath.Join(b.cfg.CreateOpts.SourceDirectory, pkg.Path)
 		}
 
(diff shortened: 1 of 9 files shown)


src/pkg/bundle/common.go
```diff
@@ -104,6 +104,11 @@ func (b *Bundle) ValidateBundleResources(bundle *types.UDSBundle, spinner *messa
 
 	// validate access to packages as well as components referenced in the package
 	for idx, pkg := range bundle.Packages {
+		// if package path is set, make it relative to source directory
+		if pkg.Path != "" {
+			pkg.Path = fmt.Sprintf("%s/%s", b.cfg.CreateOpts.SourceDirectory, pkg.Path)
```
UncleGedd, line 109:
```suggestion
			pkg.Path = filepath.Join(b.cfg.CreateOpts.SourceDirectory, pkg.Path)
```
Recommend using the library call instead of building out the path manually (this makes all OSs happy)

src/pkg/bundle/common.go
```diff
@@ -104,6 +104,11 @@ func (b *Bundle) ValidateBundleResources(bundle *types.UDSBundle, spinner *messa
 
 	// validate access to packages as well as components referenced in the package
 	for idx, pkg := range bundle.Packages {
+		// if package path is set, make it relative to source directory
```
naveensrinivasan, line 107: There is a potential path traversal vulnerability in the provided code snippet. The issue arises from using user-controlled input (pkg.Path) in constructing a file path without proper validation or sanitization. This can allow an attacker to craft pkg.Path in such a way that it navigates out of the intended directory (b.cfg.CreateOpts.SourceDirectory) and accesses files or directories that should be restricted.

An attacker could potentially supply a pkg.Path value like ../../etc/passwd or similar, aiming to construct a path that leads to sensitive files or directories outside of the SourceDirectory.

### Potential solution

```go
if pkg.Path != "" {
    fullPath := filepath.Join(b.cfg.CreateOpts.SourceDirectory, pkg.Path)
    // Clean and resolve to absolute paths to prevent directory traversal
    basePath, err := filepath.Abs(filepath.Clean(b.cfg.CreateOpts.SourceDirectory))
    if err != nil {
        // handle error
    }
    resolvedPath, err := filepath.Abs(filepath.Clean(fullPath))
    if err != nil {
        // handle error
    }
    // Ensure the resolved path is within the base directory
    if !strings.HasPrefix(resolvedPath, basePath) {
        // handle path traversal attempt
    }
    pkg.Path = resolvedPath
}
```

_notested or compiled_

UncleGedd: Love the idea of sanitizing the filepath especially because this is technically user input. Internally I believe the code is looking specifically for Zarf tarball (see code [here](https://github.com/defenseunicorns/uds-cli/blob/57f038b20ba6b23fdd5c5e681411b7157588983c/src/pkg/bundle/common.go#L156)). I made another [issue](https://github.com/defenseunicorns/uds-cli/issues/455) to track filepath sanitization

UncleGedd (approved):
🚀 thanks @decleaver , noting that we're tracking path sanitization in https://github.com/defenseunicorns/uds-cli/issues/455

PR #473, opened 2024-02-27
chore: refactor pull operation

### LINKED ISSUE #177: Pulling bundles is jacked (sometimes)
@decleaver and I observed a behavior where sometimes pulling a bundle results in strange errors about missing layers. We think this is because `pull`, unlike the other bundle operations, occurs in the cache directory as opposed to a tmp dir. 

```[tasklist]
### Tasks
- [x] Decouple the zarf-cache entirely from UDS CLI
- [x] Ensure during testing that we deploy the pulled bundle
- [x] Check for places where we null check "Annotations" and it more robust (ie. check the content of the annotations, not just their existence)
- [x] Refactor pull
```

UncleGedd:
## Description
- refactors pull operation and provider

Fixes #177


decleaver (approved):
walked through changes. lgtm

Issue #399, opened 2024-02-06
Smoke test with UDS Core


UncleGedd:
We should deploy a portion of the UDS Core stack via `uds deploy` and `uds run ..` as part of a smoke test in our release pipelines as a minimum

PR #474, opened 2024-02-27
chore: add UDS Core smoke test

justin-o12:
## Description

This adds a new job in the release workflow that will deploy a new test UDS bundle consisting of:
- k3d cluster
- zarf init
- uds core

It then validates a majority of the[ UDS core packages](https://github.com/defenseunicorns/uds-core/tree/main/src) excluding pepr (it needs npx, but we can probably add it and make this work if we think it's necessary) and the packages that uds core doesn't test themselves.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-cli/issues/399

## Type of change

- [x] Other (security config, docs update, etc) - chore for new test

## Checklist before merging

For testing, I created and deployed the new test bundle locally and ran the validate steps manually.


src/test/bundles/13-uds-core/uds-bundle.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+  - name: init
+    repository: oci://ghcr.io/defenseunicorns/packages/init
+    # renovate: datasource=github-tags depName=defenseunicorns/zarf versioning=semver
+    ref: v0.32.1
+
+  - name: core
+    repository: oci://ghcr.io/defenseunicorns/packages/uds/core
+    # renovate: datasource=github-tags depName=defenseunicorns/uds-core versioning=semver
+    ref: 0.13.1-registry1
```
mjnagel, line 38: I do like the idea of testing against uds-core in general but this would create a weird circle dependency I think? If uds-cli has a breaking change what would the path forward be to cut a release? It seems like you'd have to update uds-core to match the syntax/whatever for the un-released version of uds-cli?

.github/workflows/release.yaml
```diff
@@ -80,9 +79,42 @@ jobs:
...
+        # renovate: datasource=github-tags depName=defenseunicorns/uds-core versioning=semver
+        run: build/uds deploy ghcr.io/defenseunicorns/packages/uds/bundles/k3d-core-istio-dev:0.13.1 --confirm --no-progress
+        shell: bash
+
+      - name: Validate UDS Core deployment
+        run: |
+          build/uds zarf tools wait-for gateway admin-gateway -n istio-admin-gateway --timeout 10s
+          build/uds zarf tools wait-for gateway passthrough-gateway -n istio-passthrough-gateway --timeout 10s
+          build/uds zarf tools wait-for gateway tenant-gateway -n istio-tenant-gateway --timeout 10s
+        shell: bash
```
justin-o12, line 113: `wait-for` does not need a condition like `ready` for gateways.  Below is a command without `ready` on a gateway that exists.
```
uds zarf tools wait-for gateway passthrough-gateway -n istio-passthrough-gateway --timeout 20s                                   130 ↵

 NOTE  Using config file

 NOTE  Using config file
  ✔  Waiting for gateway/passthrough-gateway in namespace istio-passthrough-gateway to exist.
```
Next is the command with `ready` on that same gateway.
```
💻 ~ » uds zarf tools wait-for gateway passthrough-gateway ready -n istio-passthrough-gateway --timeout 20s

 NOTE  Using config file

 NOTE  Using config file
     ERROR:  Wait timed out.
```
Next is the command on a gateway that doesn't exist without `ready`:
```
💻 ~ » uds zarf tools wait-for gateway fake-gateway -n istio-passthrough-gateway --timeout 20s                                            1 ↵

 NOTE  Using config file

 NOTE  Using config file
     ERROR:  Wait timed out.
```

src/test/bundles/13-uds-core/uds-bundle.yaml
UncleGedd: That's a really good point that I hadn't thought of. On the one hand, I'd really like to know if we have an unidentified breaking change, but I wouldn't want it to block releases. @justin-o12 what do you think of making this a nightly job instead? (see [example](https://github.com/defenseunicorns/uds-cli/blob/main/.github/workflows/nightly-ghcr.yaml) here)

mjnagel: Yeah I guess technically would be a breaking change in zarf probably that would cause the above scenario - didn't think about that... since we're just consuming the zarf package here and the bundle is crafted in this test file. That's probably a pretty unlikely case then?

justin-o12: We've switched to deploying the k3d dev bundle, because we do not publish the k3d dev Zarf package to OCI.  So, we're back to your original point that we could have a breaking change.  So, we've moved this workflow to a nightly so it doesn't block releases.

decleaver (approved):
looks good! checked out and ran steps manually

PR #499, opened 2024-03-12
fix: add a wait to the registry startup during tests

UncleGedd:
## Description

Our tests are flaky as the dickens (see error below), here's a simple test to see if the registry simply isn't ready

Screenshot 2024-03-12 at 9 21 13 AM


decleaver (approved):
:shipit:

PR #502, opened 2024-03-12
fix: autocomplete and vendor refactor

decleaver:
## Description

- Fixes tab completion for `uds run`
- moves zarf and run commands to vendored.go


UncleGedd (approved):
🚀

PR #503, opened 2024-03-13
fix: ensure manifest config is included with pulls

UncleGedd:
## Description

Fixes bug where the bundle root manifest config was missing after pulls. This prevented publishing to other OCI repos


decleaver (approved):
awesome!

PR #511, opened 2024-03-14
chore: update uds to zarf v0.32.5

decleaver:
## Description

Update to zarf v0.32.5

## Type of change

## Checklist before merging


src/pkg/utils/oci.go
```diff
@@ -52,23 +53,24 @@ func ToOCIStore(t any, mediaType string, store *ocistore.Store) (ocispec.Descrip
...
 
-	var layerDesc ocispec.Descriptor
+	var layerDesc *ocispec.Descriptor
 	// if image manifest media type, push to Manifests(), otherwise normal pushLayer()
 	if mediaType == ocispec.MediaTypeImageManifest {
-		layerDesc = content.NewDescriptorFromBytes(ocispec.MediaTypeImageManifest, b)
-		if err := remote.Repo().Manifests().PushReference(context.TODO(), layerDesc, bytes.NewReader(b), remote.Repo().Reference.String()); err != nil {
-			return ocispec.Descriptor{}, fmt.Errorf("failed to push manifest: %w", err)
+		descriptorFromBytes := content.NewDescriptorFromBytes(ocispec.MediaTypeImageManifest, b)
+		layerDesc = &descriptorFromBytes
```
UncleGedd, line 66: I think you can kill `layerDesc = &descriptorFromBytes` and just pass in `descriptorFromBytes ` to `PushReference()`

decleaver: so the layerDesc ends up getting used at the end of the function as part of the success message regardless of which if block you end up in.

UncleGedd (approved):
great work @decleaver 🚀

PR #527, opened 2024-03-26
fix: err when deploying with BubbleTea with no cluster

UncleGedd:
## Description
`uds deploy k3d-core-istio-dev:0.16.1` fails with a stack trace whenever a cluster isn't present. This PR nil checks the cluster object so the program doesn't blow up

### DIFF
--- a/src/pkg/bundle/common.go
+++ b/src/pkg/bundle/common.go
@@ -16,7 +16,6 @@ import (
 	"github.com/defenseunicorns/uds-cli/src/config"
 	"github.com/defenseunicorns/uds-cli/src/pkg/bundler/fetcher"
 	"github.com/defenseunicorns/uds-cli/src/types"
-	"github.com/defenseunicorns/zarf/src/config/lang"
 	"github.com/defenseunicorns/zarf/src/pkg/cluster"
 	"github.com/defenseunicorns/zarf/src/pkg/message"
 	"github.com/defenseunicorns/zarf/src/pkg/oci"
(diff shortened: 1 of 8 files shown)


decleaver (approved):
solid awareness on new and existing zarf functions you were able to leverage!

decleaver (approved):
:approved:

src/pkg/bundle/common.go
```diff
@@ -264,22 +263,15 @@ func ValidateBundleSignature(bundleYAMLPath, signaturePath, publicKeyPath string
...
-
 // GetDeployedPackageNames returns the names of the packages that have been deployed
 func GetDeployedPackageNames() []string {
 	var deployedPackageNames []string
-	deployedPackages, _ := GetDeployedPackages()
-	for _, pkg := range deployedPackages {
-		deployedPackageNames = append(deployedPackageNames, pkg.Name)
+	c, _ := cluster.NewCluster()
+	if c != nil {
+		deployedPackages, _ := c.GetDeployedZarfPackages()
```
YrrepNoj, line 271: Unsolicited comment but I stumbled upon this PR because I ran into this bug today.

Should we explicitly check for the `err` here? It's unlikely, because of how we usually do variable returns in the Zarf library, but it is possible for us to get an err and a nil `deployedPackages` and run into a panic on the line below when we try to loop through it.

YrrepNoj: I am realizing that my comment is technically wrong. You can range over a slice even if it `nil` without resulting in a panic.

UncleGedd: Good point though, we've gotten in trouble before by skipping these err checks. My thought with ignoring it here was that I don't really know what to do with the error; if there is an error, I think we want to continue the program not return the error

PR #530, opened 2024-03-27
chore: adds TUI tests

UncleGedd:
## Description

Adds tests for TUI deploys


decleaver (approved):
this is good, you were actually able to test more than I thought.

PR #534, opened 2024-03-28
chore: adds registry health check for tests

UncleGedd:
## Description

Adds a health check to ensure registries are g2g during tests


decleaver (approved):
you would remove the sleep!!!

Issue #408, opened 2024-02-07
Request ability to `uds dev deploy` for uds bundles


zachariahmiller:
### Is your feature request related to a problem? Please describe

I'm always frustrated when I'm trying to test a bundle and have to wait for the whole zarf init process.

### Describe the solution you'd like

When i am developing a uds bundle, I would like to be able to dev on it as quickly as possible. Zarf had a `zarf dev deploy` that facilitates this by effectively skipping create and dependency on zarf init, effectively running the whole package in `yolo` mode.

It would be nice with a bundle to be able to do the same as the process of compressing the bundle, deploying the init package to the cluster and then pushing all of the packages through that mechanism is fairly time consuming. The value prop here is basically the same as `zarf dev deploy`, but for bundle authors.
 
### Describe alternatives you've considered

keep doing the same create -> deploy, but its longer cycles and as a product of the worse devx.

### Additional context

Some considerations here:

1. should work for local and oci packages
1. many bundles include the zarf init package, which ideally could be bypassed or at least partially bypassed.

decleaver:
initial feature has been released

PR #536, opened 2024-03-29
feat: uds dev deploy

decleaver:
## Description

Adds functionality to run uds in dev mode with `uds dev deploy` cmd. 
- Creates local zarf packages if they do not exist (assuming the zarf package is expected in the same directory as the zarf.yaml file)
- Creates local bundle
- Deploys bundle (deploying zarf packages in YOLO mode) removing the necessity to do a `zarf init`

currently only works with local bundles

## Related Issue

Relates to #408 

## Type of change

## Checklist before merging


README.md
```diff
@@ -24,7 +24,7 @@
 1. [Bundle Overrides](docs/overrides.md)
 1. [Bundle Anatomy](docs/anatomy.md)
 1. [Runner](docs/runner.md)
-
+1. [Dev Mode](#dev)
```
UncleGedd, line 27:
```suggestion
1. [Dev Mode](#dev-mode)

```

src/cmd/dev.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/defenseunicorns/uds-cli/src/config/lang"
+	"github.com/defenseunicorns/uds-cli/src/pkg/bundle"
+	"github.com/defenseunicorns/uds-cli/src/pkg/bundle/tui/deploy"
+	"github.com/defenseunicorns/zarf/src/pkg/message"
+	"github.com/defenseunicorns/zarf/src/pkg/utils/helpers"
+	"github.com/spf13/cobra"
+	"golang.org/x/term"
+)
+
+var devCmd = &cobra.Command{
```
UncleGedd, line 24: Nice, good call putting this in another file, `uds.go` is getting big

src/cmd/dev.go
```diff
@@ -0,0 +1,119 @@
...
+	PreRun: func(_ *cobra.Command, args []string) {
+		pathToBundleFile := ""
+		if len(args) > 0 {
+			if !helpers.IsDir(args[0]) {
+				message.Fatalf(nil, "(%q) is not a valid path to a directory", args[0])
+			}
+			pathToBundleFile = filepath.Join(args[0])
+		}
+		// Handle .yaml or .yml
+		bundleYml := strings.Replace(config.BundleYAML, ".yaml", ".yml", 1)
```
UncleGedd, line 42: Since we're now handling `yaml` vs `yml` in multiple locations, can we break this out into a function?

src/cmd/dev.go
```diff
@@ -0,0 +1,119 @@
...
+		if err := bndlClient.Create(); err != nil {
+			bndlClient.ClearPaths()
+			message.Fatalf(err, "Failed to create bundle: %s", err.Error())
+		}
+
+		// Deploy dev bundle
+		bndlClient.SetDevSource(srcDir)
+
+		// don't use bubbletea if --no-tea flag is set
+		if config.CommonOptions.NoTea {
```
UncleGedd, line 94: I like how you defaulted to using the new TUI here, I think it pushes bundle devs to see/understand what their downstream users will see (and invites more TUI feedback!)

src/pkg/bundle/dev.go
```diff
@@ -0,0 +1,66 @@
...
+	"path/filepath"
+	"regexp"
+
+	"github.com/defenseunicorns/uds-cli/src/config"
+	zarfCLI "github.com/defenseunicorns/zarf/src/cmd"
+	"github.com/defenseunicorns/zarf/src/pkg/message"
+	zarfUtils "github.com/defenseunicorns/zarf/src/pkg/utils"
+)
+
+// CreateZarfPkg creates a zarf package if its missing when in dev mode
```
UncleGedd, line 19:
```suggestion
// CreateZarfPkgs creates a zarf package if its missing when in dev mode
```

src/pkg/bundle/dev.go
```diff
@@ -0,0 +1,66 @@
...
+func (b *Bundle) CreateZarfPkgs(srcDir string) {
+	path := filepath.Join(srcDir, b.cfg.CreateOpts.BundleFile)
+	if err := zarfUtils.ReadYaml(path, &b.bundle); err != nil {
+		message.Fatalf(err, "Failed to read bundle.yaml: %s", err.Error())
```
UncleGedd, line 23: Thinking out loud, since dev deploy uses the TUI for the deploy, as a follow up to this, we could add some extra TUI indicators this particular deploy operation

src/pkg/bundle/dev.go
```diff
@@ -0,0 +1,66 @@
...
+				zarfCLI.Execute()
+				if err != nil {
+					message.Fatalf(err, "Failed to create package %s: %s", pkg.Name, err.Error())
+				}
+			}
+		}
+	}
+}
+
+// setDevSource sets the source for the bundle when in dev mode
```
UncleGedd, line 57:
```suggestion
// SetDevSource sets the source for the bundle when in dev mode
```

src/pkg/bundle/dev.go
```diff
@@ -0,0 +1,66 @@
...
+// CreateZarfPkg creates a zarf package if its missing when in dev mode
+func (b *Bundle) CreateZarfPkgs(srcDir string) {
+	path := filepath.Join(srcDir, b.cfg.CreateOpts.BundleFile)
+	if err := zarfUtils.ReadYaml(path, &b.bundle); err != nil {
+		message.Fatalf(err, "Failed to read bundle.yaml: %s", err.Error())
+	}
+
+	zarfPackagePattern := `^zarf-.*\.tar\.zst$`
+	for _, pkg := range b.bundle.Packages {
+		if pkg.Repository == "" {
```
UncleGedd, line 28: for readability it might make sense to check the inverse here since we dealing with paths (ie. `if pkg.Path != ""`)

src/pkg/bundle/dev.go
```diff
@@ -0,0 +1,66 @@
...
+			for _, file := range files {
+				if regex.MatchString(file.Name()) {
+					packageFound = true
+					break
+				}
+			}
+			// create local zarf package if it doesn't exist
+			if !packageFound {
+				os.Args = []string{"zarf", "package", "create", path, "--confirm", "-o", path}
+				zarfCLI.Execute()
```
UncleGedd, line 48: Instead of using the Zarf CLI directly, any idea what it looks like to use the Zarf lib?

src/pkg/bundle/tui/deploy/model.go
```diff
@@ -64,7 +64,6 @@ type pkgState struct {
 	percDownloaded     int
 	downloaded         bool
 	verified           bool
-	isRemote           bool
```
UncleGedd, line 67: I think this may break some TUI things?

src/pkg/sources/tarball.go
```diff
@@ -88,7 +88,18 @@ func (t *TarballBundle) LoadPackage(dst *layout.PackagePaths, filter filters.Com
 			}
 		}
 	}
+
+	if config.Dev {
```
UncleGedd, line 92: config vs DeployOpts; since this is only a thing during deploys, instead of putting `Dev` in `config`, is it possible to add it to the DeployOpts?

src/pkg/sources/tarball.go
```diff
@@ -88,7 +88,18 @@ func (t *TarballBundle) LoadPackage(dst *layout.PackagePaths, filter filters.Com
...
+		pkg.Metadata.YOLO = true
+		// strip out all images and repos
+		for idx := range pkg.Components {
```
UncleGedd, line 95: 🚀

src/test/e2e/dev_test.go
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023-Present The UDS Authors
+
+// Package test provides e2e tests for UDS.
+package test
```
UncleGedd, line 5: Do you think it's worth adding a namespace or deployment check to these tests?

src/cmd/dev.go
```diff
@@ -0,0 +1,119 @@
...
+			pathToBundleFile = filepath.Join(args[0])
+		}
+		// Handle .yaml or .yml
+		bundleYml := strings.Replace(config.BundleYAML, ".yaml", ".yml", 1)
+		if _, err := os.Stat(filepath.Join(pathToBundleFile, config.BundleYAML)); err == nil {
+			bundleCfg.CreateOpts.BundleFile = config.BundleYAML
+		} else if _, err = os.Stat(filepath.Join(pathToBundleFile, bundleYml)); err == nil {
+			bundleCfg.CreateOpts.BundleFile = bundleYml
+		} else {
+			message.Fatalf(err, "Neither %s or %s found", config.BundleYAML, bundleYml)
```
UncleGedd, line 48: Something is fishy here, if I do a `uds dev deploy` in a directory with no `uds-bundle.yaml` then I don't get any error messages printed to the console, but if I look at the `uds logs` then I see them

src/cmd/dev.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+
+		bndlClient := bundle.NewOrDie(&bundleCfg)
+		defer bndlClient.ClearPaths()
+
+		// Check if local zarf packages need to be created
+		if len(srcDir) != 0 && srcDir[len(srcDir)-1] != '/' {
+			srcDir = srcDir + "/"
+		}
+		bndlClient.CreateZarfPkgs(srcDir)
```
UncleGedd, line 81: This function signature feels a little strange, maybe something like `bndlClient.CreateLocalZarfPkgs()`. Is it possible put the `srcDir` somewhere in the bundle client? If not maybe we rename `srcDir` to `bndlSrcDir` to be more explicit about what directory we're talking about

src/cmd/dev.go
```diff
@@ -0,0 +1,119 @@
...
+
+		// Create dev bundle
+		config.Dev = true
+		if err := bndlClient.Create(); err != nil {
```
UncleGedd, line 85: What do you think of having the created bundle have a slightly different filename to indicate that it's dev bundle, something like: `uds-bundle-test-local-and-remote-arm64-0.0.1-DEV.tar.zst`. Just trying to avoid accidentally publishing a dev bundle instead of a non-dev one

mjnagel:
@UncleGedd my expectation would be that init defaults to being skipped, _maybe_ with an option to not skip it? If you're doing a dev deploy you likely don't want the extra time/load of having zarf init components, unless its a custom init package and you want to test rook-ceph pieces for example?

src/pkg/bundle/tui/deploy/model.go
decleaver: yeah, this was weird, not sure if it was a merge thing, but the `isRemote` field was making my linter complain that it isnt being used anywhere

src/cmd/dev.go
UncleGedd: ahhh gotcha, didn't realize the bundles were identical. That makes sense because you're modifying the pkg in memory

decleaver:
NVM^ RE: zarf init, @UncleGedd educated me ❤️  **kind: ZarfInitConfig**

src/pkg/bundle/dev.go
decleaver: created #540 for tracking

src/pkg/bundle/dev.go
decleaver: so we are just checking to see if its a local zarf package, for the inverse (remote zarf packages) we don't do anything. Added a comment to help make more clear, also let me know if I'm misunderstanding 😃

src/pkg/bundle/dev.go
decleaver: so the lib route ends up looking like this:
```
...
pkgConfig := types.PackagerConfig{}
pkgConfig.CreateOpts.BaseDir = path
pkgConfig.CreateOpts.Output = path

var isCleanPathRegex = regexp.MustCompile(`^[a-zA-Z0-9\_\-\/\.\~\\:]+$`)
if !isCleanPathRegex.MatchString(config.CommonOptions.CachePath) {
	message.Warnf(lang.CmdPackageCreateCleanPathErr, zarfConfig.ZarfDefaultCachePath)
	config.CommonOptions.CachePath = zarfConfig.ZarfDefaultCachePath
}

// Ensure uppercase keys from viper
v := common.GetViper()
pkgConfig.CreateOpts.SetVariables = helpers.TransformAndMergeMap(
	v.GetStringMapString(common.VPkgCreateSet), pkgConfig.CreateOpts.SetVariables, strings.ToUpper)

// Configure the packager
pkgClient := packager.NewOrDie(&pkgConfig)
defer pkgClient.ClearTempPaths()

// Create the package
if err := pkgClient.Create(); err != nil {
	message.Fatalf(err, lang.CmdPackageCreateErr, err.Error())
}
```
I agree the cli route is a little weird, but it is less code/cleaner which is nice

src/pkg/sources/tarball.go
decleaver: actually don't have access to the DeployOpts at this point, so would also have to pass it in if we wanted to go the DeployOpts route.

src/pkg/bundle/dev.go
UncleGedd: 👍 cli route sounds good to me!

src/pkg/sources/tarball.go
UncleGedd: 👍 gotcha, config is fine for now, and good point that dev mode may expand to other ops in the future

src/cmd/dev.go
decleaver: this one is interesting, since we only have the pretty tui setup for the deploy piece and not the create piece. May consider removing the "tea" for now

src/pkg/bundle/dev.go
decleaver: updated to check path instead of repo

src/cmd/dev.go
decleaver: removing tea for now, will add as part of follow on feature issue

src/cmd/dev.go
decleaver: updated, srcDir isn't needed any more

UncleGedd (approved):

PR #538, opened 2024-03-29
fix: nil check pkg components in TUI

UncleGedd:
## Description

`uds deploy k3d-core-istio-dev:0.16.1` fails 1/10 times due a race condition with the deploy ticker, this PR adds an additional check to ensure we don't seg fault

Screenshot 2024-03-28 at 11 23 31 AM


decleaver (approved):
👍

PR #539, opened 2024-03-29
feat: enables setting namespaces in bundled Helm charts

### LINKED ISSUE #442: Allow ability to override a components namespace when imported into a bundle.

I would like to be able to override the namespace a chart in a zarf package is deploying into when i create a bundle. Currently in zarf, namespace is a mandatory field and depending on the use case that makes it difficult to reuse packages. For [example](https://github.com/defenseunicorns/uds-package-dependencies/blob/main/src/postgres/zarf.yaml) 

This package is used for dev dependencies in the SWF individual packages. For SWF integration, i cannot use this package as it stands today because it just deploys over/overwrites itself in the dev-postgres namespace. 

something like:

```yaml
kind: UDSBundle
metadata:
  name: uds-stuff
  version: dev

packages:
  - name: dev-postgres
    repository: ghcr.io/defenseunicorns/packages/uds/dev-postgres
    ref: 0.0.1
  - name: dev-postgres-sonarqube
    repository: ghcr.io/defenseunicorns/packages/uds/dev-postgres
    ref: 0.0.1
    overrides:
      postgres:
        postgresql:
          namespace: dev-postgres-sonarqube
```

I dont feel strongly about the syntax at all except for a general desire for it to be as simple and clear as possible. 

For the example case i've entertained using the zalando operator instead of the existing chart. Also creating multiple duplicae optional components in the source zarf package that override the namespace

### Additional context

I can see multiple places where this might break down/would still very much depend on how the upstream package was composed. For example, in the common dev postgres chart, the zarf package does a `wait` on the cluster resource that also depends on knowing the [namespace](https://github.com/defenseunicorns/uds-package-dependencies/blob/52138991993adfd625210c6c6cf8979a5ab8de0e/src/postgres/zarf.yaml#L42)

UncleGedd:
## Description

enables setting namespaces in bundled Helm charts

## Related Issue

Fixes #442

### DIFF
--- a/docs/overrides.md
+++ b/docs/overrides.md
@@ -9,6 +9,7 @@ Bundle overrides provide a mechanism to customize Helm charts inside of Zarf pac
     - [Syntax](#syntax)
     - [Values](#values)
     - [Variables](#variables)
+    - [Namespace](#namespace)
 
 ## Quickstart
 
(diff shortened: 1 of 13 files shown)


decleaver (approved):
pair reviewed

PR #542, opened 2024-04-01
fix: adds global GracefulPanic and checks to deploy TUI

UncleGedd:
## Description

adds global GracefulPanic and checks to deploy TUI


decleaver (approved):
looks good, hopefully addresses terminal access issues on panics

PR #543, opened 2024-04-02
fix: ensure we are pulling all components

UncleGedd:
## Description

ensure we are pulling all Zarf pkg layers by grabbing them from the `zarf.yaml` instead of the pkg root manifest


decleaver (approved):
verified with anthony this fixes issue as well

PR #547, opened 2024-04-04
feat: add ability to uds create to local output path

### LINKED ISSUE #461: Add an output flag for local bundle create
## Description
Today we do a `uds create -o` to output an OCI registry, let's also ensure we can do this for local directories

TristanHoladay:
## Description
Adds the ability to pass a local output path to `uds create`.

## Related Issue

resolves #461 

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/utils/utils.go
+++ b/src/pkg/utils/utils.go
@@ -13,6 +13,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/defenseunicorns/uds-cli/src/config"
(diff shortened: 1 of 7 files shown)


src/pkg/utils/utils.go
```diff
@@ -138,3 +139,48 @@ func ToLocalFile(t any, filePath string) error {
 func IsRemotePkg(pkg types.Package) bool {
 	return pkg.Repository != ""
 }
+
+func hasScheme(s string) bool {
+	return strings.Contains(s, "://")
+}
+
+func hasDomain(s string) bool {
+	return strings.Contains(s, ".") && len(s) > 1
```
UncleGedd, line 148: May need to add a bit more logic here, `-o .tmp` is not a domain but rather a hidden file

UncleGedd:
Great work! Could you also add some unit tests to the `IsRegistryURL` function? For example of the type of unit tests we do, see: https://github.com/defenseunicorns/uds-cli/blob/main/src/pkg/bundle/common_test.go

src/pkg/utils/utils.go
```diff
@@ -138,3 +139,54 @@ func ToLocalFile(t any, filePath string) error {
...
+		// port valid number ?
+		_, err := strconv.Atoi(port)
+		if err == nil {
+			return true
+		}
+	}
+	return false
+}
+
+// Checks if string is an oci url
```
UncleGedd, line 177:
```suggestion
// IsRegistryURL Checks if string is an oci url
```

src/pkg/utils/utils_test.go
```diff
@@ -0,0 +1,92 @@
...
+		{
+			name:        "HasRareDotInLocalDirectoryPath",
+			description: "Output is a hidden directory nested in path",
+			args:        args{output: "/pathto/test.dev/"},
+			wantResult:  false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
UncleGedd, line 86: Great work! Only other thing is instead of manually checking the tests with an `if` block, we prefer to use `require.Equal` so this would be something like:
```
actualResult := IsRegistryURL(tt.args.output)
require.Equal(t, tt.wantResult, actualResult)
```

UncleGedd (approved):
really great work on this 🚀

PR #554, opened 2024-04-04
fix: permit absolute paths for bundle create

### LINKED ISSUE #505: Absolute paths for local packages do not work
### Environment

Device and OS: ubuntu 22
App version: 0.9.3

### Steps to reproduce

1. Write full absolute path to package in a bundle

### Expected result

Bundle creates

### Actual Result

Absolute path is appended to current directory and fails

image

image

ZachGallagher:
## Description
Performs a simple `filepath.IsAbs` check on `pkg.Path` before appending relative paths to current working directory.

## Related Issue

Fixes #505 

## Type of change

## Checklist before merging


UncleGedd:
@ZachGallagher ! Welcome to UDS CLI and thank you for the PR! Couple of notes:
- I think the logic to add the absolute path check should go in this [function](https://github.com/defenseunicorns/uds-cli/blob/e364437dcd8a65f370c1e73d318e88b18b3994d3/src/pkg/bundle/common.go#L215)
- If you add it to that function you can unit test it over [here](https://github.com/defenseunicorns/uds-cli/blob/e364437dcd8a65f370c1e73d318e88b18b3994d3/src/pkg/bundle/common_test.go#L150)
- Note that we should accept absolute paths to directories (ie. `/tmp/packages/podinfo`) and Zarf tarballs (ie. `/tmp/packages/podinfo/zarf-package-podinfo-arm64-0.0.1.tar.zst`)

ZachGallagher:
Aha! Thank you. I appreciate (need) the feedback, as this was definitely about getting myself used to our conventions by picking a low-hanging fruit.

Regarding where I did make the change, would it be confusing to leave the comment there about it making the path relative if it's set, since that might no longer be the case?

UncleGedd:
> Regarding where I did make the change, would it be confusing to leave the snippet there about it making the path relative if path is set, since that might no longer be the case?

Great point, I think you can move this entire snippet (or at least its logic) to `getPkgPath`:
```
	if !filepath.IsAbs(pkg.Path) {
		pkg.Path = filepath.Join(b.cfg.CreateOpts.SourceDirectory, pkg.Path)
```

For more context, we recently refactored parts of this file and created `getPkgPath` to handle all of the `path` logic; your PR reminded me that we forgot to move all the `path` mutations to that single function. Let me know if that doesn't make sense, happy to pair on it in Gather

ZachGallagher:
I **_think_** it should be good now? 

Regarding your third original note, specifying the tar.zst file is for deploying only, right? I may have missed something because `create` still only accepts a directory with uds-bundly.yaml anyway. `uds deploy` still works specifying the full path just fine.

UncleGedd:
> I **_think_** it should be good now?
> 
> Regarding your third original note, specifying the tar.zst file is for deploying only, right? I may have missed something because `create` still only accepts a directory with uds-bundly.yaml anyway. `uds deploy` still works specifying the full path just fine.

You're right in that `uds create` accepts just a path to a directory; what I'm referring to the is something like this:
```yaml
kind: UDSBundle
metadata:
  name: example
  version: 0.0.1

packages:
  - name: nginx
    path: /tmp/nginx/zarf-package-nginx-arm64-0.0.1.tar.zst # abs path to Zarf tarball
    ref: 0.0.1
  - name: podinfo
    path: /tmp/podinfo # abs path to dir containing Zarf tarball
    ref: 0.0.1
```

src/pkg/bundle/common.go
```diff
@@ -212,9 +208,15 @@ func (b *Bundle) ValidateBundleResources(spinner *message.Spinner) error {
 	return nil
 }
 
-func getPkgPath(pkg types.Package, arch string) string {
+func getPkgPath(pkg types.Package, arch string, srcDir string) string {
 	var fullPkgName string
 	var path string
+	// Set path relative to the source directory if not absolute
+	if pkg.Path != "" {
```
UncleGedd, line 215: This works! Only thing left is to remove the outer `if pkg.Path != ""` statement. We check for that condition before this function is called

src/pkg/bundle/common_test.go
```diff
@@ -160,39 +161,52 @@ func Test_getPkgPath(t *testing.T) {
 		{
 			name: "init full path",
 			args: args{
-				pkg:  types.Package{Name: "init", Ref: "0.0.1", Path: "../fake/path/custom-init.tar.zst"},
-				arch: "fake64",
+				pkg:    types.Package{Name: "init", Ref: "0.0.1", Path: "../fake/path/custom-init.tar.zst"},
+				arch:   "fake64",
+				srcDir: "/mock/source",
 			},
-			want: "../fake/path/custom-init.tar.zst",
```
UncleGedd, line 166: No need to remove the existing tests because those test for existing functionality that we'd like to keep, you'll need to add the absolute path test cases to this list of tests

ZachGallagher:
Hey, @UncleGedd. Hope this ping is okay. Didn't want to let this linger open for too long.

I believe the main concerns are resolved. The unit tests are _functionally_ the same, as the result of the function is now always an absolute path. And `uds create` does indeed still work with absolute or relative paths in the UDSBundle definition.

Were there any other concerns before merging?

src/pkg/bundle/common_test.go
UncleGedd: Clearly I didn't read carefully enough 😆

UncleGedd:
> The unit tests are functionally the same, as the result of the function is now always an absolute path

You're right! Just resolved my own comment. Appreciate the pring

UncleGedd (approved):
thanks again @ZachGallagher

Issue #550, opened 2024-04-04
Refine dev deploy


UncleGedd:
```[tasklist]
### Tasks
- [x] ensure SBOMs are skipped by default
- [x] add --packages
- [x] mark as beta
- [ ] https://github.com/defenseunicorns/uds-cli/issues/537
- [x] add --flavor flag
- [ ] https://github.com/defenseunicorns/uds-cli/issues/628
```

decleaver:
blocked while #537 gets worked

andrewg-xyz:
All tasks are complete? closing.

PR #560, opened 2024-04-08
docs: dev deploy ADR

decleaver:
## Description

ADR for `dev deploy` implementation

## Related Issue

Relates to #550 

## Type of change

## Checklist before merging

### DIFF
--- a/adr/0005-dev-mode.md
+++ b/adr/0005-dev-mode.md
@@ -21,7 +21,7 @@ The current bundle development lifecycle is:
 Currently the [Local Artifacts](#bundle-create) option has been implemented to provide an MVP (minimal viable product). This ADR is intended to determine potential future implementation.
 
 ## Alternatives
-Regardless of implementation, the plan is to introduce `uds dev deploy` which allows you to deploy a UDS bundle in dev mode. If a **local** zarf package is missing, this command will create that zarf package for you assuming that your `zarf.yaml` file and zarf package are expected in the same directory. It will then create your bundle and deploy your zarf packages in [YOLO](https://docs.zarf.dev/docs/faq#what-is-yolo-mode-and-why-would-i-use-it) mode, eliminating the need to do a `uds zarf init`.
+Regardless of implementation, the plan is to introduce `uds dev deploy` which allows you to deploy a UDS bundle in dev mode. When deploying in dev mode, any `kind: ZarfInitConfig` packages in the bundle will be ignored. If a **local** zarf package is missing, this command will create that zarf package for you assuming that your `zarf.yaml` file and zarf package are expected in the same directory. It will then create your bundle and deploy your zarf packages in [YOLO](https://docs.zarf.dev/docs/faq#what-is-yolo-mode-and-why-would-i-use-it) mode, eliminating the need to do a `uds zarf init`.
 
 ### Local Artifacts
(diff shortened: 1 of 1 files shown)


adr/0005-dev-mode.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Local Artifacts
+ In order to maximize code reuse and leverage existing logic we will be creating local artifacts (both bundle and zarf package artifacts) in the same way that UDS bundles are currently created and deployed. This solution also minimizes the amount of code needed to process a zarf package regardless if it is a remote or local package.
+
+ + code readibility/reuse regardless if `dev` or not and if processing local or remote packages
+ - less efficient
+
+### In Memory
+The current `create` and `deploy` functionality does additional work creating local artifacts that potentially isn't necessary. We can speed up the dev cycle even more by doing the package and bundling processing in memory. Zarf currently has a `dev deploy` command that allows you to deploy zarf packages in memory in YOLO mode. This code can be leveraged, but only works for local packages. Additional work would be needed to handle remote packages. We will also need to handle the passing of variables and overrides between zarf packages within a bundle.
```
Racer159, line 33: To be clear Zarf doesn't do this in memory - it uses a temp dir on disk to "create" into and deploy from (basically just eliminating the intermediate tarball (which does provide a speed benefit).  Because the package is YOLO'ed before create it also ignores images and remote resources like git repos which provides a speed boost too.

adr/0005-dev-mode.md
```diff
@@ -0,0 +1,45 @@
...
+2. Create a `uds-bundle.yaml` and add packages
+3. Create a bundle with `uds create <dir>`
+4. Start up a cluster if one does not already exist
+5. Run `uds zarf init` to initialize th cluster
+6. Deploy the bundle with `uds deploy BUNDLE_TARBALL|OCI_REF]`
+
+Currently the [Local Artifacts](#bundle-create) option has been implemented to provide an MVP (minimal viable product). This ADR is intended to determine potential future implementation.
+
+## Alternatives
+Regardless of implementation, the plan is to introduce `uds dev deploy` which allows you to deploy a UDS bundle in dev mode. If a **local** zarf package is missing, this command will create that zarf package for you assuming that your `zarf.yaml` file and zarf package are expected in the same directory. It will then create your bundle and deploy your zarf packages in [YOLO](https://docs.zarf.dev/docs/faq#what-is-yolo-mode-and-why-would-i-use-it) mode, eliminating the need to do a `uds zarf init`.
```
Racer159, line 24: One nice thing to handle here would also be to ignore any `kind: ZarfInitConfig` packages that might be inside a bundle (since they would mess up a dev deploy anyway by activating the state/agent).  That would make for a very clean way to filter them out whatever init package people were using.

adr/0005-dev-mode.md
```diff
@@ -0,0 +1,45 @@
+# 5. Dev Mode
```
Racer159, line 1: Really looking forward to this!!!

adr/0005-dev-mode.md
decleaver: oh yeah, forgot to mention that is implemented!

UncleGedd (dismissed):
🚀

UncleGedd (approved):
🚀

PR #582, opened 2024-04-23
fix: ensure we handle paths correctly in dev deploy

UncleGedd:
## Description

- ensure we can use absolute paths in `dev deploy`


decleaver (approved):
tested locally with absolute path set in bundle for local zarf package running `dev deploy`

PR #594, opened 2024-05-07
feat: allow helm overrides from valuesfile

### LINKED ISSUE #572: Allow helm override values to be set via values file

When working with large bundles with many helm overrides across many packages the uds-bundle.yaml quickly becomes unwieldy and large from so much extra yaml being under the `values` in helm overrides. Adopting a pattern similar to how helm already works with values files I think would be beneficial.

- **Given** a bundle with many values based helmOverrides
- **When** a values file is provided and referenced in the `uds-bundle.yaml`
- **Then** those values are passed to the helm chart in that package

### Additional context

possible syntax:

```yaml
  - name: package
    repository: ghcr.io/some/package
    ref: 0.1
    overrides:
      component:
        chart:
          valuesFiles:
          - values-1.yaml
          - values-2.yaml
```

decleaver:
## Description

Allows for passing helm overrides values from a values file.

## Related Issue

Fixes #572 

## Type of change

## Checklist before merging

### DIFF
--- a/docs/overrides.md
+++ b/docs/overrides.md
@@ -123,8 +123,6 @@ In this example, the `helm-overrides-package` Zarf package has a component calle
 
 The `valuesFiles` in an `overrides` block are a list of `file`'s. It allows users to override multiple values in a Zarf package component's underlying Helm chart, by providing a file with those values instead of having to include them all indiviually in the `overrides` block.
 
-<b>NOTE:</b> If a value is specified in both a `valuesFile` and as a `value` in the `overrides` block, the value set in the `valuesFile` will take precedence.
-
 ### Values
 
 The `values` in an `overrides` block are a list of `path` and `value` pairs. They allow users to override values in a Zarf package component's underlying Helm chart. Note that values are specified by bundle authors and **cannot be modified** after the bundle has been created.
@@ -180,6 +178,12 @@ packages:
             - path: "podinfo.ui.color"
               value: ${COLOR}
 ```
+
+#### Value Precedence
+Value precedence is as follows:
+1. The `values` in an `overrides` block
+1. `values` set in the first `valuesFile` (if specified)
+1. `values` set in the next `valuesFile` (if specified)
 
 ### Variables
 Variables are similar to [values](#values) in that they allow users to override values in a Zarf package component's underlying Helm chart; they also share a similar syntax. However, unlike `values`, `variables` can be overridden at deploy time. For example, consider the `variables` key in the following `uds-bundle.yaml`:
--- a/docs/overrides.md
+++ b/docs/overrides.md
@@ -119,7 +119,7 @@ podAnnotations:
 ```
 In this example, the `helm-overrides-package` Zarf package has a component called `helm-overrides-component` which contains a Helm chart called `podinfo`; note how these names are keys in the `overrides` block. The `podinfo` chart has a `replicaCount` value that is overridden to `2`, a `podAnnotations` value that is overridden to include `customAnnotation: "customValue"` and a variable called `UI_COLOR` that is overridden to `purple`.
 
-### ValuesFiles
+### Values Files
 
 The `valuesFiles` in an `overrides` block are a list of `file`'s. It allows users to override multiple values in a Zarf package component's underlying Helm chart, by providing a file with those values instead of having to include them all indiviually in the `overrides` block.
 
--- a/docs/overrides.md
+++ b/docs/overrides.md
@@ -182,8 +182,8 @@ packages:
 #### Value Precedence
 Value precedence is as follows:
 1. The `values` in an `overrides` block
-1. `values` set in the first `valuesFile` (if specified)
-1. `values` set in the next `valuesFile` (if specified)
+1. `values` set in the last `valuesFile` (if more than one specified)
+1. `values` set in the previous `valuesFile` (if more than one specified)
 
 ### Variables
 Variables are similar to [values](#values) in that they allow users to override values in a Zarf package component's underlying Helm chart; they also share a similar syntax. However, unlike `values`, `variables` can be overridden at deploy time. For example, consider the `variables` key in the following `uds-bundle.yaml`:
--- a/docs/overrides.md
+++ b/docs/overrides.md
@@ -59,6 +59,8 @@ packages:
     overrides:
       helm-overrides-component:
         podinfo:
+          valuesFiles:
+            - file: values.yaml
           values:
             - path: "replicaCount"
               value: 2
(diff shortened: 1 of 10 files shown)


src/pkg/bundle/create.go
```diff
@@ -26,6 +28,32 @@ func (b *Bundle) Create() error {
 		return err
 	}
 
+	// Populate values from valuesFiles if provided
```
UncleGedd, line 31: I was about to make a comment asking why this is in `create` vs `deploy` but I think I understand. The nuance is that this values file is only for chart defaults

docs/overrides.md
```diff
@@ -102,7 +112,18 @@ packages:
...
+podAnnotations:
+  customAnnotation: "customValue"
+```
+In this example, the `helm-overrides-package` Zarf package has a component called `helm-overrides-component` which contains a Helm chart called `podinfo`; note how these names are keys in the `overrides` block. The `podinfo` chart has a `replicaCount` value that is overridden to `2`, a `podAnnotations` value that is overridden to include `customAnnotation: "customValue"` and a variable called `UI_COLOR` that is overridden to `purple`.
+
+### ValuesFiles
+
+The `valuesFiles` in an `overrides` block are a list of `file`'s. It allows users to override multiple values in a Zarf package component's underlying Helm chart, by providing a file with those values instead of having to include them all indiviually in the `overrides` block.
+
+<b>NOTE:</b> If a value is specified in both a `valuesFile` and as a `value` in the `overrides` block, the value set in the `valuesFile` will take precedence.
```
UncleGedd, line 126: ooo what are your thoughts on this? I would think the `overrides` block would take precedence over the values file?

decleaver: sounds good,  i can be smarter about blindly adding the values to the array as well. As is, the file only takes precedence because it gets added to the values array after the overrides values get added, so on deploy the overrides values get overwritten by the values in the file (if duplicates) when the values map gets populated.

docs/overrides.md
```diff
@@ -102,7 +112,16 @@ packages:
...
 
-In this example, the `helm-overrides-package` Zarf package has a component called `helm-overrides-component` which contains a Helm chart called `podinfo`; note how these names are keys in the `overrides` block. The `podinfo` chart has a `replicaCount` value that is overridden to `2` and a variable called `UI_COLOR` that is overridden to `purple`.
+```yaml
+#values.yaml
+podAnnotations:
+  customAnnotation: "customValue"
+```
+In this example, the `helm-overrides-package` Zarf package has a component called `helm-overrides-component` which contains a Helm chart called `podinfo`; note how these names are keys in the `overrides` block. The `podinfo` chart has a `replicaCount` value that is overridden to `2`, a `podAnnotations` value that is overridden to include `customAnnotation: "customValue"` and a variable called `UI_COLOR` that is overridden to `purple`.
+
+### ValuesFiles
```
UncleGedd, line 122:
```suggestion
### Values Files
```

src/pkg/bundle/create.go
```diff
@@ -107,3 +114,60 @@ func (b *Bundle) confirmBundleCreation() (confirm bool) {
...
+// processValuesFiles reads values from valuesFiles and updates the bundle with the override values
+func (b *Bundle) processValuesFiles() error {
+	// Populate values from valuesFiles if provided
+	for i, pkg := range b.bundle.Packages {
+		for j, overrides := range pkg.Overrides {
+			for k, bundleChartOverrides := range overrides {
+				for _, valuesFile := range bundleChartOverrides.ValuesFiles {
+					// Check relative vs absolute path
+					fileName := filepath.Join(b.cfg.CreateOpts.SourceDirectory, valuesFile)
+					if filepath.IsAbs(valuesFile) {
```
UncleGedd, line 127: Noting that we do this in other areas of the codebase (handling absolute vs relative file paths), will make a follow-up issue to refactor into a single fn

src/pkg/bundle/create.go
```diff
@@ -107,3 +114,60 @@ func (b *Bundle) confirmBundleCreation() (confirm bool) {
...
+						return err
+					}
+					if len(values) > 0 {
+						// populate BundleChartValue slice to use for merging existing values
+						valuesFileValues := make([]types.BundleChartValue, 0, len(values))
+						for key, value := range values {
+							valuesFileValues = append(valuesFileValues, types.BundleChartValue{Path: key, Value: value})
+						}
+						// update bundle with override values
+						overrides := b.bundle.Packages[i].Overrides[j][k]
```
UncleGedd, line 142: The name `overrides` shadows `overrides` from above which can be confusing, recommend changing them to be different

src/pkg/bundle/create.go
```diff
@@ -107,3 +114,60 @@ func (b *Bundle) confirmBundleCreation() (confirm bool) {
...
+					}
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// mergeBundleChartValues merges two lists of BundleChartValue using the values from list1 if there are duplicates
+func mergeBundleChartValues(list1, list2 []types.BundleChartValue) []types.BundleChartValue {
```
UncleGedd, line 155: Nice! My only recommendation is to have the second arg override the first (I think this is more standard), and rename `list1` and `list2` to reflect what var is going to override the other (`originals` vs `overrides` or something like that)

src/test/bundles/07-helm-overrides/values-file/values.yaml
```diff
@@ -0,0 +1,12 @@
+# podinfo.replicaCount: 3
```
UncleGedd, line 1: I think we want this uncommented so the test below shows `values` overwriting the `valuesFiles` value

src/test/bundles/07-helm-overrides/values-file/uds-bundle.yaml
```diff
@@ -0,0 +1,43 @@
...
+packages:
+  - name: helm-overrides
+    path: "../../../packages/helm"
+    ref: 0.0.1
+
+    overrides:
+      podinfo-component:
+        unicorn-podinfo:
+          valuesFiles:
+            - values.yaml
```
UncleGedd, line 16: Can we also test what it looks to have multiple `valuesFiles`? Maybe just an add-on to the e2e test?

src/pkg/bundle/create.go
```diff
@@ -107,3 +114,69 @@ func (b *Bundle) confirmBundleCreation() (confirm bool) {
...
+
+	// Iterate over each list in order
+	for _, bundleChartValues := range bundleChartValueLists {
+		// Add entries from the current list to the merged list, overwriting any existing entries
+		for _, bundleChartValue := range bundleChartValues {
+			if _, ok := paths[bundleChartValue.Path]; ok {
+				// Remove the existing entry from the merged list
+				for i, mergedValue := range merged {
+					if mergedValue.Path == bundleChartValue.Path {
+						merged = append(merged[:i], merged[i+1:]...)
```
UncleGedd, line 171: Pretty sure the code will panic if `[i+1]` goes over the slice's len. Couple of options here:
1. Process these values as a map instead of a list and convert the result to a list at the end
2. Can potentially use `helpers.MergeSlices`

UncleGedd: 🤖 can gave me the exact same logic the other day when I was de-duping lists, and I ended up getting panics during tests

UncleGedd: ❤️ the variadic args!

decleaver: lol :robot: doubled down, and then apologized. Went with option 1. map -> slice

UncleGedd (approved):
🚀

PR #596, opened 2024-05-08
feat: strict bundle yaml validation

### LINKED ISSUE #558: uds create silently succeeds on invalid bundle override definitions
### Environment

Device and OS: m2 MBP
App version: 0.10.3
Other:

### Steps to reproduce

1. Make a bundle like this initial part of the gitlab test bundle
```yaml
kind: UDSBundle
metadata:
  name: gitlab-test
  description: A UDS bundle for deploying Gitlab and it's dependencies on a development cluster
  # x-release-please-start-version
  version: 16.10.1-uds.1
  # x-release-please-end

packages:
  - name: dev-minio
    repository: ghcr.io/defenseunicorns/packages/uds/dev-minio
    ref: 0.0.2

  - name: dev-namespace
    path: ../
    ref: 0.1.0

  - name: postgres-operator
    repository: ghcr.io/defenseunicorns/packages/uds/postgres-operator
    ref: 1.10.1-uds.4-upstream
    overrides:
      postgres-operator:
        uds-postgres-config:
          variables:
            - name: POSTGRESQL
              description: "Configure postgres using CRs via the uds-postgres-config chart"
              path: postgresql
```
2. move the postgres-operator overrides and below back two spaces. (like in screenshot). Note the schema can tell its invalid.
image

3. run bundle create

### Expected result

bundle creates invalid syntax

### Actual Result

bundle create succeeds but the override definition is invalid and doesn't work when subsequently deploying the bundle

I tested multiple times after this issue was brought up to me but the create is embedded inside a longer series of tasks so the output obfuscated. Rest assured it creates. 

### Severity/Priority
🤷 
### Additional Context

Add any other context or screenshots about the technical debt here.

decleaver:
## Description

Enables strict bundle validation

## Related Issue

Fixes #558 

## Type of change

## Checklist before merging


src/pkg/utils/utils.go
```diff
@@ -183,3 +185,14 @@ func IsRegistryURL(s string) bool {
 
 	return false
 }
+
+func ReadYaml(path string, destConfig any) error {
```
UncleGedd, line 189: Maybe `ReadYAMLStrict` vs `ReadYaml`?

src/pkg/utils/utils.go
```diff
@@ -183,3 +185,15 @@ func IsRegistryURL(s string) bool {
...
+// ReadYAMLStrict reads a YAML file into a struct, with strict parsing
+func ReadYAMLStrict(path string, destConfig any) error {
+	message.Debugf("Reading YAML at %s", path)
+
+	file, err := os.ReadFile(path)
+	if err != nil {
+		return err
+	}
+
+	return goyaml.UnmarshalWithOptions(file, destConfig, goyaml.Strict())
```
UncleGedd, line 198: Nice and simple! I've tried this with a couple of mistyped bundles and it works as expected. Noting that we could get fancier and use [validators](https://github.com/defenseunicorns/zarf/blob/29d3ee39dfcbe173405f6e0010cd0dfb9468cd6e/src/pkg/packager/lint/lint.go#L269) like how Zarf does it, but this solution is really simple and it works

UncleGedd (changes requested):
In all the places where we return an error with `ReadYAMLStrict`, let's make the error msg more explicit about what went wrong, something like:
```
message.Fatalf(err, "Failed to read %s, error in YAML: %s", b.cfg.CreateOpts.BundleFile, err.Error())
```

decleaver:
> In all the places where we return an error with `ReadYAMLStrict`, let's make the error msg more explicit about what went wrong, something like:
> 
> ```
> message.Fatalf(err, "Failed to read %s, error in YAML: %s", b.cfg.CreateOpts.BundleFile, err.Error())
> ```

Updated the ReadYAMLStrict function to return more explicit error messaging

UncleGedd:
Couple of lingering merge conflicts and I think this one is g2g

UncleGedd (approved):

PR #599, opened 2024-05-09
feat: adds --list-images arg to inspect

eFAILution:
This PR introduces two major enhancements to the project:

1. inspect --list-images Option: This new command-line option allows users to list all the images contained in the packages of a bundle. This feature enhances the usability of the tool by providing a quick and easy way to inspect the images without having to manually parse the bundle or package metadata.

2. YAML Bundle Reading Capability: The tool can now read and parse bundle metadata directly from a YAML file as the tarball is not always available. The tool reads the YAML file, unmarshals the data into a Bundle struct, and then performs the same operations as it would if the bundle were loaded from an OCI registry.

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


UncleGedd:
Paired with @eFAILution and refactored this PR to introduce the `--list-images` flag. Expected behavior is that user's should be able to point `uds inspect` at a bundle YAML and get a list of images that will be included in the bundle before its created

README.md
```diff
@@ -107,6 +107,13 @@ Inspect the `uds-bundle.yaml` of a bundle
 1. From an OCI registry: `uds inspect oci://ghcr.io/defenseunicorns/dev/<name>:<tag>`
 1. From your local filesystem: `uds inspect uds-bundle-<name>.tar.zst`
 
+#### Viewing Images in a Bundle
+It is possible derive images from a `uds-bundle.yaml`. This can be useful for situations where you need to know what images will be bundled before you actually create the bundle. This is accomplished with the `--list-images`. For example:
```
TristanHoladay, line 111:
```suggestion
It is possible to derive images from a `uds-bundle.yaml`. This can be useful for situations where you need to know what images will be bundled before you actually create the bundle. This is accomplished with `--list-images`. For example:
```

README.md
```diff
@@ -107,6 +107,13 @@ Inspect the `uds-bundle.yaml` of a bundle
...
+
+`uds inspect ./uds-bundle.yaml --list-images`
+
+This command will return a list of images derived from the bundle's packages and taking into account optional and required package components.
```
TristanHoladay, line 115:
```suggestion
This command will return a list of images derived from the bundle's packages, taking into account optional and required package components.
```

src/pkg/bundle/inspect.go
```diff
@@ -5,13 +5,48 @@
...
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pterm/pterm"
 )
 
 // Inspect pulls/unpacks a bundle's metadata and shows it
 func (b *Bundle) Inspect() error {
+	//  handle --list-images flag
+	if b.cfg.InspectOpts.ListImages {
+		err := utils.CheckYAMLSourcePath(b.cfg.InspectOpts.Source)
+		if err != nil {
```
TristanHoladay, line 32: nit, but is there a particular reason to have the err check separate on this one but "inlined" (if that's the right word) for the next call. just wondering if we can make them all look the same for consistency.

Also, do you think there's any value in extracting the list image functions into a separate function `listImages()` that gets called in `Inspect()`. Feels like it's getting a little long.

decleaver (approved):
looks good. paired on a portion and tested different use cases locally. :shipit:

Issue #583, opened 2024-04-23
Improve dev docs and use Maru


UncleGedd:
## Description

Refactor dev docs to make it easier for new contributors to know where to start and use Maru tasks (ie. `uds run`) to dogfood the runner capability

```[tasklist]
### Tasks
- [x] Add sections to `CONTRIBUTING.md` about test strategy and expectations
- [x] Refactor Makefile to use Maru tasks where applicable
- [x] Refactor Github workflows to use Maru tasks where applicable
- [x] Explore how to do `go install github.com/defenseunicorns/uds-cli@v0.10.4`
```

UncleGedd:
I researched `go install` but it looks like it's untenable because of all of our Go build args/vars

UncleGedd:
Closed with https://github.com/defenseunicorns/uds-cli/pull/602

PR #602, opened 2024-05-09
chore: swap Makefile for Maru

UncleGedd:
## Description

Removes `Makefile` in an effort to dogfood Maru / `uds run`

## Related Issue
https://github.com/defenseunicorns/uds-cli/issues/583


tasks/tests.yaml
```diff
@@ -0,0 +1,38 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: 2023-Present The UDS Authors
+
+tasks:
+  - name: test-unit
```
zachariahmiller, line 5: with these test tasks... As it is today they wont actually display in `uds run --list`, which i less than ideal from a discoverability standpoint. 

Even if that is acceptable for now (we with maru should be able to facilitate a way to expose certain included tasks in the list command), prefixing each test with `test-` is redundant. The call end up being `uds run test:test-e2e`.

Since its coming form the test task imported as `test:` you can drop the `test-` in all the tests and just then the run commands just become `uds run test:e2e`

tasks.yaml
```diff
@@ -0,0 +1,83 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: 2023-Present The UDS Authors
+
+includes:
+  - test: tasks/tests.yaml
+  - schema: tasks/schema.yaml
+
+variables:
```
zachariahmiller, line 8: I dont think you actually need to declare the variables here, since they are set below you wont be able to override them and they're not getting passed anywhere.

tasks/tests.yaml
UncleGedd: Understood the `uds run --list` constraint and good point on dropping the `test-` prefix, will update

UncleGedd: https://github.com/defenseunicorns/maru-runner/issues/75

decleaver: can do a `uds run --list-all` though

UncleGedd: lol I forget we had added that!

UncleGedd: It's even in the docs!


Screenshot 2024-05-10 at 10 37 59 AM

.github/workflows/nightly-ghcr.yaml
```diff
@@ -25,8 +25,11 @@ jobs:
       - name: Setup golang
         uses: ./.github/actions/golang
 
+      - name: Install UDS CLI
+        uses: ./.github/actions/install-uds-cli
+
```
mjnagel, line 30: It might not matter but is there a reason for installing + using `uds run` for this first "bootstrap" vs using `maru`/`run` direct from maru upstream? `uds` is probably easier to install (brew) but felt odd at first glance that we would require `uds` to build `uds` 😆. This is probably a super edge case but I think you could end up in a weird situation if you needed a new maru feature to update a task, in a single PR - the `brew` installed uds-cli would have an older maru version.

Happy to punt on this one - just some food for thought.

mjnagel (commented):
Overall lgtm - couple of minor comments I'm curious on. Happy to approve once resolved.

tasks.yaml
```diff
@@ -0,0 +1,74 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: 2023-Present The UDS Authors
+
+includes:
+  - test: tasks/tests.yaml
+  - schema: tasks/schema.yaml
```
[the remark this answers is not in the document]
UncleGedd: good catch! I missed that one

tasks.yaml
```diff
@@ -0,0 +1,74 @@
...
+          - name: CRANE_VERSION
+      - cmd: go list -f '{{.Version}}' -m github.com/mholt/archiver/v3
+        setVariables:
+          - name: ARCHIVER_VERSION
+      - cmd: go list -f '{{.Version}}' -m helm.sh/helm/v3
+        setVariables:
+          - name: HELM_VERSION
+      - cmd: git describe --tags 2>/dev/null || echo "UnknownVersion"
+        setVariables:
+          - name: CLI_VERSION
```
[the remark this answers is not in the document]
UncleGedd: 👍

.github/workflows/nightly-ghcr.yaml
UncleGedd: This one is intentional, we wanted to use `uds` to build `uds`. The edge case is a good point, my thinking was that we would want to know if there was a breaking Maru change so we could handle it on behalf of downstream teams

mjnagel: That sounds good - I think this works either way to dogfood uds.

mjnagel (approved):

Issue #613, opened 2024-05-15
Optional components aren't excluded from the bundle


UncleGedd:
### Steps to reproduce

1. Create a Zarf package with optional components
2. Create a bundle to omit those optional components
3. The optional components are still included in the bundle artifact

### Notes
The `optionalComponents` key in the `uds-bundle.yaml` works behaviorally, but the excluded components are still bundled 👎

UncleGedd:
closed with https://github.com/defenseunicorns/uds-cli/pull/622

PR #622, opened 2024-05-18
feat: remove unnecessary bundle layers and refactor verification

UncleGedd:
## Description
- ensure we only bundle images from required + selected components (results in smaller bundles)
- remove slow layer verification

## Related Issue
https://github.com/defenseunicorns/uds-cli/issues/613

### DIFF
--- a/src/test/e2e/optional_bundle_test.go
+++ b/src/test/e2e/optional_bundle_test.go
@@ -1,0 +1,163 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023-Present The UDS Authors
+
+// Package test provides e2e tests for UDS.
+package test
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBundleOptionalComponents(t *testing.T) {
+	deployZarfInit(t)
+	e2e.SetupDockerRegistry(t, 888)
+	defer e2e.TeardownRegistry(t, 888)
+
+	// create 2 Zarf pkgs to be bundled
+	zarfPkgPath := "src/test/packages/podinfo-and-nginx"
+	e2e.CreateZarfPkg(t, zarfPkgPath, false)
+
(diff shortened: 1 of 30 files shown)


src/test/e2e/optional_bundle_test.go
```diff
@@ -0,0 +1,163 @@
...
+)
+
+func TestBundleOptionalComponents(t *testing.T) {
+	deployZarfInit(t)
+	e2e.SetupDockerRegistry(t, 888)
+	defer e2e.TeardownRegistry(t, 888)
+
+	// create 2 Zarf pkgs to be bundled
+	zarfPkgPath := "src/test/packages/podinfo-and-nginx"
+	e2e.CreateZarfPkg(t, zarfPkgPath, false)
```
decleaver, line 26: consider making this force create? setting to true. These packages also get created in other tests I believe, which was causing this test to fail locally.

src/test/e2e/optional_bundle_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+	zarfPkgPath = "src/test/packages/prometheus"
+	pkg := filepath.Join(zarfPkgPath, fmt.Sprintf("zarf-package-prometheus-%s-0.0.1.tar.zst", e2e.Arch))
+	e2e.CreateZarfPkg(t, zarfPkgPath, false)
```
decleaver, line 30: consider making this force create? setting to true. These packages also get created in other tests I believe, which was causing this test to fail locally.

src/test/e2e/optional_bundle_test.go
UncleGedd: good call!

src/test/e2e/optional_bundle_test.go
UncleGedd: good call!

src/test/e2e/optional_bundle_test.go
decleaver: disregard, should be a one and done, will only have to recreate this package if the zarf.yaml gets updated. So good as is.

src/test/e2e/optional_bundle_test.go
decleaver: same as previous comment. good as is

decleaver (approved):
Looks good! synchronous walk through, async review and local testing

PR #626, opened 2024-05-21
chore: removes bubbletea tui

UncleGedd:
## Description

- removes BubbleTea and related TUI code to revert back to using Zarf and `pterm`
- updates ADR


TristanHoladay (dismissed):

decleaver (approved):
will just need to let teams know once this is released that all of there uds commands where they use --no-tea will break until removed.

Issue #628, opened 2024-05-21
uds dev deploy remote bundle


decleaver:
Allow dev deployment of remote bundles

UncleGedd:
closed with https://github.com/defenseunicorns/uds-cli/pull/629

PR #629, opened 2024-05-21
feat: dev deploy remote bundles

decleaver:
## Description

allows you to run `uds dev deploy` with remote bundles

## Related Issue

Relates to #628 

## Type of change

## Checklist before merging

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,9 +288,9 @@ uds dev deploy <path-to-bundle-yaml-dir> | <oci-ref>
 
 The `dev deploy` command performs the following operations
 
-- (If local bundle) Creates Zarf packages for all local packages in a bundle
+- If local bundle: Creates Zarf packages for all local packages in a bundle
   - Creates the Zarf tarball in the same directory as the `zarf.yaml`
   - Will only create the Zarf tarball if one does not already exist
   - Ignores any `kind: ZarfInitConfig` packages in the bundle
-- (If local bundle) Creates a bundle from the newly created Zarf packages
+  - Creates a bundle from the newly created Zarf packages
 - Deploys the bundle in [YOLO](https://docs.zarf.dev/faq/#what-is-yolo-mode-and-why-would-i-use-it) mode, eliminating the need to do a `zarf init`
--- a/README.md
+++ b/README.md
@@ -278,18 +278,19 @@ UDS CLI includes a vendored version of Zarf inside of its binary. To use Zarf, s
 ## Dev Mode
 
 > [!NOTE]  
-> Dev mode is a BETA feature and currently only works with local bundles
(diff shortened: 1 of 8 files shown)


src/cmd/dev.go
```diff
@@ -25,26 +24,30 @@ var devDeployCmd = &cobra.Command{
...
-		}
+		// Check if source is a local bundle
+		localBundle := helpers.IsDir(src)
+
+		if localBundle {
+			// Create Bundle
+			setBundleFile(args)
+
+			// Ensure source ends with a slash
+			if len(src) != 0 && src[len(src)-1] != '/' {
```
UncleGedd, line 44: Why do we need the `/` at the end? (could also use `strings.HasSuffix` here)

README.md
```diff
@@ -278,18 +278,19 @@ UDS CLI includes a vendored version of Zarf inside of its binary. To use Zarf, s
...
 
 ```
-uds dev deploy <path-to-bundle-yaml-dir>
+uds dev deploy <path-to-bundle-yaml-dir> | <oci-ref>
 ```
 
 The `dev deploy` command performs the following operations
-- Creates Zarf packages for all local packages in a bundle
+
+- (If local bundle) Creates Zarf packages for all local packages in a bundle
```
UncleGedd, line 291: Let's remove the parentheses from "(If local bundle)" and just use a colon, can also remove the "(If local bundle)" below and tab the "Creates a bundle from the..." over

src/pkg/sources/remote.go
```diff
@@ -85,6 +90,16 @@ func (r *RemoteBundle) LoadPackage(dst *layout.PackagePaths, filter filters.Comp
 		}
 	}
 	addNamespaceOverrides(&pkg, r.nsOverrides)
+
+	if config.Dev {
```
UncleGedd, line 94: I think we have similar code in `sources/tarball.go` (same for the `ZarfInitConfig` check above), what do you think of refactoring these snippets to be in `sources/common.go`?

src/cmd/dev.go
decleaver: deleted.... was a band-aid for another bug. 🐛

src/pkg/sources/remote.go
```diff
@@ -85,6 +90,11 @@ func (r *RemoteBundle) LoadPackage(dst *layout.PackagePaths, filter filters.Comp
...
+		YOLOPackage(&pkg)
```
UncleGedd, line 95: Naming suggestion: `setAsYOLO`

UncleGedd:
Testing this locally, looks like I can't use `--set` to set a variable, let's add that flag (noting that setting vars through others means works fine)

PR #618, opened 2024-05-16
feat: uds config validation

decleaver:
## Description

Adds validation for unmarshalling uds-config files

## Related Issue

Fixes #552 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Other (security config, docs update, etc)

## Checklist before merging

- [X] Test, docs, adr added or updated as needed
- [ ] [Contributor Guide Steps](https://github.com/defenseunicorns/uds-cli/blob/main/CONTRIBUTING.md) followed

PR #631, opened 2024-05-22
feat: add file type variables to bundles

### LINKED ISSUE #417: Allow variable `type: file` in uds bundles

It would be nice to have the ability to pass files as variables with uds bundles (ex. cert/key files)

- If I want to override the default tls.crt/tls.key for an ingress-gateway
- I could set the TLS.CRT variable to point to a file
- and the cert/key file would get added to the ingress-gateway deployment 

(optional) A clear and concise description of any alternative solutions or features you've considered.

TristanHoladay:
## Description
Add the ability to pass files as values for bundle variables.

## Related Issue

Resolves #417 

## Type of change

## Checklist before merging


src/cmd/uds.go
```diff
@@ -56,6 +56,16 @@ var deployCmd = &cobra.Command{
 		bundleCfg.DeployOpts.Source = chooseBundle(args)
 		configureZarf()
 
+		// load uds-config if it exists
+		config := v.ConfigFileUsed()
+		if config != "" {
+			bundleCfg.DeployOpts.Config = config
+			if err := loadViperConfig(); err != nil {
```
TristanHoladay, line 63: did we remove the `loadViperConfig()` call from here in a different PR?

docs/overrides.md
```diff
@@ -242,6 +242,52 @@ Variable precedence is as follows:
...
+     ref: 0.0.1
+     overrides:
+        podinfo-component:
+          unicorn-podinfo:
+           variables:
+           - name: test_secret
+             path: "testSecret"
+             description: "Key to be set in a secret"
+             type: file
+```
```
UncleGedd, line 269: Love the example! Let's be a bit more explicit about what `raw` vs `file` means. Recommend adding another variable to the list of `variables` and say explicitly what each will resolve to (ie. variable `test-secret` will resolve to the contents of the file `testSecret` <- even better if we make this a more realistic-looking file with an actual extension)

docs/overrides.md
```diff
@@ -242,6 +242,52 @@ Variable precedence is as follows:
...
+  helm-overrides:
+    test_secret: test.pub
+```
+
+This means when `test.pub` is evalutated it will first be appended to the config path like so `src/test/bundles/07-helm-overrides/variable-files/test.pub`.
+
+If the file path is already set to the same relative path as the config, then no merging will take place.
+
+> [!NOTE]  
+> uds-cli does not encrypt or base64 encode any file contents before passing said data to Zarf or Helm.
```
UncleGedd, line 287: For consistency, let's capitalize "UDS CLI" and remove the `-`

docs/overrides.md
```diff
@@ -242,6 +242,52 @@ Variable precedence is as follows:
...
+    test_secret: test.pub
+```
+
+This means when `test.pub` is evalutated it will first be appended to the config path like so `src/test/bundles/07-helm-overrides/variable-files/test.pub`.
+
+If the file path is already set to the same relative path as the config, then no merging will take place.
+
+> [!NOTE]  
+> uds-cli does not encrypt or base64 encode any file contents before passing said data to Zarf or Helm.  
+> So if for example the file contains a key to be used in a kubernetes secret, it must be base64 encoded before given to uds-cli.
```
UncleGedd, line 288:
```suggestion
> For example, if the file contains a key to be used in a Kubernetes secret, it must be base64 encoded before being ingested by UDS CLI.
```

src/cmd/uds.go
UncleGedd: Yes, it was refactored in https://github.com/defenseunicorns/uds-cli/pull/618 and we'll follow up in https://github.com/defenseunicorns/uds-cli/issues/641

src/pkg/bundle/deploy.go
```diff
@@ -443,9 +446,19 @@ func addOverrideValue(overrides map[string]map[string]*values.Options, component
 			templatedVariable := fmt.Sprintf("%v", v)
 			value = setTemplatedVariables(templatedVariable, pkgVars)
 		}
-		// handle default case of simple values like strings and numbers
-		helmVal := fmt.Sprintf("%s=%v", valuePath, value)
-		overrides[component][chart].Values = append(overrides[component][chart].Values, helmVal)
+
+		// Check for files else handle default case of simple values like strings and numbers
+		if valueType == "file" {
```
UncleGedd, line 451: I think you could pull this whole `if` block up above the switch statement and wrap it in a function. This would remove the need for the `else` clause

TristanHoladay: i like that idea. Sorry if i'm misunderstanding the ask, but sounds like your suggesting wrapping the entire conditional in a function. wrapping everything though, even the valuteType check, allows for some misdirection in that after adding the file to the `.FileValues` it will continue on to the switch, fall through to the default case and be added to `.Values` as well. It looks like things still work out because Helm evaluates `FileValues` after `Values` so the key gets set back to the file. But maybe to avoid a potential issue and confusion, we should leave the valueType check outside of the file handler function and return if it was a file before hitting the switch.

docs/overrides.md
TristanHoladay: added a "side by side" raw example and some descriptions to more explicitly detail what's going on. let me know what you think.

src/pkg/bundle/deploy.go
UncleGedd: You nailed it, I should've used more words. You can leave the if statement as is, process the files inside it, and return

src/test/e2e/variable_test.go
```diff
@@ -373,3 +374,59 @@ func TestExportVarsAsGlobalVars(t *testing.T) {
...
+	deployZarfInit(t)
+	e2e.HelmDepUpdate(t, "src/test/packages/helm/unicorn-podinfo")
+	e2e.CreateZarfPkg(t, "src/test/packages/helm", false)
+	bundleDir := "src/test/bundles/07-helm-overrides/variable-files"
+	bundlePath := filepath.Join(bundleDir, fmt.Sprintf("uds-bundle-variable-files-%s-0.0.1.tar.zst", e2e.Arch))
+	os.Setenv("UDS_CONFIG", filepath.Join(bundleDir, "file-not-found-config.yaml"))
+
+	createLocal(t, bundleDir, e2e.Arch)
+
+	cmd := strings.Split(fmt.Sprintf("deploy %s --retries 1 --confirm", bundlePath), " ")
```
UncleGedd, line 388: Is `--retries 1` necessary?

src/pkg/bundle/deploy.go
```diff
@@ -464,3 +479,34 @@ func setTemplatedVariables(templatedVariables string, pkgVars map[string]string)
...
+	verifiedPath, err := formAndCheckFilePath(b.cfg.DeployOpts.Config, filePath)
+	if err != nil {
+		return nil, err
+	}
+	helmVal := fmt.Sprintf("%s=%v", key, verifiedPath)
+	return append(helmFileVals, helmVal), nil
+}
+
+// formAndCheckFilePath merges relative paths together to form full path and checks if the file exists
+func formAndCheckFilePath(anchorPath string, filePath string) (string, error) {
```
UncleGedd, line 494: Are we handling the case of a user setting the `default` value as opposed to using a `uds-config` or `--set`?

UncleGedd, line 494: It's idiomatic in Go to have short variable and function names (where it makes sense!). In the case of `verifyAndAddFile` and `formAndCheckFilePath`, IMO it's unnecessary to say that we're "verifying" and "checking" in the function names. I think `formandCheckFilePath` could be `formFilePath` and `verifyAndAddFile` could be `addFileValue` or something like that

src/pkg/bundle/deploy.go
```diff
@@ -153,7 +157,7 @@ func deployPackages(packages []types.Package, resume bool, b *Bundle) error {
 }
 
 // loadVariables loads and sets precedence for config-level and imported variables
-func (b *Bundle) loadVariables(pkg types.Package, bundleExportedVars map[string]map[string]string) map[string]string {
+func (b *Bundle) loadVariables(pkg types.Package, bundleExportedVars map[string]map[string]string) (map[string]string, error) {
```
decleaver, line 160: do we need to return an error? looks like we only ever return nil for the error

src/pkg/bundle/deploy.go
TristanHoladay: yep, is handled as part of `processOverrideVariables()`

src/pkg/bundle/deploy_test.go
```diff
@@ -429,3 +431,95 @@ func TestHelmOverrideVariablePrecedence(t *testing.T) {
 		})
 	}
 }
+
+func TestFileVariableHandlers(t *testing.T) {
```
UncleGedd, line 435: Love the unit tests, what do you think of adding to `TestHelmOverrideVariablePrecedence` to test the full code path starting in `processOverrideVariables`?

src/test/e2e/variable_test.go
```diff
@@ -269,13 +286,6 @@ func TestBundleWithEnvVarHelmOverrides(t *testing.T) {
 		require.NoError(t, err)
 	})
 
-	t.Run("check override secret val", func(t *testing.T) {
```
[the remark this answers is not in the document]
TristanHoladay: gotcha. adding it back and adding a different secret to the helm package chart for testing a cert passed as a file.

src/pkg/bundle/deploy_test.go
```diff
@@ -429,3 +430,176 @@ func TestHelmOverrideVariablePrecedence(t *testing.T) {
...
+	const (
```
UncleGedd, line 435: 😍

src/pkg/bundle/deploy_test.go
```diff
@@ -429,3 +430,176 @@ func TestHelmOverrideVariablePrecedence(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			os.Unsetenv("UDS_CERT")
+			if tc.loadEnv {
+				os.Setenv("UDS_CERT", fmt.Sprintf("%s/test.cert", relativePath))
+			}
+
+			overrideMap := map[string]map[string]*values.Options{}
```
UncleGedd, line 595: let's also inspect the `overrideMap` and make sure it contains the expected values. You can do this by adding another field to the table tests called "expected" or something like that

src/pkg/bundle/deploy_test.go
```diff
@@ -429,3 +431,184 @@ func TestHelmOverrideVariablePrecedence(t *testing.T) {
...
+			err := tc.bundle.processOverrideVariables(&overrideMap, tc.args.pkgName, tc.args.variables, tc.args.componentName, tc.args.chartName)
+
+			if tc.requireNoErr {
+				require.NoError(t, err)
+				require.Contains(t, tc.expected, overrideMap[componentName][chartName].FileValues[0])
```
UncleGedd, line 608:
```suggestion
				require.Equals(t, tc.expected, overrideMap[componentName][chartName].FileValues[0])
```

UncleGedd (dismissed):
really great work on this 🚀

UncleGedd (approved):

PR #636, opened 2024-05-24
chore: update to de-zarfed Maru

Racer159:
## Description

This PR updates to the new de-zarfed Maru (resolving the breakages in the library API), specifically:

Fixed Breaking Changes

* `config.EnvPrefix` is now a constant and has been replaced by `config.VendorPrefix`
* `MARU_` prefixed env vars are now always present with `UDS_` env vars needing to be added manually
* `./` replacements must be registered with supported values of `zarf`, `uds`, and `kubectl`

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging


decleaver (commented):
@Racer159 thanks for taking care of these updates. I think we can actually delete runner_test.go and the associated tasks now that maru has been handed over. @UncleGedd thoughts?

UncleGedd:
So I think this is fine for now, but we used to be able to do `uds run -a amd64` and have the arch persist down to the command. So for example:
```
tasks:
  - name: default
    description: Run Default Task
    actions:
      - cmd: echo $UDS_ARCH
```
If you run `uds run -a amd64` then "amd64" would get printed to the screen. But now that command yields my system arch. This can be worked around by doing `UDS_ARCHITECTURE=amd64 uds run` but noting that the behavior has changed slightly.

I think this is fine though because in all the cases I've seen folks are passing architecture on the `cmd`s within the task, ie something like:
```
  - name: k3d-standard-bundle
    description: "Create the K3d-UDS Core Bundle"
    actions:
      - description: "Create the UDS Core Standard Bundle"
        cmd: "uds create bundles/k3d-standard --confirm --no-progress --architecture=${ZARF_ARCHITECTURE}"
```

UncleGedd (approved):
thanks again @Racer159 !

Issue #537, opened 2024-03-29
Set refs during dev deploy


UncleGedd:
### Is your feature request related to a problem? Please describe

As Ezra
I want to be able to change the ref of a package without modifying the `uds-bundle.yaml`
So that I can test multiple versions of a package in a bundle

For dev purposes, let's add a mechanism to set a package's `ref` outside of the `uds-bundle.yaml`, this can be either through a flag, `uds-config.yaml` or some other method

PR #638, opened 2024-05-26
feat: dev deploy --ref and --flavor flags

decleaver:
## Description

Adds --ref and --flavor flags for dev deploy

## Related Issue

Relates to #537 

## Type of change

## Checklist before merging


src/cmd/dev.go
```diff
@@ -79,11 +89,31 @@ var devDeployCmd = &cobra.Command{
...
+}
+
 func init() {
 	initViper()
 	rootCmd.AddCommand(devCmd)
 	devCmd.AddCommand(devDeployCmd)
 	devDeployCmd.Flags().StringArrayVarP(&bundleCfg.DeployOpts.Packages, "packages", "p", []string{}, lang.CmdBundleDeployFlagPackages)
+	devDeployCmd.Flags().StringToStringVarP(&bundleCfg.DevDeployOpts.Ref, "ref", "r", map[string]string{}, lang.CmdBundleDeployFlagRef)
+	devDeployCmd.Flags().StringToStringVarP(&bundleCfg.DevDeployOpts.Flavor, "flavor", "f", map[string]string{}, lang.CmdBundleCreateFlagFlavor)
+	devDeployCmd.Flags().StringVar(&bundleCfg.DevDeployOpts.FlavorAll, "flavor-all", "", lang.CmdBundleCreateFlagFlavorAll)
```
UncleGedd, line 115: Instead of `--flavor-all`, what do you think of having just `--flavor` and inspecting the value to see if it should be applied to all packages or not? For example:
- `uds dev deploy ... --flavor ib` would apply to all packages in the bundle
- `uds dev deploy ... --flavor fooPkg=ib` would apply to only the "fooPkg` package

src/config/config.go
```diff
@@ -77,6 +77,9 @@ var (
 
 	// Dev specifies if we are running in dev mode
 	Dev = false
+
+	// Map of refs specified for dev deploy
+	DevDeployRefs map[string]string
```
UncleGedd, line 82: Should this be in `config` or maybe in the DeployOpts struct?

src/pkg/bundle/create.go
```diff
@@ -86,16 +85,56 @@ func (b *Bundle) Create() error {
 		}
 	}
 
+	// for dev mode update package ref for local bundles, refs for remote bundles updated on deploy
```
UncleGedd, line 88: 2 suggestions:
- I think you can remove "refs for remote bundles updated on deploy"
- Lift that second conditional up, something like `if config.Dev && len(b.cfg.DevDeployOpts.Ref) != 0`

src/cmd/dev.go
decleaver: yeah, I liked that better and it's what I had before, but cobra doesn't like it since it's expecting a map/entry. It wants a key=value

src/config/config.go
decleaver: Don't love having to define it here, but needed to be able to access it [here](https://github.com/defenseunicorns/uds-cli/pull/638/files#diff-cc6b07ec48af179448d51a500be72923462835fd2e22b0f5d6b49667c5be4ac9R49), at which point I don't have access to Bundle.BundleConfig.DevDeployOpts

UncleGedd: What do you think of adding the entire[ b.cfg](https://github.com/defenseunicorns/uds-cli/blob/8f46e31aa4803b4a44a1eda2eb466a50f5073988/src/pkg/bundle/deploy.go#L128) to the `TarballBundle` and `RemoteBundle` structs? As in, modify that `sources.New` function to take `b.cfg` instead of just the source, then we can handle `DeployOpts` and `DevDeployOpts` in the sources

src/cmd/dev.go
UncleGedd: Yeah we could get fancy and have Cobra expect a string and do the parsing ourselves, but I'm not sure if the juice is worth the squeeze, `--flavor-all` is fine for now

src/cmd/dev.go
```diff
@@ -79,11 +89,31 @@ var devDeployCmd = &cobra.Command{
...
 
+// isLocalBundle checks if the bundle source is a local bundle
+func isLocalBundle(src string) bool {
+	return helpers.IsDir(src) || strings.Contains(src, ".tar.zst")
+}
+
+// validateDevDeployFlags validates the flags for dev deploy
+func validateDevDeployFlags(isLocalBundle bool) error {
+	if !isLocalBundle {
+		//Throw error if trying to run with --flavor, --flavor-all, or --force-create flag with remote bundle
```
UncleGedd, line 100: Is this because we append the pkg root manifest to the refs in the bundle YAML once it's been bundled? (and we don't want to modify the remote bundle?)

src/config/config.go
decleaver: the only potential issue I see with that is that sources.New gets called on deploy and remove. Would need to know where to grab the source from,  from within the sources.New function (from DeployOpts.Source or RemoveOpts.Source)

src/cmd/dev.go
decleaver: yeah, flavor is applied on create so would need to recreate the remote zarf parckages with the new flavor. I'm sure this can be handled, was just going to be more complicated.

src/cmd/dev.go
```diff
@@ -79,11 +91,50 @@ var devDeployCmd = &cobra.Command{
...
+		//Throw error if trying to run with --flavor or --force-create flag with remote bundle
+		if len(bundleCfg.DevDeployOpts.Flavor) > 0 || bundleCfg.DevDeployOpts.ForceCreate {
+			return fmt.Errorf("Cannot use --flavor or --force-create flags with remote bundle")
+		}
+	}
+	return nil
+}
+
+// populateFlavorMap populates the flavor map based on the string input to the --flavor flag
+func populateFlavorMap() {
```
UncleGedd, line 111: This is great! Considering all the logic inside this fn, I think it's worth unit testing as well

src/cmd/dev.go
```diff
@@ -68,7 +80,7 @@ var devDeployCmd = &cobra.Command{
 		}
 
 		// Set dev source
-		if localBundle {
+		if isLocalBundle {
```
UncleGedd, line 83: Instead of having 2 `if isLocalBundle` clauses, can they be combined?

src/cmd/dev_test.go
```diff
@@ -0,0 +1,99 @@
+package cmd
+
+import (
+	"testing"
+
+	"github.com/defenseunicorns/uds-cli/src/types"
+	"github.com/stretchr/testify/assert"
```
UncleGedd, line 7: Recommend `require` over `assert` because `require` will fail the test faster (unless you find value in continuing the test suite after a test has failed)

src/cmd/dev.go
decleaver: good catch :)

src/pkg/bundle/create.go
```diff
@@ -86,16 +85,54 @@ func (b *Bundle) Create() error {
...
+		remote, err := zoci.NewRemote(url, platform)
+		if err != nil {
+			message.Fatalf(err, errMsg)
+		}
+		if err := remote.Repo().Reference.ValidateReferenceAsDigest(); err != nil {
+			manifestDesc, err := remote.ResolveRoot(context.TODO())
+			if err != nil {
+				message.Fatalf(err, errMsg)
+			}
+			pkg.Ref = pkg.Ref + "@sha256:" + manifestDesc.Digest.Encoded()
```
UncleGedd, line 128:
```suggestion
			pkg.Ref = ref + "@sha256:" + manifestDesc.Digest.Encoded()
```

Should this reference the `ref` passed in by the flag instead of the original `pkg.Ref`?

UncleGedd: This would also correct the bundle YAML that gets printed to the screen, right now it shows the old ref

UncleGedd: ^ maybe not actually, I keep forgetting that `dev deploy` creates local bundles

src/pkg/bundle/dev.go
```diff
@@ -26,7 +27,8 @@ func (b *Bundle) CreateZarfPkgs() {
 	}
 
 	zarfPackagePattern := `^zarf-.*\.tar\.zst$`
-	for _, pkg := range b.bundle.Packages {
+	for i := range b.bundle.Packages {
```
UncleGedd, line 30: is the indexing necessary here or can we do `for _, pkg range ...`

src/pkg/bundle/dev.go
```diff
@@ -47,8 +49,13 @@ func (b *Bundle) CreateZarfPkgs() {
 				}
 			}
 			// create local zarf package if it doesn't exist
-			if !packageFound {
-				os.Args = []string{"zarf", "package", "create", pkgDir, "--confirm", "-o", pkgDir, "--skip-sbom"}
+			if !packageFound || b.cfg.DevDeployOpts.ForceCreate {
+				if len(b.cfg.DevDeployOpts.Flavor) != 0 {
+					b.setPackageFlavor(&pkg)
```
UncleGedd, line 54: `b.setPackageFlavor` mutates `pkg`, but `b.setPackageRef` returns the `pkg` and reassigns. Personally I prefer the immutable way that `b.setPackageRef` does it, either way thought we should make it consistent unless there's a reason not to

src/pkg/bundle/dev.go
```diff
@@ -47,8 +49,13 @@ func (b *Bundle) CreateZarfPkgs() {
...
+					os.Args = []string{"zarf", "package", "create", pkgDir, "--confirm", "-o", pkgDir, "--skip-sbom", "--flavor", pkg.Flavor}
```
UncleGedd, line 55: I think you may be able to refactor some of this, looks like `--flavor` can be `""`, so `--flavor=""` is equivalent to omitting the flavor flag. (note that I tested this locally but my knowledge of flavors is limited, may be more nuance here, recommend testing more)

src/pkg/bundle/dev.go
```diff
@@ -58,6 +65,15 @@ func (b *Bundle) CreateZarfPkgs() {
 	}
 }
 
+func (b *Bundle) setPackageFlavor(pkg *types.Package) {
+	// handle case when flavor applies to all packages
+	if len(b.cfg.DevDeployOpts.Flavor) == 1 && b.cfg.DevDeployOpts.Flavor[""] != "" {
```
UncleGedd, line 70: What does `b.cfg.DevDeployOpts.Flavor[""] != ""` check for?

src/pkg/sources/new.go
```diff
@@ -17,8 +18,17 @@ import (
 )
 
 // New creates a new package source based on pkgLocation
-func New(pkgLocation string, pkg types.Package, opts zarfTypes.ZarfPackageOptions, sha string, nsOverrides NamespaceOverrideMap) (zarfSources.PackageSource, error) {
+func New(bundleCfg types.BundleConfig, pkg types.Package, opts zarfTypes.ZarfPackageOptions, sha string, nsOverrides NamespaceOverrideMap) (zarfSources.PackageSource, error) {
 	var source zarfSources.PackageSource
+	var pkgLocation string
+	if bundleCfg.DeployOpts.Source != "" {
```
UncleGedd, line 24: Is it necessary to check that the `.Source != ""` or is it possible to nil check `bundleCfg.DeployOpts`? Like, if your calling remove, is `DeployOpts` nil?

src/pkg/bundle/dev.go
UncleGedd: Have you seen a folder called "refs" randomly be created when using flavors?

src/pkg/bundle/deploy.go
```diff
@@ -54,10 +54,10 @@ func (b *Bundle) Deploy() error {
 		if len(userSpecifiedPackages) != len(packagesToDeploy) {
 			return fmt.Errorf("invalid zarf packages specified by --packages")
 		}
-		return deployPackages(packagesToDeploy, resume, b)
+	} else {
+		packagesToDeploy = b.bundle.Packages
```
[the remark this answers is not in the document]
decleaver: lol not seeing this one. We only want all the packages if the packages flag is not set. If the packages flag is set then we want to only add the specified packages to the initially empty list.

src/pkg/bundle/dev.go
decleaver: thats how i'm saving the flavor if it applies to all packages (an entry with an empty key). Added a comment for clarity.

src/pkg/sources/new.go
decleaver: Good question, but just double checked and the unused "Opts" are not nil, so on Deploy, `RemoveOpts` has `Source` set to "" and `Packages` set to an empty list.

src/pkg/bundle/dev.go
decleaver: so when we run populateFlavorMap() in src/cmd/dev.go we check for `--flavor ""` in which case we don't populate the flavor map

src/pkg/bundle/deploy.go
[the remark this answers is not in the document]
UncleGedd: u right! Noting that we refactored this over in https://github.com/defenseunicorns/uds-cli/pull/631 as well, pretty sure it's the exact same change

src/pkg/bundle/dev.go
UncleGedd: gotcha, makes sense

src/pkg/bundle/dev.go
UncleGedd: ohhh gotcha. Maybe add another comment indicating that the empty key references a value that is applied to all package flavors or something like that

src/test/packages/nginx/refs/zarf.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+components:
+  - name: nginx-remote
+    required: true
+    manifests:
+      - name: simple-nginx-deployment
+        namespace: nginx
+        files:
+          - deployment.yaml
+    actions:
```
UncleGedd, line 14: 2 suggestions:
- I think we can kill this boilerplate action
- add a description to the `metadata` saying why this pkg exists

src/test/packages/nginx/refs/zarf.yaml
```diff
@@ -0,0 +1,28 @@
...
+          - wait:
+              cluster:
+                kind: deployment
+                name: nginx-deployment
+                namespace: nginx
+                condition: available
+    # image discovery is supported in all manifests and charts using:
+    # zarf prepare find-images
+    images:
+      - docker.io/library/nginx:1.26.0
```
UncleGedd, line 28: Nice! Good call making the image be a different version, if that's not called out in the `metadata.description` I'd call it out here

src/test/packages/podinfo/refs/zarf.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - name: podinfo
+    required: true
+    charts:
+      - name: podinfo
+        version: 6.6.2
+        namespace: podinfo
+        url: https://github.com/stefanprodan/podinfo.git
+        gitPath: charts/podinfo
+    images:
+      - ghcr.io/stefanprodan/podinfo:6.6.2
```
UncleGedd, line 16: Same comment as above, make a note that we are using a different version of podinfo and why

src/test/packages/podinfo/zarf.yaml
```diff
@@ -23,3 +23,47 @@ components:
...
+      flavor: three
+    charts:
+      - name: podinfo
+        version: 6.6.3
+        namespace: podinfo-flavor
+        url: https://github.com/stefanprodan/podinfo.git
+        gitPath: charts/podinfo
+    images:
+      - ghcr.io/stefanprodan/podinfo:6.6.3
+    actions:
```
UncleGedd, line 61: I think you could kill these actions

src/test/packages/nginx/refs/zarf.yaml
UncleGedd: Disregard that first suggestion, it's probably important that we ensure the pods actually start up if we're referring to different refs

src/test/packages/podinfo/zarf.yaml
```diff
@@ -23,3 +23,47 @@ components:
                 name: podinfo
                 namespace: podinfo
                 condition: available
+
+  - name: podinfo-flavor
+    required: true
+    only:
+      flavor: two
```
UncleGedd, line 30: Pretty sure `two` here refers to the patch version in 6.6.2, could we make that more apparent in the flavor name? Something like patchVersion2 or something like that?

decleaver: updated the naming for flavor `two` and `three`

UncleGedd:
For the 0.0.2 versions of nginx and podinfo packages, let's add those to `hack/push-test-artifacts.sh`

UncleGedd:
SInce `dev deploy` is specifically a `deploy` cmd, I think we can remove the confirmation when the bundle is getting created
Screenshot 2024-06-05 at 5 11 24 PM

README.md
```diff
@@ -293,22 +293,54 @@ UDS CLI includes a vendored version of Zarf inside of its binary. To use Zarf, s
...
 
-- If local bundle: Creates Zarf packages for all local packages in a bundle
-  - Creates the Zarf tarball in the same directory as the `zarf.yaml`
-      - Can use `--flavor` flag to specify what flavor of package you want to create
-  - Will only create the Zarf tarball if one does not already exist or can use `--force-create` to force the creation of a new zarf package even if one currently exists
-  - Ignores any `kind: ZarfInitConfig` packages in the bundle
-  - Creates a bundle from the newly created Zarf packages
 - Deploys the bundle in [YOLO](https://docs.zarf.dev/faq/#what-is-yolo-mode-and-why-would-i-use-it) mode, eliminating the need to do a `zarf init`
-- For remote packages you can specify what package ref you want to deploy by using the `--ref` flag and specifying the package name and desired ref. (example: `--ref podinfo=0.2.0`)
+  - any `kind: ZarfInitConfig` packages in the bundle will be ignored
```
UncleGedd, line 307:
```suggestion
  - Any `kind: ZarfInitConfig` packages in the bundle will be ignored
```

README.md
```diff
@@ -293,23 +293,25 @@ UDS CLI includes a vendored version of Zarf inside of its binary. To use Zarf, s
...
 
-- If local bundle: Creates Zarf packages for all local packages in a bundle
-  - Creates the Zarf tarball in the same directory as the `zarf.yaml`
-  - Will only create the Zarf tarball if one does not already exist
-  - Ignores any `kind: ZarfInitConfig` packages in the bundle
-  - Creates a bundle from the newly created Zarf packages
 - Deploys the bundle in [YOLO](https://docs.zarf.dev/faq/#what-is-yolo-mode-and-why-would-i-use-it) mode, eliminating the need to do a `zarf init`
+  - any `kind: ZarfInitConfig` packages in the bundle will be ignored
+- For local bundles:
+  - Creates the Zarf tarball if one does not already exist or the `--force-create` flag can be used to force the creation of a new zarf package
```
UncleGedd, line 309:
```suggestion
  - Creates the Zarf tarball if one does not already exist or the `--force-create` flag can be used to force the creation of a new Zarf package
```

README.md
```diff
@@ -293,23 +293,26 @@ UDS CLI includes a vendored version of Zarf inside of its binary. To use Zarf, s
...
 
-- If local bundle: Creates Zarf packages for all local packages in a bundle
-  - Creates the Zarf tarball in the same directory as the `zarf.yaml`
-  - Will only create the Zarf tarball if one does not already exist
-  - Ignores any `kind: ZarfInitConfig` packages in the bundle
-  - Creates a bundle from the newly created Zarf packages
 - Deploys the bundle in [YOLO](https://docs.zarf.dev/faq/#what-is-yolo-mode-and-why-would-i-use-it) mode, eliminating the need to do a `zarf init`
+  - Any `kind: ZarfInitConfig` packages in the bundle will be ignored
+- For local bundles:
+  - Local Packages:
```
UncleGedd, line 309:
```suggestion
  - For local packages:
```

UncleGedd (approved):
🚀

PR #643, opened 2024-05-29
fix: ensures partial pkgs are correct and adds smoke test to workflows

UncleGedd:
## Description

ensures partial pkgs are correct and adds smoke test to workflows


decleaver (approved):
:pear:'d

PR #645, opened 2024-05-29
chore: run CI tests in parallel, remove alpha status

UncleGedd:
## Description

run CI tests in parallel


decleaver (dismissed):
yes please

PR #646, opened 2024-05-29
feat: integrated docs between uds and uds-cli

UnicornChance:
## Description
Part of the move to setting up the integrated docs means formatting existing docs in a hugo style so that everything can be nice and pretty when pulled together. This is that.

All of the new docs are copy and paste from the `uds` repo. A follow on PR in the `uds` repo will remove those duplicated docs.

[PR that was just merged in the `uds` repo.](https://github.com/defenseunicorns/uds/pull/389) which will now pull docs from individual repos into the `uds` repo.

## Type of change

## Checklist before merging


UncleGedd:
Thanks for this @UnicornChance , still poking around. Any reason `runner.md` was removed?

UnicornChance:
@UncleGedd my focus was on keeping things as similar to the original `uds-docs` as possible. But since no one seems to want to own that, we can probably do whatever we want

decleaver (approved):

PR #657, opened 2024-06-04
fix: arch check in PreDeployValidation

### LINKED ISSUE #656: Check bundle arch against cluster arch
### Steps to reproduce

Assume my cluster arch is amd64 and I'm deploying from my arm64 Mac. Consider the following:
```yaml
# uds-bundle.yaml
kind: UDSBundle
metadata:
  name: ghcr-test
  description: testing a bundle of pkgs from ghcr
  version: 0.0.1

packages:
  - name: podinfo
    repository: ghcr.io/defenseunicorns/uds-cli/podinfo
    ref: 0.0.1
  - name: nginx
    repository: ghcr.io/defenseunicorns/uds-cli/nginx
    ref: 0.0.1

---
# tasks.yaml
  tasks:
  - name: default
    actions:
      - cmd: ./uds create . -a amd64 --confirm --no-progress
      - task: deploy

  - name: deploy
    actions:
      - cmd: |
          set -x
          ./uds deploy $(ls -tr uds-bundle*.tar.zst | tail -1) --confirm --no-progress

```

### Expected result

When running `uds run` the bundle creates and deploys successfully

### Actual Result

Bundle create works but on deploy we get this error:
```
          ERROR:  Failed to validate bundle: arch arm64 does not match cluster arch, [amd64]
     ERROR:  Failed to run action: command "set -x; ./uds deploy $(ls -tr uds-bundle*.tar.zst | tail ..." timed
             out after 0 seconds
```

It appears that UDS CLI is checking the user's system arch against the cluster arch

### Additional Context
The workaround is to simply add `-a amd64` to the `uds deploy` command, but this behavior is still bad. During deploy we should check the bundle's arch against the cluster's arch

TristanHoladay:
## Description
Fix PreDeployValidation() arch check to look at the bundle arch set at create time.

## Related Issue

Fixes #656 

## Type of change

## Checklist before merging

### DIFF
--- a/src/test/common.go
+++ b/src/test/common.go
@@ -166,9 +166,9 @@ func (e2e *UDSE2ETest) DownloadZarfInitPkg(t *testing.T, zarfVersion string) {
 }
 
 // CreateZarfPkg creates a Zarf package in the given path
-func (e2e *UDSE2ETest) CreateZarfPkg(t *testing.T, path string, forceCreate bool) {
+func (e2e *UDSE2ETest) CreateZarfPkg(t *testing.T, path string, forceCreate bool, arch string) {
 	//  check if pkg already exists
-	pattern := fmt.Sprintf("%s/*-%s-*.tar.zst", path, e2e.Arch)
+	pattern := fmt.Sprintf("%s/*-%s-*.tar.zst", path, arch)
 	matches, err := filepath.Glob(pattern)
 	require.NoError(t, err)
 	if !forceCreate && len(matches) > 0 {
--- a/src/test/common.go
+++ b/src/test/common.go
@@ -166,7 +166,22 @@ func (e2e *UDSE2ETest) DownloadZarfInitPkg(t *testing.T, zarfVersion string) {
 }
 
 // CreateZarfPkg creates a Zarf package in the given path
-func (e2e *UDSE2ETest) CreateZarfPkg(t *testing.T, path string, forceCreate bool, arch string) {
+func (e2e *UDSE2ETest) CreateZarfPkg(t *testing.T, path string, forceCreate bool) {
+	//  check if pkg already exists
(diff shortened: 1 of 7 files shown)


src/test/common.go
```diff
@@ -166,9 +166,9 @@ func (e2e *UDSE2ETest) DownloadZarfInitPkg(t *testing.T, zarfVersion string) {
 }
 
 // CreateZarfPkg creates a Zarf package in the given path
-func (e2e *UDSE2ETest) CreateZarfPkg(t *testing.T, path string, forceCreate bool) {
+func (e2e *UDSE2ETest) CreateZarfPkg(t *testing.T, path string, forceCreate bool, arch string) {
```
TristanHoladay, line 169: i originally create a duplicate function with the arch param, but then switched this one to take the param. thoughts?

TristanHoladay: ended up doing the duplicate func after all to pass the arch as a cli flag to the create command.

src/test/e2e/bundle_test.go
```diff
@@ -624,3 +624,16 @@ func TestInvalidBundle(t *testing.T) {
 	stderr := createLocalError(bundleDir, e2e.Arch)
 	require.Contains(t, stderr, "unknown field")
 }
+
+func TestArchCheck(t *testing.T) {
```
UncleGedd, line 628: This test passes on my arm64 Mac, may need a bit more logic to ensure we're setting the proper arch depending on the system on which this test runs

TristanHoladay: mmm, yeah i wondered if something like that was going to happen. Thanks @UncleGedd i'll fix it.

TristanHoladay: interesting i saw that too in a previous CI run, but couldn't figure out where it was coming from so i re-ran the job and it went away.

UncleGedd (approved):

Issue #655, opened 2024-06-03
Add ability to view override and config options for a given bundle prior to deployment


zachariahmiller:
### Is your feature request related to a problem? Please describe

As a bundle creator i would like to be able to have the consumer of my bundle be able to ascertain the specific configuration options they have available to them. While this could totally be readme documentations it would be nice for it to be something self-contained to the bundle definition and uds cli as they might be deploying the bundle in an environment without network access. Currently all configuration options, zarf vars, helm overrides, etc are available to the uds-cli as uds vars when reading in the bundle. As such i think it makes sense to have some command in uds-cli to point to a bundle and be able to have the CLI output the given configuration options available to the bundle deployer.

TristanHoladay:
There's currently `uds inspect` and the pre-deploy view when deploying a bundle that shows some of this info, but not all and not in a great format for user experience. 

@Madeline-UX is tracking this with us now and we're thinking this potentially is not only a "show me what i can configure" feature but also a "let me check my configuration" feature to make sure things are set the way you expected (similar to `helm template`). So the 3 questions we're seeking to answer from a user's view are:

1) What overrides are possible / what has been overridden?
2) What is the value that will be deployed (by default or if the user changes)?
3) What is the source of that value (eg. config, cli, env vars, defaults in the bundle yaml)

Madeline-UX:
Ideating through this design sketch - to help better indicate what variable values are being overridden and the source of those overrides. 
![IMG_7338](https://github.com/defenseunicorns/uds-cli/assets/92826525/c90f657c-1238-4bdd-81ee-7d0d262f34f3)

TristanHoladay:
Notes from discussions with @Madeline-UX and @zachariahmiller:

There is a desire to see all possible configurations (bundle helm overrides, zarf vars, zarf helm overrides), but there are issues with showing everything in terms of user experience (could overload users, could show a lot of stuff they don't care about). And the pre-deploy view, while a good update, doesn't necessarily help when most users are probably not stopping to read through that output before actually deploying. It could be more helpful to have a separate command as well for getting this info.

So next steps look like:

1. keep pre-deploy view changes, except remove color coding for now.

1. Mads is going to talk with some Delivery people to understand their process and what's beneficial to them.

1. potentially modify `uds inspect` to output bundle overrides and maybe a flag or some other mechanism for opting for showing all zarf package variables as well.

UncleGedd:
[...]

PR #674, opened 2024-06-07
feat: show overrides for deployers

TristanHoladay:
## Description
Adds user friendly path and view for discovering configurable overrides when deploying a bundle.

## Related Issue

Relates to #655 

## Type of change

## Checklist before merging


TristanHoladay:
pre-deploy view:
Screenshot from 2024-06-24 20-45-40


`uds inspect --list-variables ...` : 
Screenshot from 2024-06-24 20-45-08

src/cmd/dev.go
```diff
@@ -62,11 +62,12 @@ var devDeployCmd = &cobra.Command{
 		configureZarf()
 
 		// load uds-config if it exists
-		if v.ConfigFileUsed() != "" {
+		if config := v.ConfigFileUsed(); config != "" {
```
UncleGedd, line 65: nice refactor

src/pkg/bundle/deploy.go
```diff
@@ -34,6 +36,7 @@ type PkgOverrideMap map[string]map[string]map[string]interface{}
 
 // templatedVarRegex is the regex for templated variables
 var templatedVarRegex = regexp.MustCompile(`\${([^}]+)}`)
+var includedSources = make(map[types.ValueSources]types.ValueSources, 0)
```
UncleGedd, line 39: I think you can remove the `0` arg here + include a comment for this global var

src/pkg/bundle/deploy.go
```diff
@@ -77,7 +80,7 @@ func deployPackages(packages []types.Package, resume bool, b *Bundle) error {
 		packagesToDeploy = packages
 	}
 
-	// deploy each package
+	// setup each package client
```
UncleGedd, line 83:
```suggestion
	// setup each package client and deploy
```

src/pkg/bundle/deploy.go
```diff
@@ -354,6 +285,7 @@ func (b *Bundle) processOverrideValues(overrideMap *map[string]map[string]*value
 
 // processOverrideVariables processes bundle variables overrides and adds them to the override map
 func (b *Bundle) processOverrideVariables(overrideMap *map[string]map[string]*values.Options, pkgName string, variables *[]types.BundleChartVariable, componentName string, chartName string) error {
+	// processedVars := getVarValues(pkgName, variables, b)
```
UncleGedd, line 288:
```suggestion
```

src/pkg/bundle/deploy.go
```diff
@@ -366,47 +298,48 @@ func (b *Bundle) processOverrideVariables(overrideMap *map[string]map[string]*va
 				setVal := strings.Split(k, ".")
 				if setVal[0] == pkgName && strings.ToUpper(setVal[1]) == v.Name {
 					overrideVal = val
-					v.Source = b.getSourcePath(types.CLI)
+					v.Source = types.CLI
```
UncleGedd, line 301: Instead of having these consts in the `types` package, since they are only used in the `bundle` package, what do you think of defining the `ValueSources` type + the constants in the `bundle` package? IMO `types.CLI` looks odd

src/pkg/bundle/deploy.go
UncleGedd: On second thought, can we make `includedSources` part of the DeployOpts?

UncleGedd: even better to find a way to remove the global var entirely

src/pkg/bundle/deploy.go
```diff
@@ -541,3 +474,170 @@ func formFilePath(anchorPath string, filePath string) (string, error) {
...
+	bundleName := b.bundle.Metadata.Name
+	return bundleName, string(bundleYAML), source, err
+}
+
+type PkgView struct {
+	meta      map[string]string
+	overrides map[string]interface{}
+}
+
+// ConfirmBundleDeploy uses Zarf's pterm logging to prompt the user to confirm bundle creation
```
UncleGedd, line 532:
```suggestion
// ConfirmBundleDeploy prompts the user to confirm bundle creation
```
More of a Zarf `message` thing than `pterm`

src/pkg/bundle/deploy.go
```diff
@@ -541,3 +474,170 @@ func formFilePath(anchorPath string, filePath string) (string, error) {
...
+						val := processedVars[paths[0]]
+						for i := range paths[1:] {
+							if val == nil {
+								val = "not set"
+								break
+							}
+							val = val.(map[string]interface{})[paths[i+1]]
+						}
+
+						// var.helm.path: val
```
UncleGedd, line 624:
```suggestion
```

src/pkg/bundle/inspect.go
```diff
@@ -99,6 +109,35 @@ func (b *Bundle) listImages() error {
...
+		source, err := sources.New(*b.cfg, pkg, zarfTypes.ZarfPackageOptions{}, sha, nil)
+		if err != nil {
+			return err
+		}
+
+		tmpDir, err := zarfUtils.MakeTempDir(config.CommonOptions.TempDirectory)
+		if err != nil {
+			return err
+		}
+		pkgPaths := layout.New(tmpDir)
```
decleaver, line 128: not sure if I'm missing it, but are the tmpDirs being cleaned up? If not, you can add `defer os.RemoveAll(tmpDir)` here

src/pkg/bundle/deploy.go
TristanHoladay: i was thinking about that yesterday for a moment, cause i agree i don't like the global vars stuff. I'll take another look.

src/pkg/bundle/inspect.go
TristanHoladay: nice catch. gonna add to `getPackageImages()` as well in that case.

UncleGedd:
```[tasklist]
### Items to address
- [ ] Remove "Config" view
- [ ] Show Helm chart
- [ ] Swap paths (dot notation) -> variable names
- [ ] Remove items from the variable view if they aren't set
```

TristanHoladay:
Screenshot from 2024-06-26 21-08-39

docs/quickstart-and-usage.md
```diff
@@ -97,6 +97,14 @@ By default all the packages in the bundle are deployed, regardless of if they ha
 
 As an example: `uds deploy uds-bundle-<name>.tar.zst --resume`
 
+#### Pre-Deploy View
+
+When `deploy` is executed, the bundle's metadata, along with a list of its packages and each package's overrides or Zarf variables, will be outputted to the terminal. Unlike [`inspect --list-variables`](#viewing-variables), this output will show the value set for each override or Zarf variable. Overrides and variables that have not been set will not show in the output.
```
UncleGedd, line 102:
```suggestion
When `uds deploy` is executed, the bundle's metadata, along with a list of its packages and each package's overrides and Zarf variables, will be outputted to the terminal. Unlike [`inspect --list-variables`](#viewing-variables), this output will show the value set for each override or Zarf variable. Overrides and variables that have not been set will not be shown in the output.
```

docs/quickstart-and-usage.md
```diff
@@ -119,7 +127,13 @@ There are 2 additional flags for the `uds inspect` command you can use to extrac
...
 - Output SBOMs into a directory as files: `uds inspect ... --sbom --extract`
 
-This functionality will use the `sboms.tar` of the  underlying Zarf packages to create new a `bundle-sboms.tar` artifact containing all SBOMs from the Zarf packages in the bundle.
+This functionality will use the `sboms.tar` of the underlying Zarf packages to create new a `bundle-sboms.tar` artifact containing all SBOMs from the Zarf packages in the bundle.
+
+#### Viewing Variables
+
+To view the configurable overrides and Zarf variables of a bundle's packages:
+
+`uds inspect --list-variables <local-or-remote-tarball>:ref`
```
UncleGedd, line 136: Slight correction, when running `uds inspect --list-variables` on a local tarball, no need to include the ref

docs/overrides.md
```diff
@@ -319,18 +321,22 @@ It's also possible to specify a namespace for a packaged Helm chart to be instal
...
+        unicorn-podinfo:
+          namespace: custom-podinfo # custom namespace!
+          values:
+            - path: "podinfo.replicaCount"
+              value: 1
 ```
+
+### View All Variables
+
+When working with a local or remote bundle you can view all overrides and zarf variables by running `uds inspect --list-variables <tarball>:ref`
```
UncleGedd, line 342: Slight correction, when running uds inspect --list-variables on a local tarball, no need to include the ref

UncleGedd: I think you can do something like `uds inspect --list-variables BUNDLE_TARBALL|OCI_REF]`

docs/quickstart-and-usage.md
UncleGedd: same comment as above

src/pkg/bundle/deploy.go
```diff
@@ -512,15 +425,32 @@ func setTemplatedVariables(templatedVariables string, pkgVars map[string]string)
...
+	verifiedPath, err := formFilePath(getSourcePath(override.Source, b), filePath)
 	if err != nil {
 		return nil, err
 	}
 	helmVal := fmt.Sprintf("%s=%v", override.Path, verifiedPath)
 	return append(helmFileVals, helmVal), nil
 }
 
+// getSourcePath returns the path from where a value is set
+func getSourcePath(pathType valuesources.Source, b *Bundle) string {
```
UncleGedd, line 438: any reason to not make this a receiver function as well? (as opposed to passing in `b`)

UncleGedd:
On second review, discussed a few more action items:
- collapse `Overrides` and `Zarf-Variables` into a single key called `variables`
- make an issue to refactor the e2e tests into unit tests
- refactor portion of `loadChartOverrides` to remove complex path render logic
- comment on change name of `filterOutOverrides`

TristanHoladay:
@UncleGedd maybe there's a better way to collapse the overrides and variables but it's looking a little odd considering that the zarf vars are not attached to a particular component/chart like the overrides.

pre-deploy view:
Screenshot from 2024-06-27 20-29-49

inspect:
Screenshot from 2024-06-27 20-15-21

src/pkg/bundle/deploy.go
```diff
@@ -630,11 +631,11 @@ func formPkgViews(b *Bundle) []PkgView {
...
 						// removing the entry if map[path] returns nil
 						chartVars[v.Name] = processedVars
-						for _, p := range paths {
+						for _, path := range paths {
 
-							val, ok := chartVars[v.Name].(map[string]interface{})[p]
-							if !ok {
-								// delete previously set entry of v.Name
+							val, notNil := chartVars[v.Name].(map[string]interface{})[path]
+							if !notNil {
```
UncleGedd, line 637:
```suggestion
							val, exists := chartVars[v.Name].(map[string]interface{})[path]
							if !exists {
```

src/pkg/bundle/deploy.go
```diff
@@ -541,3 +485,200 @@ func formFilePath(anchorPath string, filePath string) (string, error) {
...
+			}
+		}
+
+		pkgViews = append(pkgViews, PkgView{meta: pkgMeta, overrides: map[string]interface{}{"overrides": variables}})
+	}
+	return pkgViews
+}
+
+func filterOverrides(chartVars []types.BundleChartVariable, varMap map[string]overrideView) map[string]overrideView {
+	varMapCopy := varMap
```
UncleGedd, line 677: Maps are passed by reference in Go. This means that `varMapCopy` points to the same underlying data as `varMap`, and changes to `varMapCopy` affect `varMap`. 

Also, what exactly are we filtering here? Looks like sometimes values are an empty string?

src/pkg/bundle/deploy.go
```diff
@@ -541,3 +485,200 @@ func formFilePath(anchorPath string, filePath string) (string, error) {
...
+					continue
+				}
+
+				chartVars := map[string]interface{}{}
+
+				for _, v := range chart.Variables {
+
+					// Mask potentially sensitive variables
+					if v.Type == "file" || v.Source == valuesources.Env {
+						chartVars[v.Name] = "****"
```
UncleGedd, line 623: I noticed that we set `v.Name` to `****`, but is the name actually sensitive or just the value?

UncleGedd: silly comment, disregard

src/pkg/bundle/deploy.go
```diff
@@ -541,3 +485,200 @@ func formFilePath(anchorPath string, filePath string) (string, error) {
...
+	return true
+}
+
+type PkgView struct {
+	meta      map[string]string
+	overrides map[string]interface{}
+}
+
+// formPkgViews creates a unique pre deploy view of each packages set overrides and Zarf variables
+func formPkgViews(b *Bundle) []PkgView {
```
decleaver, line 588: Make sense to break this f'n up? Nearing a hundred lines, could help future us with readability :)

src/pkg/bundle/deploy.go
TristanHoladay: hmm. ok yeah so we'd have to make a deep copy. even though i prefer immutability i don't know if it matters much here and is worth the effort / performance. maybe the fact that we're returning and resetting the `filteredVars` is enough to indicate the map is being updated. 

so we're filtering out entries generated during `loadVariables` that are associated with a bundle override variable. We should be left with only the Zarf Variables that were set. if we find an actual entry (not empty string)based on the bundle override variable name then we want to delete it. i'll work on making this more explicit.

decleaver (approved):
good stuff! appreciate the updates

UncleGedd (changes requested):
Love the changes to `filterOverrides` (now `removeOverides`). I think we still have the issue of that not being a pure function. When we pass in `loadedVariables` and modify it, we're modifying the underlying map, which is eventually passed to `deploy`. Hard to tell if there's actually a bug here, but IMO it's unexpected behavior. Here's a small example showing what I mean: https://go.dev/play/p/bc0uUlOK3o0

Let's go over the rest this morning

src/pkg/bundle/deploy.go
```diff
@@ -541,3 +485,215 @@ func formFilePath(anchorPath string, filePath string) (string, error) {
...
+	return pkgMeta
+}
+
+func addZarfVars(setZarfVars map[string]overrideView, variables []interface{}) []interface{} {
+	for key, fv := range setZarfVars {
+		// config is location of uds-config.yaml, not a variable
+		// Mask potentially secret ENV vars
+		if key != "CONFIG" {
+			if fv.source == valuesources.Env {
+				fv.value = "****"
```
UncleGedd, line 641: Let's make "***" into a constant because we use it in several places

src/pkg/bundle/deploy_test.go
```diff
@@ -616,3 +616,197 @@ func TestFileVariableHandlers(t *testing.T) {
...
+
+	t.Run(zarfVarTest.name, func(t *testing.T) {
+		pkgViews := formPkgViews(&zarfVarTest.bundle)
+
+		v := pkgViews[0].overrides["overrides"].([]interface{})[0]
+
+		require.Contains(t, v.(map[string]string)[zarfVarTest.expectedKey], zarfVarTest.expectedVal)
+	})
+
+	nilCheckTests := []TestCase{
```
UncleGedd, line 770: Not quite understanding the `nilCheckTests` naming

src/pkg/bundle/deploy_test.go
```diff
@@ -616,3 +616,197 @@ func TestFileVariableHandlers(t *testing.T) {
...
+			require.Contains(t, v.(map[string]interface{})[tc.expectedKey], tc.expectedVal)
+
+			// cleanup ENV Vars to avoid conflicts
+			if tc.loadEnv {
+				os.Unsetenv(tc.envKey)
+			}
+		})
+	}
+
+	zarfVarTest := TestCase{
```
UncleGedd, line 743: I can't seem to run this test in my editor. Is it possible to combine all the test cases into a single list of `[]TestCase`?

src/pkg/bundle/deploy_test.go
UncleGedd: done

decleaver (approved):
👍

PR #677, opened 2024-06-10
Spike: adding repository pattern

UncleGedd:
## Description

Exploring what a repository pattern looks like with Svelte to pass fake data in and encapsulate "real" data calls


BillyFigueroa:
@UncleGedd just took a quick look over the code. Can tell setup linting for the UI? We have a prettier config setup so if you install the prettier plugin in vscode or if you work on the terminal, it will auto format for you. The files should have 2 spaces for tabs and convert tabs to spaces

ui/src/routes/(platform)/cluster/pods/pod.test.ts
```diff
@@ -1,18 +1,18 @@
...
-        await waitFor(() => expect(getByText('pod1')).toBeDefined());
-        expect(getByText('pod2')).toBeDefined();
+  test('pod rendering', async () => {
+    const { getByText } = render(Page, {
+      data: {
+        repo: new FakePodRepository()
+      }
     });
+
+    await waitFor(() => expect(getByText('pod1')).toBeDefined());
```
BillyFigueroa, line 15: Does using `waitFor` and `getByText` here work any better than using `await` and `findByText`?

i.e.
```JavaScript
expect(await findByText('pod1')).toBeDefined()
```

UncleGedd: Good call, will update to `await screen.findByText('pod1');`

PR #682, opened 2024-06-10
chore: uds engine installation adr

decleaver:
## Description

ADR for UDS Engine installation


jeff-mccoy (commented):
Im not sure we need a unique way to dev for UDS engine in the CLI vs what all the other UDS Core apps do. Recommend we keep the dev flow for UDS engine separate for true dev things and not introduce a new command for this.

UncleGedd (approved):

PR #696, opened 2024-06-12
feat: uds monitor pepr

jeff-mccoy:
## Description

Introduces the `uds monitor` command tree, with the first command being `uds monitor pepr`. This provides the same capabilities as `npx pepr monitor` from Pepr, and many more capabilities specific to UDS. Commands:

`uds monitor pepr`
Aggregates all admission and operator logs into a single stream

`uds monitor pepr operator`
Stream UDS Operator actions (Package processing, status updates, and errors)

`uds monitor pepr policies`
Stream UDS Policy logs (Allow, Deny, Mutate)

`uds monitor pepr allowed`
Stream UDS Policy allow logs

`uds monitor pepr denied`
Stream UDS Policy deny logs

`uds monitor pepr mutated`
Stream UDS Policy mutation logs

`uds monitor pepr failed`
Stream UDS Policy deny logs and UDS Operator error logs

Flags:

`-n, --namespace` - Namespace to monitor. Default is all namespaces

`-f, --follow` - Follow the logs

`-t, --timestamps` - Show timestamps in logs. Default is false

`--since` - Only return logs newer than a relative duration like 5s, 2m, or 3h. Defaults to all logs

`--json` - Return the raw JSON output that would be processed the command (extra log lines filtered out)

#### With timestamps
Screenshot 2024-06-12 at 11 32 20 PM

#### Without timestamps
Screenshot 2024-06-12 at 11 33 38 PM

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


src/cmd/monitor/monitor.go
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023-Present The UDS Authors
+
+// Package cmd contains the CLI commands for UDS.
+package monitor
```
UncleGedd, line 5:
```suggestion
// Package monitor contains the CLI commands for UDS monitor.
package monitor
```

src/cmd/monitor/pepr.go
```diff
@@ -0,0 +1,89 @@
...
+
+		// Set the stream flags
+		peprReader.JSON = json
+		peprReader.FilterStream = streamKind
+		peprStream.Follow = follow
+		peprStream.Since = since
+
+		// Start the stream
+		if err := peprStream.Start(); err != nil {
+			log.Fatalf("Error streaming Pepr logs: %v", err)
```
UncleGedd, line 77:
```suggestion
			message.Fatalf("Error streaming Pepr logs: %v", err)
```
^ will also need to bring in the import

src/pkg/engine/pepr/stream.go
```diff
@@ -0,0 +1,436 @@
...
+	// FailureStream represents all admission controller deny logs and operator failure logs
+	FailureStream StreamKind = "failed"
+)
+
+// NewStreamReader creates a new PeprStreamReader
+func NewStreamReader(timestamp bool, filterNamespace, filterName string) *StreamReader {
+	// Use a longer indent when timestamps are enabled
+	indent := ""
+	if timestamp {
+		indent = "                     "
```
UncleGedd, line 95: ~This is fine for now and we can make a follow up issue, but it'd be nice to get the size of the terminal and make this a bit more responsive in smaller terminals~

I take it back, this looks fine even in small terminals (also they can just resize and it'll be fine)

src/pkg/engine/stream/common.go
```diff
@@ -0,0 +1,134 @@
...
+			// Get the log stream for the pod
+			logStream, err := s.Client.CoreV1().Pods(s.Namespace).GetLogs(podName, podOpts).Stream(context.TODO())
+			if err != nil {
+				log.Printf("Error streaming logs for pod %s: %v", podName, err)
+				return
+			}
+			defer logStream.Close()
+
+			if err := s.reader.LogStream(s.writer, logStream); err != nil {
+				log.Printf("Error streaming logs for pod %s: %v", podName, err)
```
UncleGedd, line 100: Nice, I like the decision to simply log out errors instead of blowing up

src/pkg/engine/pepr/stream_test.go
```diff
@@ -0,0 +1,394 @@
...
+)
+
+func TestNewStreamReader(t *testing.T) {
+	timestamp := true
+	filterNamespace := "test-namespace"
+	filterName := "test-name"
+
+	reader := NewStreamReader(timestamp, filterNamespace, filterName)
+
+	assert.True(t, reader.showTimestamp)
```
UncleGedd, line 22: Generally speaking, we prefer `require` over `assert` to ensure the test fails immediately, but if you think `assert` is appropriate here I'm to that as well

UncleGedd:
@jeff-mccoy I've looked through this PR and everything is generally easy to follow/understand. My only real callout is the tight coupling between the stream logic and Pepr logs ([example](https://github.com/defenseunicorns/uds-cli/blob/88506389db94b0054c1b4ad98f159f47390e2bb6/src/pkg/engine/pepr/stream.go#L172)). Understood this is a necessary evil (we are streaming logs after all 😅 ), just want to ensure @defenseunicorns/pepr are tracking this coupling as well

src/pkg/engine/pepr/stream.go
jeff-mccoy: @Madeline-UX and I discussed that a lot, first thing is they don't have to enable timestamps and it's more narrow, second it does wrap and we try to add enough line breaks to make it acceptable still.

UncleGedd:
@jeff-mccoy I went through this code with @TristanHoladay and @decleaver, and discussed with some of the UDS Core folks as well, TLDR: we're g2g

I have some nits that can be addressed in a follow-up ticket along with adding some e2e tests (would be good for the UDS CLI team to get some more practice with those `Package` CRs anyway). I'll also address the test linting + adding docs in that follow-on issue

UncleGedd (dismissed):

UncleGedd (approved):
🚀 

Follow up ticket to add docs + e2e tests

Issue #653, opened 2024-06-03
Caching Refactor


decleaver:
Currently we are handling caching in two different ways. We have the caching that we are managing in https://github.com/defenseunicorns/uds-cli/blob/main/src/pkg/cache/cache.go and we have the caching that we get for free with oras.  

The issue with the oras caching is that we randomly but consistently run into issues where the cache becomes "corrupt" and we end up having to clear it out. This may be due to the fact that it is caching everything including config, bundle root manifest, zarf manifest, index.json, and it seems that if any of those get modified or don't match what is expected, the cache must be cleared in order to get things working again.

The goal is to be able to consolidate the way we are caching.

- [x] Create ADR for caching

PR #707, opened 2024-06-17
chore: adr for caching implementation

decleaver:
## Description

ADR

## Related Issue

Relates to #653 

## Type of change

## Checklist before merging


adr/0007-caching.md
```diff
@@ -0,0 +1,39 @@
+# 7. Caching
+
+Date: 17 June 2024
+
+## Status
+IN PROGRESS
+
+## Context
+UDS interacts with both local and remote bundles that are made up of both local and remote zarf packages. When dealing with remote artifacts, the process can be slow due to the time it takes to download the artifacts. We want to speed up that process by caching the artifacts. There are different ways to cache these artifacts, and we need to decide which one to use. We want to take into consideration the following: efficiency, ease of implementation, and code readability. We want to be able to leverage the same caching  implementation everywhere it makes sense. We also want the caching implementation to be resilient and able to recover from any corruption.
```
UncleGedd, line 9:
```suggestion
UDS interacts with both local and remote bundles that are made up of both local and remote zarf packages. When dealing with remote artifacts, the process can be slow due to the time it takes to download the artifacts. We want to speed up that process by caching the layers of the artifacts. There are different ways to cache these artifacts, and we need to decide which one to use. We want to take into consideration the following: efficiency, ease of implementation, and code readability. We want to be able to leverage the same caching  implementation everywhere it makes sense. We also want the caching implementation to be resilient and able to recover from any corruption.
```

adr/0007-caching.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Options
+
+### ORAS Content-Addressable Storage (CAS) as Cache
+By leveraging how the oras-go library copies a rooted directed acyclic graph (DAG) from the source CAS to the destination CAS we can treat the destination CAS as a local storage cache if we make that destination CAS persistent. While traversing the graph, for each node, it checks if it already exists in the destination. If so, it skips the node. If the node doesn't exist in the destination, it finds the successors of the node. If the node has successors, it processes the successors first. After all successors are processed, it checks if the node exists in the in-memory cache. If so, it copies the node from the in-memory cache to the destination. If not, it copies the node from the source to the destination.
```
UncleGedd, line 14:
```suggestion
By leveraging how the oras-go library copies a rooted directed acyclic graph (DAG) from the remote source CAS to the destination CAS we can treat the destination CAS as a local storage cache if we make that destination CAS persistent. While traversing the graph, for each node, it checks if it already exists in the destination. If so, it skips the node. If the node doesn't exist in the destination, it finds the successors of the node. If the node has successors, it processes the successors first. After all successors are processed, it checks if the node exists in the in-memory cache. If so, it copies the node from the in-memory cache to the destination. If not, it copies the node from the remote source to the destination.
```

UncleGedd, line 14: Can we drop a link to oras-go?

UncleGedd:
Awesome! Couple of suggestions:
- Add some context for the current state of our caching (ie. how we're using both options right now)
- Describe the chosen solution is more detail; discuss the `.uds-cache` dir and what layers we will/won't cache and why

UncleGedd (dismissed):

TristanHoladay (dismissed):

PR #717, opened 2024-06-19
fix(engine): use k8s timestamps instead of pepr timestamps

jeff-mccoy:
## Description

This PR uses the native K8s timestamp feature (which is the [CRI Log format](https://github.com/kubernetes/design-proposals-archive/blob/main/node/kubelet-cri-logging.md) instead of the timestamp from the pod. This ensure behavior consistent with normal K8s log output, reduces the need to parse each timestamp and eleminates an issue that exists for timestamp parsing currently.

## Type of change

### DIFF
--- a/src/pkg/engine/pepr/stream.go
+++ b/src/pkg/engine/pepr/stream.go
@@ -12,7 +12,6 @@ import (
 	"io"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/defenseunicorns/uds-cli/src/pkg/style"
 	"github.com/defenseunicorns/zarf/src/pkg/message"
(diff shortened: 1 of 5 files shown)


src/pkg/engine/pepr/stream.go
```diff
@@ -149,7 +139,12 @@ func (p *StreamReader) PodFilter(pods []corev1.Pod) map[string]string {
 }
 
 // LogStream processes the log stream from Pepr and writes formatted output to the writer
-func (p *StreamReader) LogStream(writer io.Writer, logStream io.ReadCloser) error {
+func (p *StreamReader) LogStream(writer io.Writer, logStream io.ReadCloser, timestamp bool) error {
+	// Use a longer indent when timestamps are enabled
+	if timestamp {
+		p.indent = "                                "
```
UncleGedd, line 145:
```suggestion
		p.indent = strings.Repeat(" ", 32)
```

UncleGedd (approved):
🚀 appreciate the context lifting and improved timestamp handling

Issue #733, opened 2024-06-27
Remove color from --list-images arg


UncleGedd:
### Description

Adding that beautiful magenta to the output of `--list-images` make its difficult for other automation to read, remove this color in favor raw output

PR #734, opened 2024-06-27
fix: ensure inspected image list is machine readable

UncleGedd:
## Description

Fixes https://github.com/defenseunicorns/uds-cli/issues/733

Noting that this is a breaking change as `--list-images` will now only go to stdout


decleaver (approved):
👍

PR #797, opened 2024-07-15
fix: intermittent unit test failure

### LINKED ISSUE #787: Randomly failing unit test: TestFormPkgViews
Unit Test: `TestFormPkgViews/ensure_multiple_charts_under_same_component_are_handled` is randomly failing with error:  
```
panic: interface conversion: interface {} is nil, not map[string]interface {}
```

Doesn't happen all of the time and most of the time rerunning the test in ci passes.

TristanHoladay:
## Description
Fixes an intermittent unit test failure in deploy_test.go with the "ensure multiple charts under one component" case in the `TestPkgViews` tests. The failure was occurring because we expected the second chart to always be in index `1` of the overrides map but because Go doesn't loop through maps in a guaranteed order the second chart values are some times in the `0` index.

## Related Issue

Fixes #787 

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/bundle/deploy_test.go
+++ b/src/pkg/bundle/deploy_test.go
@@ -351,6 +351,9 @@ func TestFileVariableHandlers(t *testing.T) {
 }
 
 func TestFormPkgViews(t *testing.T) {
+	type anyArr = []interface{}
+	type viewOverVars = map[string]map[string]interface{}
+
 	const (
 		componentName = "test-component"
 		chartName     = "test-chart"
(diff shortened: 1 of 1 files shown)


UncleGedd:
> The failure was occurring because we expected the second chart to always be in index 1 of the overrides map but because Go doesn't loop through maps in a guaranteed order the second chart values are some times in the 0 index.

Heck yes, great find!

src/pkg/bundle/deploy_test.go
```diff
@@ -351,6 +351,9 @@ func TestFileVariableHandlers(t *testing.T) {
 }
 
 func TestFormPkgViews(t *testing.T) {
+	type anyArr = []interface{}
```
UncleGedd, line 354: a comment here would be good so we don't forget why these exist

UncleGedd (approved):

PR #800, opened 2024-07-16
feat: nightly releases

UncleGedd:
## Description

Sets up nightly releases

### DIFF
--- a/.github/workflows/nightly-release.yaml
+++ b/.github/workflows/nightly-release.yaml
@@ -1,12 +1,11 @@
-name: Release UDS-CLI on Tag
+name: Nightly Release
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *' ## Every day at midnight UTC
 
 permissions:
   contents: read
-
-on:
-  push:
-    tags:
-      - "v*"
 
 jobs:
   checkout:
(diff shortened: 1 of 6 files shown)


.github/workflows/nightly-release.yaml
```diff
@@ -0,0 +1,76 @@
...
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *' ## Every day at midnight UTC
+
+permissions:
+  contents: read
+
+jobs:
+  checkout:
```
TristanHoladay, line 11: is this checkout needed as a separate job instead of just including as the first step in the test job? noticed this in `release.yaml` as well.

.github/workflows/nightly-release.yaml
```diff
@@ -0,0 +1,76 @@
...
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          script: |
+            github.rest.git.createRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: 'refs/tags/nightly',
+              sha: context.sha
+            }).catch(err => {
+              if (err.status !== 422) throw err;
```
TristanHoladay, line 60: maybe a comment here that 422 is not processable, i'm guessing because of the ref .... and therefore doing the update in that case?

UncleGedd: Exactly, if the ref already exists it will fail, so we need to call update, I'll add a comment

.github/workflows/nightly-release.yaml
UncleGedd: I used `parallel-tests.yaml` as the example here. I believe it is necessary because in order to call the test job, you have to need to be able to reference it from a local repo. Then the test job needs the checkout because its in a new container <- my understanding of GH Actions works

UncleGedd: good call, updated 🚀

decleaver (approved):
Lets see it in action!

Issue #782, opened 2024-07-12
Refactor Testing Commands


UncleGedd:
### Describe what should be investigated or refactored

Today in our e2e tests we are using dedicated functions that wrap UDS commands (ie. `deploy()`, `devDeploy()`, `devDeployPackages()`, etc).

In a separate PR the `runCmd` function was introduced which is a generic way to pass in UDS Commands. Discuss refactoring our test suite to remove the individual functions in favor of `runCmd`

TristanHoladay:
I like `runCmd` because you can very clearly see the details of the cli command that will execute, but after converting a few tests i'm wondering if that quick transparency is worth the redundant code and clutter that i think it brings.

Maybe it's better for those one off cases that the generic commands don't allow for?

![Screenshot from 2024-07-16 08-08-03](https://github.com/user-attachments/assets/d551f880-3f3d-4e66-98a4-51ae36740352)

UncleGedd:
My 2 cents:
- Having `runCmd` saves me from having to click into those dedicated testing functions
- Agreed `runCmd` is sort of redundant, but so is creating the same `create` command with multiple flags

Thoughts @decleaver ?

PR #823, opened 2024-07-18
chore: update tests to use runCmd()

TristanHoladay:
## Description
Switch tests from the commands in commands_test.go to runCmd() for clarity of what is actually happening in each cmd.

## Related Issue

Relates to #782 

## Type of change

## Checklist before merging


src/test/e2e/bundle_test.go
```diff
@@ -21,17 +21,11 @@ import (
 	"github.com/defenseunicorns/uds-cli/src/config"
 )
 
-func TestUDSCmd(t *testing.T) {
```
UncleGedd, line 24: was this test moved somewhere?

TristanHoladay: no i just removed it cause i thought it wasn't useful. i see now though it was calling `uds` which should return the help so probably a good check to have after all.

UncleGedd (approved):

PR #829, opened 2024-07-19
fix: split override envs properly

mjnagel:
## Description

Splits envs into 2 sections explicitly, handling situations where envs have `=` in them.

Also updated the base64 secret test to include `=` since this is a common place where this could occur (base64 padding).

## Related Issue

No issue opened, can be if required.

## Type of change

## Checklist before merging


UncleGedd (approved):
holy geez that's a good find, really appreicate this 🚀

Issue #839, opened 2024-07-22
--registry flag


bm54cloud:
### Is your feature request related to a problem? Please describe

The HNCD team houses their Zarf registry external to the cluster on a dedicated server. We let the cluster know about the external Zarf registry by using the following command:
zarf init --registry-url <external_registry_URL> --registry-push-username zarf-push --registry-push-password $ZARF_REGISTRY_PASSWORD --registry-pull-username zarf-pull --registry-pull-password $ZARF_REGISTRY_PASSWORD --components git-server --confirm --no-progress

The --registry flag does not currently exist in the uds-cli, but we need it and the following flags incorporated so that once we switch from DUBBD to UDS-core, we are still able to notify the cluster of our external Zard registry:

--registry-url
--registry-push-username
--registry-push-password
--registry-pull-username
--registry-pull-password
--registry-secret

### Describe the solution you'd like

- Given that not having the --registry flags available as part of the uds-cli prevents us from deploying an external Zarf registry (which is required for our mission hero), we request that 6 flags listed above be incorporated into uds-cli. 

### Describe alternatives you've considered

Considered not using an external Zarf registry, however this has rendered our cluster to have availability errors in the HNCD air gapped environment in the past, which is why we chose to use an external registry. Thank you!

PR #842, opened 2024-07-23
feat: handle external registry with zarf init opts

UncleGedd:
## Description

enables users to configure `zarf init` options

## Related Issue

Fixes https://github.com/defenseunicorns/uds-cli/issues/839


decleaver (approved):
Discussed IRL, nodeport was added

PR #843, opened 2024-07-23
chore: bundle signing

### LINKED ISSUE #831: Bundle Signing

Bundle signing has been around since the start of UDS CLI but I'm reasonably sure no one is using it. In an effort to not introduce breaking changes, let's instead bolster this feature by making sure it works by adding tests, docs, etc

decleaver:
## Description

Fixes functionality for signing bundles, and verifying, inspecting and deploying signed bundles

## Related Issue

Fixes #831 

## Type of change

## Checklist before merging


UncleGedd:
Lingering questions because I know nothing about cosign 😅 
- does this work in the airgap?
- need to provide on docs on usage and maybe a blurb on "why"
- Is the signing artifact that gets put in the bundle a secret?

src/cmd/viper.go
```diff
@@ -38,6 +38,9 @@ const (
 	// Bundle pull config keys
 	V_BNDL_PULL_OUTPUT = "bundle.pull.output"
 	V_BNDL_PULL_KEY    = "bundle.pull.key"
+
+	// Bundle deploy config keys
+	V_BNDL_DEPLOY_KEY = "bundle.deploy.key"
```
UncleGedd, line 43: What do you think of `bundle.deployKey`? Just thinking of ways to reduce YAML layers

decleaver:
@UncleGedd so what is currently in the PR covers the actual signing of the bundle.yaml similar to what zarf is doing. The difference is that zarf keeps an aggregatechecksum of everything in the zarf.yaml metadata, which we currently don't. We will need to add this to make the signing worth while.

That being said, zarf is looking into updating the way that they do signing based on: 
https://github.com/zarf-dev/zarf/issues/2805

https://github.com/sigstore/cosign?tab=readme-ov-file#verify-a-container-in-an-air-gapped-environment

So not sure where this falls in terms of priority

catsby:
Hello 👋 I'm going to close this for now as it's quite out of date. If we don't have this feature yet, and decide to add it, we may revisit this PR and pull any relevant commits if possible.

PR #846, opened 2024-07-29
chore: update zarf

schristoff:
Hi! Zarf has released v0.37.0 which not only changes the name of repository organization (`defenseunicorns/zarf` to `zarf-dev/zarf`) but also we had some functions we moved around that I attempted to fix for y'all.

Feedback welcome to get this PR merged in with littlest effort from y'all. Thank you for your patience :)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/defenseunicorns/uds-cli
 
-go 1.22.4
+go 1.22.5
```
schristoff, line 3: Call out that I also updated your go vers here

src/cmd/common.go
```diff
@@ -131,5 +131,10 @@ func cliSetup(cmd *cobra.Command) error {
 			return fmt.Errorf(err.Error())
 		}
 	}
+
+	// TODO(schristoff): Leverage SDK and refactor
+	// This sets up zarf CLI with the same configuration options as UDSCLI
+	zarfCommon.SetupCLI(logLevel, config.SkipLogFile, config.NoColor)
```
schristoff, line 137: We moved `config.NoColor` to be a var and not put of the config, so this is the "way to pass it in", happy to move this function somewhere better suited if there is a place

UncleGedd:
Thanks again! Merged changes in https://github.com/defenseunicorns/uds-cli/pull/851

Issue #614, opened 2024-05-15
engine: UDS state and intelligent removes


UncleGedd:
### Is your feature request related to a problem? Please describe

Intelligent package handling on bundle upgrades

### Describe the solution you'd like

- **Given** a bundle has been deployed successfully once, and now I'm upgrading
- **When** I upgrade and a package has been removed from the `uds-bundle.yaml`
- **Then** I want the option to remove the removed package from my cluster

### Additional context

- Creating this issue based on a Slack thread with @mjnagel and @YrrepNoj 
- State was implemented in https://github.com/defenseunicorns/uds-cli/pull/868

mjnagel:
I suspect the first part of this is likely keeping track of bundles in a secret (similar to how zarf stores package state in a secret)? If the "bundle state secret" keeps track of zarf packages (at a minimum) then on an "upgrade" we would know which packages were removed/added. I think there's probably another layer that zarf might need to handle (component removals) but handling bundle state is a good first thing to tackle.

MxNxPx:
@UncleGedd we definitely have hit this problem with `uds-infrastructure` and would benefit from solving.
we needed to remove a zarf package that was defined in a bundle (example: a LFAI model gets swapped out), and ended up manually connecting to the cluster and doing a `zarf package remove`.

UncleGedd:
@decleaver is working on a design doc for this feature over in https://github.com/defenseunicorns/uds-cli/pull/834

UncleGedd:
Need to figure out vision for UDS state with @jeff-mccoy

UncleGedd:
Spoke with Jeff and agreed to implement a minimal version of UDS state

catsby:
Update: we're evaluating options here, so I'm going to close this issue at this time.

PR #868, opened 2024-08-20
feat: adds uds state

UncleGedd:
## Description

adds UDS state

## Related Issue
https://github.com/defenseunicorns/uds-cli/issues/614

## Type of change

## Checklist before merging


src/pkg/state/state.go
```diff
@@ -0,0 +1,279 @@
...
+				},
+				Data: map[string][]byte{
+					"data": {},
+				},
+			}
+			_, err = c.client.CoreV1().Secrets(stateNs).Create(context.TODO(), stateSecret, metav1.CreateOptions{})
+			if err != nil {
+				return nil, err
+			}
+		}
```
decleaver, line 121: Is it possible to get another error here that is not a "IsNotFound"?

src/pkg/bundle/deploy.go
```diff
@@ -56,23 +58,55 @@ func (b *Bundle) Deploy() error {
...
+	}
+
+	err = sc.InitBundleState(b.bundle)
+	if err != nil {
+		return err
+	}
+
+	// update bundle state with deploying
+	err = sc.UpdateBundleState(b.bundle.Metadata.Name, state.Deploying)
+
```
decleaver, line 78:
```suggestion
if err != nil {
		return err
	}
```

src/pkg/bundle/remove.go
```diff
@@ -106,8 +152,21 @@ func removePackages(packagesToRemove []types.Package, b *Bundle) error {
 			if err := pkgClient.Remove(context.TODO()); err != nil {
 				return err
 			}
+
+			err = sc.UpdateBundlePkgState(b.bundle.Metadata.Name, pkg.Name, state.Removed)
```
decleaver, line 156:
```suggestion
			err = sc.UpdateBundlePkgState(b.bundle.Metadata.Name, pkg.Name, state.Removed)
			if err != nil {
				return err
			}
```

src/pkg/bundle/remove.go
```diff
@@ -106,8 +152,21 @@ func removePackages(packagesToRemove []types.Package, b *Bundle) error {
 			if err := pkgClient.Remove(context.TODO()); err != nil {
 				return err
```
decleaver, line 153: if we run into an error while removing, is it ok for the state to remain state.Removing?

src/pkg/bundle/deploy.go
UncleGedd: Good catch, updated!

src/pkg/state/state.go
UncleGedd: yes updated

src/pkg/bundle/remove.go
UncleGedd: We could add another state called `FailedRemove`

UncleGedd: 👍 added

src/pkg/bundle/remove.go
UncleGedd: fixed!

decleaver (approved):
tested out locally, looks good!

PR #885, opened 2024-08-29
fix: updates nightly uds core smoke test

UncleGedd:
## Description

Ensures we are always testing against the `latest` version of UDS Core and runs better validations


decleaver (approved):
👍

catsby (approved):
🚀 pending CI

PR #887, opened 2024-08-29
chore: updates uds state design doc

UncleGedd:
## Description

Before we too much further on the implementation of UDS state, let's take a step back and formally decide on some implementation details

### DIFF
--- a/docs/design/bundle-state.md
+++ b/docs/design/bundle-state.md
@@ -1,68 +1,92 @@
-# Bundle state
+# Bundle State
 
 ## Context
-As a user of the UDS-CLI, I want to be able to see the state of my bundles in the cluster so that I can understand what is currently deployed. This information is useful to know before applying upgrades, troubleshooting, removing bundles, etc.
 
-Currently Zarf provides a `package list` command that lists all the packages in the cluster and users would like similar functionality for bundles.
+The following 2 issues provide context driving the need for a UDS state tracking mechanism:
 
-## Options
+#### 1. What's in my cluster?
 
-### Use Secrets
-Save metadata about a package that has been deployed to the cluster in a secret. Saved data includes name, package kind, metadata, build data, components, constants, variables, and installed charts among other data. (Could also use a configmap)
+UDS CLI users have requested the ability to see the state of their bundles that are currently deployed in K8s cluster. This information is useful to know before applying upgrades, troubleshooting, removing bundles, etc. Currently, Zarf provides a `package list` command that lists all the packages in the cluster and users would like similar functionality for bundles.
 
-pros:
-- already have template of implementation in zarf
-- (main reason zarf took this approach) this is how helm does it, it uses secrets as a default storage driver
+#### 2. Orphaned Zarf Packages
+Today it's possible for packages that have been removed from a `uds-bundle.yaml` but previously deployed in the cluster to become orphaned. For example, an engineer has created a UDS bundle consisting of 3 packages and deployed this bundle to a mission environment. Later, the engineer decides one of those packages is no longer needed, so they remove it from the bundle and deploy the bundle again. The package that was removed from the bundle is still present in the cluster and is now orphaned, and the engineer must manually remove it.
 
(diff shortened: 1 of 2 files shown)


docs/design/bundle-state.md
```diff
@@ -1,68 +1,92 @@
...
-We create a Bundle custom resource that stores all the bundle information.
-https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/
+#### Design Principles
 
-Pros:
-- we can create a custom resource that is specifically designed to store bundle information, that we can easily push to and retrieve from the cluster.
+- Keep state as simple as possible. Meaning that we should think of state as a record of an event, as opposed to a complex object that drives CLI behavior.
+- No destructive action should be taken based on UDS state unless the user explicitly requests it.
+- State should be backwards compatible and should not interfere with existing UDS CLI functionality.
+  - On backards compatibility: if a user attempts an action that is based on state but state does not exist, CLI should fail quickly and indicate to the user that state does not exist and provide instructions on how to create it (likely simply re-deploying the bundle)
```
catsby, line 23:
```suggestion
  - On backwards compatibility: if a user attempts an action that is based on state but state does not exist, CLI should fail quickly and indicate to the user that state does not exist and provide instructions on how to create it (likely simply re-deploying the bundle)
```

catsby (dismissed):
1 small typo otherwise LGTM

docs/design/bundle-state.md
```diff
@@ -1,68 +1,123 @@
...
+- If using the `--packages` flag, the CLI will only update the state for the specified packages, but the bundle secret will indicate that a package in the bundle hasn't been deployed, or in the case of removal, that a package has been removed.
+
+
+### API Design
+UDS CLI will provide a bundle state API in the form of a Go pkg called `github.com/defenseunicorns/uds-cli/src/pkg/state`. This package will provide an API for creating and interacting with state during the bundle lifecycle. Proposed public methods include:
+
+- `NewClient`: creates a new state client
+- `InitBundleState`: creates the `uds` namespace if it doesn't exist; creates a new state secret or returns an existing state secret for a previously deployed bundle
+- `GetBundleState`: retrieves the state for a given bundle
+- `GetBundlePkg`: retrieves the state for a given package inside a bundle
```
TristanHoladay, line 87: nit: make this name conform to the others and end with `State`

TristanHoladay: this is a really great and clear design doc.

docs/design/bundle-state.md
```diff
@@ -1,68 +1,123 @@
...
-Cons:
-- we would be overloading the package metadata with bundle information, which could be confusing.
-- we compromise the integrity of the package by modifying it.
+The `BundleState` struct will be stored in the secret's `data` field as a base64 encoded JSON string.
 
-## Decision
-Use secrets to store metadata about a bundle that has been deployed to the cluster
+#### Namespace
+
+The UDS state secret will be stored in the `uds` namespace. If the `uds` namespace doesn't exist, the CLI will create it.
```
noahpb, line 67: Is there a security risk posed by using a shared namespace for all bundle state secrets? Also curious if this could lead to conflicting state secrets if the same bundle were deployed in more than one namespace.

UncleGedd: > Is there a security risk

IMO no, the state secrets don't contain any sensitive data

> conflicting state secrets if the same bundle were deployed in more than one namespace

Good point that we hadn't considered. I don't _think_ this will be an issue because bundles themselves aren't namespaced, thoughts?

noahpb: Fair point, I didn't consider that bundles aren't namespace scoped. Shouldn't be an issue then I don't think.

decleaver (approved):

docs/design/bundle-state.md
UncleGedd: updated!

decleaver (approved):
👍

PR #906, opened 2024-09-05
fix: cleanup color and typo

### LINKED ISSUE #897: No styling for Bundle Kind in pre-deploy view and misspelled word
### Environment

Screenshot 2024-09-04 at 12 01 53 PM

### Steps to reproduce

1. Deploy any bundle without using `--confirm`

### Expected result

- Add `kind: UDS Bundle` to Metadata and make sure it's styled correctly (adding to `Metadata` isn't a hard req, but the styling is)
- Correct spelling of `Metadata`

catsby:
## Description

~Cleans up the positioning of `"kind:"` to be under Metadata, fixes the color output, and fixes a typo~. UPDATE: removes `"kind:"` entirely and fixes typo.

Unfortunately I couldn't merge this in with the existing `b.bundle.Metadata` struct. I tried a new inline struct that embedded the `Metadata` one but the yaml printer ended up making a new section for it, which isn't what I wanted. 

New look:

Screenshot 2024-09-06 at 10 37 51 AM

Also unfortunately, `ColorPrintYAML()` forces an empty line before printing, so that's why there's a gap there.

While I was here, I renamed the imported zarf utils package to `zarfUtils` because I noticed `uds-cli` also has a `utils` package. While we don't use the `uds-cli` one here, while experimenting here with `pterm` I found that we did and there was confusion as to which I was trying to use... naming the import `zarfUtils` is consistent with what we do in [pkg/bundle/create.go](https://github.com/defenseunicorns/uds-cli/blob/95747525dab27403a622b766fe4dd492cb013442/src/pkg/bundle/create.go#L20) so I went with that. 

## Related Issue

Fixes #897

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/bundle/deploy.go
+++ b/src/pkg/bundle/deploy.go
@@ -358,8 +358,7 @@ func (b *Bundle) ConfirmBundleDeploy() (confirm bool) {
 
 	message.HeaderInfof("🎁 BUNDLE DEFINITION")
 
-	message.Title("Metatdata:", "information about this bundle")
-	zarfUtils.ColorPrintYAML(map[string]string{"kind": b.bundle.Kind}, nil, false)
+	message.Title("Metadata:", "information about this bundle")
 	zarfUtils.ColorPrintYAML(b.bundle.Metadata, nil, false)
 
 	message.HorizontalRule()
--- a/src/pkg/bundle/deploy.go
+++ b/src/pkg/bundle/deploy.go
@@ -21,7 +21,6 @@ import (
 	"github.com/defenseunicorns/uds-cli/src/types/chartvariable"
 	"github.com/defenseunicorns/uds-cli/src/types/valuesources"
 	goyaml "github.com/goccy/go-yaml"
-	"github.com/pterm/pterm"
 	"github.com/zarf-dev/zarf/src/api/v1alpha1"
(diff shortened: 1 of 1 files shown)


UncleGedd:
That newline! I don't love that either. What do you think of removing `kind: UDSBundle` from that output entirely? I think it's fairly self-evident with the `BUNDLE DEFINITION` header

src/pkg/bundle/deploy.go
```diff
@@ -358,11 +357,9 @@ func (b *Bundle) ConfirmBundleDeploy() (confirm bool) {
 	pkgviews := formPkgViews(b)
 
 	message.HeaderInfof("🎁 BUNDLE DEFINITION")
-	pterm.Println("kind: UDS Bundle")
-
-	message.HorizontalRule()
 
 	message.Title("Metatdata:", "information about this bundle")
```
UncleGedd, line 361: Still seeing `Metatdata` vs `Metadata`, but your screenshot is correct?

catsby: Whoops, I think that got restored when I rebased earlier. I fixed it (again) and removed the `kind:` part in 0a57611

catsby: I also updated the screenshot included to reflect the latest

UncleGedd (approved):
🚀

PR #909, opened 2024-09-06
feat: adds feature flag to state

UncleGedd:
## Description

It was discovered during our release pipeline that the new state feature contains a bug where state could be introduced in the wrong cluster. This PR wraps all state-related code in feature flag so we can release


decleaver (approved):
👍

PR #910, opened 2024-09-06
fix: remove uds core nightly state checks for now

UncleGedd:
## Description

remove uds core nightly state checks for now


decleaver (approved):
apology accepted

PR #914, opened 2024-09-09
feat: Validate and format bundle names to exclude special characters

### LINKED ISSUE #886: Sanitize Bundle Name
Based on a comment from @catsby (https://github.com/defenseunicorns/uds-cli/pull/882#discussion_r1735361049) we should be doing some kind of sanitizing on bundle names. 

- [ ] ensure bundle name is sanitized
- [ ] look into if there are other pieces of metadata getting used in potentially compromising ways and should be sanitized

catsby:
## Description

Prevent special characters in bundle names, which ultimately affect the tarball filename but are in general probably not a good idea to allow anyway. 

In this PR if a bundle's name contains non letters, numbers, or hyphens, uds-cli will error. If the name is valid then we remove all leading and trailing spaces (`" "`), lowercase everything, and replace any remaining spaces with hyphens. 

Ex:

`worDpress` -> `wordpress`
`" Word pRess       "` -> `word-press`

It's currently possible to use a bundle name with odd characters or even spaces which result in tarball artifacts with either the special characters or spaces in the name. 

Example:

The `name` here is unlikely but still valid

```yaml
kind: UDSBundle
metadata:
  name: "a exAmple   ??}doOm\n"
  description: an example UDS bundle
  version: 0.0.1
```

Output (trimmed):

```console
 ➜ uds create
  🎁 BUNDLE DEFINITION
	
kind: UDSBundle
metadata:
  name: |
    a exAmple   ??}doOm
  description: an example UDS bundle
  version: 0.0.1
packages:
- name: dos-game
  repository: ghcr.io/zarf-dev/packages/dos-games
  ref: 1.1.0
[...]
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
[...]
? Create this bundle? Yes
  ✔  Bundle Validated
[...]
  🚧 Building Bundle
[...]
  ✔  Created bundle archive at: /Users/clint/go/github.com/defenseunicorns/uds-cli/uds-bundle-a exAmple   ??}doOm
       -arm64-0.0.1.tar.zst
```

Generates this file:

```
-rw-r--r--@  1 clint  staff  3547344 Sep  9 11:39 uds-bundle-a exAmple   ??}doOm?-arm64-0.0.1.tar.zst
```

## Note:

This is currently a draft for review and feedback on next steps. It's not entirely clear if we should outright error and abort initially, or if we should at first output some kind of warning that the name is not valid and in the future we'll error, to enable a sort of grace period. 

## Related Issue

Fixes #886 

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/bundle/create.go
+++ b/src/pkg/bundle/create.go
@@ -41,6 +41,13 @@ func (b *Bundle) Create() error {
 		return err
 	}
 
+	// validate bundle name to avoid bad characters
+	name, err := utils.FormatBundleName(b.bundle.Metadata.Name)
+	if err != nil {
+		return err
+	}
+	b.bundle.Metadata.Name = name
+
 	// confirm creation
 	if ok := b.confirmBundleCreation(); !ok {
 		return fmt.Errorf("bundle creation cancelled")
--- a/src/pkg/utils/utils.go
+++ b/src/pkg/utils/utils.go
@@ -7,6 +7,7 @@ package utils
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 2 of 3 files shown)


decleaver:
Did someone external create an issue for this or was it something you found? I would say we just error out and abort if someone tries to use a bad name

catsby:
@decleaver it came from me. I was working through onboarding things and for reasons I don't recall I had a space in my bundle name (`"example doom game"` or something) and noticed the resulting tarball had spaces in the file name that had to be escaped if I wanted to reference it

UncleGedd:
+1 to aborting and erroring out, but when we do, show the user how they can fix the error

UncleGedd:
"I want to not break things and not be terribly surprising" - @catsby

Issue #870, opened 2024-08-21
Add `uds ui` command


decleaver:

catsby:
#917 added this, closing

PR #917, opened 2024-09-09
feat: add uds ui cmd

decleaver:
## Description
Adds `uds ui` which launches uds runtime

## Related Issue

Relates to #870 

## Type of change

## Checklist before merging


src/cmd/uds.go
```diff
@@ -18,6 +22,9 @@ import (
 	"github.com/spf13/cobra"
 )
 
+//go:embed bin/uds-runtime-*
```
UncleGedd, line 25: Can we remove the asterisk and refer to the specific binary?

UncleGedd:
Can we get around having ot add the binaries to the repo? Also we shouldn't need to bundle all versions of Runtime into the binary. For example the darwin-arm64 CLI should embed only the darwin-arm64 Runtime. Thoughts?

design-docs/template.md
```diff
@@ -0,0 +1,48 @@
+# Design Doc Title
```
UncleGedd, line 1: thanks for including this!

decleaver:
> Can we get around having to add the binaries to the repo? Also we shouldn't need to bundle all versions of Runtime into the binary. For example the darwin-arm64 CLI should embed only the darwin-arm64 Runtime. Thoughts?

Good point, I can update it to only bundle the matching runtime binary. As far as adding the binaries to the repo, have some thoughts... pros/cons to different approaches.

src/cmd/uds.go
decleaver: yup, will update so we only bundle necessary binary at build time

docs/command-reference/uds_ui.md
```diff
@@ -0,0 +1,37 @@
+---
+title: uds ui
+description: UDS CLI command reference for <code>uds ui</code>.
+type: docs
+---
+## uds ui
+
+[beta] Deploy UDS Runtime and view UI
```
catsby, line 8: It's my understanding we are not "deploying" Runtime here, so if nothing else we should update the wordage. 

```suggestion
[beta] Launch UDS Runtime and view UI
```

decleaver: good catch, these docs are auto generated based on the code, but i'll update the working in lang.go

design-docs/0002-uds-ui.md
```diff
@@ -0,0 +1,30 @@
...
+
+The goal of the `uds ui` command is to allow uds-cli users to launch the UDS Runtime application from the command line.
+
+### Proposal
+
+Bundle the UDS Runtime binaries as a part of uds-cli allowing users to launch UDS Runtime from the command line.
+
+### Scope and Requirements
+
+Allow users to launch UDS Runtime from uds-cli
```
UncleGedd, line 18: could you mention token-based auth as a requirement?

docs/quickstart-and-usage.md
```diff
@@ -509,3 +509,7 @@ uds scan --org <organization> --package-name <package-name> --tag <tag> [options
 ```sh
 uds scan -o defenseunicorns -n packages/uds/gitlab-runner -g 16.10.0-uds.0-upstream -u docker-username -p docker-password -f gitlab-runner.csv
 ```
+
+## UDS Runtime
+
+The `uds ui` command launches UDS Runtime, which gives you a user interface to view what is running in your cluster. More information regarding UDS Runtime can be found [here](https://github.com/defenseunicorns/uds-runtime).
```
UncleGedd, line 515:
```suggestion
The `uds ui` command launches UDS Runtime, which provides a web-based user interface to view what is running in your K8s cluster. More information regarding UDS Runtime can be found [here](https://github.com/defenseunicorns/uds-runtime).
```

src/cmd/uds.go
```diff
@@ -222,6 +228,74 @@ var logsCmd = &cobra.Command{
 	},
 }
 
+var uiCmd = &cobra.Command{
+	Use:     "ui",
+	Aliases: []string{"u"},
+	Short:   lang.CmdUIShort,
+	RunE: func(_ *cobra.Command, _ []string) error {
+
+		// Create a temporary directory to hold the embedded files
```
UncleGedd, line 237: what do you think of moving this code into its own `ui.go` or even in `vendored.go`?

decleaver: Yeah, I considered adding it to `vendored.go` but didn't think it really fit with the way we are vendoring the other tools. I'll go ahead and add it into its own `ui.go`

design-docs/0002-uds-ui.md
```diff
@@ -0,0 +1,30 @@
...
+
+### Implementation Details
+
+To execute UDS Runtime from the `uds-cli`, the appropriate runtime binary is pulled into the `src/cmd/bin` directory during the build process. This is based on the specific build task being run (e.g., `build-cli-linux-arm`). The runtime binary is then packaged with the `uds-cli` binary.
+
+To ensure the `uds ui` command works correctly, the `uds-cli` must be built using the build tasks specified in `tasks.yaml`. If not, the runtime binary might be missing, causing the `uds ui` command to fail.
+
+### Alternatives Considered
```
UncleGedd, line 26: I'd like to include another considered alternative detailing why vendoring doesn't work in this situation. Happy to write it myself and push to this branch

UncleGedd:
Great work! I explored vendoring for a bit before realizing that was a dead end. One con to the embedding binary approach is that we're creating tmp dirs with the runtime binary on the user's system, and those tmp dirs can potentially be abandoned if the process is killed and the `defer` function doesn't get called (like when the user runs ctrl-c)

This can be a separate issue, but I'd like to see the following scenarios handled:
1. What happens when the user hits ctrl-c? Can we force a cleanup of the tmp dir?
2. Besides ctrl-c is there a graceful way to exit `uds ui`?

decleaver:
> * What happens when the user hits ctrl-c? Can we force a cleanup of the tmp dir?
> * Besides ctrl-c is there a graceful way to exit `uds ui`?

Created follow on issue: https://github.com/defenseunicorns/uds-cli/issues/922

src/cmd/ui.go
```diff
@@ -0,0 +1,93 @@
...
+		tmpDir, err := os.MkdirTemp("", "uds-runtime-*")
+		if err != nil {
+			return fmt.Errorf("failed to create temp directory: %v", err)
+		}
+		defer os.RemoveAll(tmpDir)
+
+		var runtimeBinaryPath string
+		// Walk through the embedded files and write them to the temporary directory, eventhough we only expect one file
+		// to be embedded, the name of the binary is based on the architecture and os
+		err = fs.WalkDir(embeddedFiles, "bin", func(path string, d fs.DirEntry, err error) error {
```
UncleGedd, line 38: instead of walking through the dir, can we get the arch and OS via `runtime.GOOS` and `runtime.GOARCH`?

decleaver: ooh, yeah!

design-docs/0002-uds-ui.md
decleaver: I can add some words, but i didn't try going down that route, so feel free to additional info on what you found.

UncleGedd (approved):

TristanHoladay (approved):

PR #919, opened 2024-09-10
fix: remove extra Zarf logging output

UncleGedd:
## Description

Remove extra Zarf logging output. Previously, we saw an extra note from Zarf saying where a log file was located, this PR removes that behavior by disabling `pterm` output briefly while setting up the Zarf CLI

Noting that the Zarf output pointed to a log file with no useful information because we capture all the useful output in the UDS log file


decleaver (approved):
tested locally :shipit:

Issue #923, opened 2024-09-18
Inspect SBOM Errors With 2+ Local Packages


TristanHoladay:
## UDS Version
`0.15.0`

## Context
Inspecting a bundle for sboms, that contains multiple local packages, gives a file already exists error.

```console
     ERROR:  failed to inspect bundle: failed to extract 18ec08c709b513b9970af655e6571bf0a88a5245219acb104c8be6d15434804e from bundles/deps/uds-bundle-deps-amd64-0.1.tar.zst: walking 9f3f02e684b7fe56949e48843bdf36fa7e0cc8825e0071348d5bb5c98db7cd12: extracting file blobs/sha256/9f3f02e684b7fe56949e48843bdf36fa7e0cc8825e0071348d5bb5c98db7cd12: file already exists: /tmp/zarf-177103816/blobs/sha256/9f3f02e684b7fe56949e48843bdf36fa7e0cc8825e0071348d5bb5c98db7cd12
```

## Further troubleshooting
Have not checked yet if this occurs with any 2 local packages or just the ones used by Delivery engineer that present the issue.

UncleGedd:
The issue occurs when you bundle >1 Zarf pkg containing either only manifests or only Zarf actions

PR #924, opened 2024-09-18
fix: inspecting sboms on simple pkgs

UncleGedd:
## Description

- Fixes `inspect --sbom` issue with multiple 'simple' Zarf pkgs
- Prints warnings at the end of `inspect` cmds (to stderr)

## Related Issue

https://github.com/defenseunicorns/uds-cli/issues/923

## Type of change


src/pkg/utils/sbom.go
```diff
@@ -86,3 +86,32 @@ func SBOMExtractor(dst string, SBOMArtifactPathMap map[string]string) func(_ con
 	}
 	return extractor
 }
+
+// HandleSBOM handles the extraction and creation of bundle SBOMs after populating SBOMArtifactPathMap
+func HandleSBOM(extractSBOM bool, SBOMArtifactPathMap map[string]string, bundleName, dstPath string) ([]string, error) {
```
catsby, line 91: Nit: We should probably unit test this function since it's exported. 

Is there any case where `HandleSBOM` will return more than 1 warning? Reading the code I only see 2 `append` calls for the slice, and they both end up returning immediately after, so they wouldn't both be called in a single invocation.

catsby: I'm wondering if we will ever return a warning and an error? Right now the code doesn't look like it, but maybe we anticipate things in the future?

catsby (commented):
Questions on the `HandleSBOM()` function. I'm wondering if we'll ever return more than 1 warning, or a warning *and* and error. Right now the code wont do that (to my knowledge) and if that's the case going forward maybe we should just have `HandleSBOM()` return a named error and check for that whenever we invoke it.

src/pkg/utils/sbom.go
UncleGedd: > Nit: We should probably unit test this function since it's exported.

IMO we've got this pretty well covered in our e2e suite. Unit testing would be difficult because of the reliance on the tarball lib

UncleGedd: reviewed sync, made a couple refactors

catsby (approved):

PR #926, opened 2024-09-19
fix: ensures bin cleanup on uds ui

UncleGedd:
## Description

- ensures bin cleanup on `uds ui`
- updates tests and ensures tests are run in CI


decleaver (approved):
tested sync

PR #939, opened 2024-09-20
fix: ensure runtime bins are included in releases

UncleGedd:
## Description

ensure runtime bins are included in releases


catsby (approved):
👍 pending CI

PR #944, opened 2024-09-26
chore: matrix testing pre-existing k3d clusters

### LINKED ISSUE #940: Ensure smoke tests cover both pre-existing cluster and no pre-existing cluster test cases
### Context

When we attempted to release the UDS state feature, our release pipeline caught a bug where state was being created in a cluster that existed before deploying the slim-dev bundle.

This bug should have been caught sooner by the nightly pipelines but was not. Noting that the bug was caught because our release pipeline starts up a cluster using the [k3d action](https://github.com/defenseunicorns/uds-cli/blob/main/.github/actions/k3d/action.yaml) which deploys an empty k3d cluster.

Smoke tests in our [release pipeline](https://github.com/defenseunicorns/uds-cli/blob/main/.github/workflows/release-tests.yaml) vs [nightly-uds-core](https://github.com/defenseunicorns/uds-cli/blob/main/.github/workflows/nightly-uds-core.yaml) are slightly different.

- In our release-pipeline, we use the our k3d Github action to download the k3d binary and start a k3d cluster ([ref](https://github.com/defenseunicorns/uds-cli/blob/c7b8ec8c9f4008665c1a6fba195358fb3c220fcc/.github/workflows/release-tests.yaml#L70-L71)). Because of this, 2 clusters get created in the Github runner
- In our nightly-uds-core smoke test, we install k3d but do not start a k3d cluster; in this case, only 1 cluster is created

There are 2 scenarios that should be tested in both pipelines:
1. Cluster already exists when we deploy slim-dev
2. Cluster does _not_ already exist when we deploy slim-dev

Update the smoke tests to cover the above scenarios in both nightly and release pipelines (can potentially create a resuable workflow to avoid duplicates configs)

catsby:
## Description

Cleanup some of our GitHub Actions, and run (some of?) our e2e tests in a matrix where one scenario we have k3d installed and a cluster created, and another with k3d installed and no cluster created.

**Note about our `setup-golang` action:** I noticed that we were using `setup-go` to install go `1.21.x`, however the `setup-go` action does not list the `.x` syntax as supported. Furthermore our `go.mod` file says go 1.22.4, so the setup go action was setting 1.21.x, and then immediately switching to go 1.22.4 because that's what our go.mod says. The result was a long list of errors trying to use 1.22 because it was being pulled from cache but conflicted with the 1.21 we just setup. 

See here for example: https://github.com/defenseunicorns/uds-cli/actions/runs/11037302233/job/30657907519#step:3:102

**Note about our `setup-from-previous` action:** I noticed that every invocation was preceded by a `checkout` action, and the first step of `setup-from-previous` was also a checkout. 

See [here](https://github.com/defenseunicorns/uds-cli/actions/runs/11037302233/job/30657907519#step:2:165) and [here](https://github.com/defenseunicorns/uds-cli/actions/runs/11037302233/job/30657907519#step:3:67) for example of trying to checkout the repo 2x in the same job

## Related Issue

Fixes #940

## Type of change

## Checklist before merging

### DIFF
--- a/.github/actions/setup-from-previous/action.yaml
+++ b/.github/actions/setup-from-previous/action.yaml
@@ -4,12 +4,6 @@ description: grabs artifact from a previous job and sets up the env for tests
 runs:
   using: composite
   steps:
-    # Checkout the repo and setup the tooling for this job
-    - name: Checkout
-      uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
-      with:
-        fetch-depth: 0
-
     - name: Download build artifacts
       uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
       with:
--- a/.github/workflows/nightly-uds-core.yaml
+++ b/.github/workflows/nightly-uds-core.yaml
@@ -42,14 +42,12 @@ jobs:
       - name: Make UDS-CLI executable
         run: |
           chmod +x build/uds
-      
+
       - name: Setup K3d
         uses: ./.github/actions/k3d
         with:
           create-cluster: ${{ matrix.create-cluster }}
-  
 
       - name: Run UDS Core smoke test
         run: build/uds run test:ci-uds-core-smoke-test
         shell: bash
-
--- a/.github/workflows/nightly-uds-core.yaml
+++ b/.github/workflows/nightly-uds-core.yaml
@@ -43,7 +43,7 @@ jobs:
       - name: Setup K3d
         uses: ./.github/actions/k3d
         with:
-          create-cluster: "${{ matrix.type == 'with-cluster' && 'false' || 'true'}}"
+          create-cluster: "${{ matrix.type == 'with-cluster' && 'true' || 'false'}}"
 
       - name: Run UDS Core smoke test
         run: build/uds run test:ci-uds-core-smoke-test
--- a/.github/workflows/nightly-uds-core.yaml
+++ b/.github/workflows/nightly-uds-core.yaml
@@ -18,12 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        existing-cluster: [with-cluster, without-cluster]
-        include:
-          - existing-cluster: without-cluster
-            create-cluster: 'false'
-          - existing-cluster: with-cluster
-            create-cluster: 'true'
+        type: [with-cluster, without-cluster]
     permissions:
       contents: read
     steps:
--- a/.github/workflows/nightly-uds-core.yaml
+++ b/.github/workflows/nightly-uds-core.yaml
@@ -52,3 +52,4 @@ jobs:
       - name: Run UDS Core smoke test
         run: build/uds run test:ci-uds-core-smoke-test
         shell: bash
+
--- a/.github/workflows/nightly-uds-core.yaml
+++ b/.github/workflows/nightly-uds-core.yaml
@@ -21,7 +21,9 @@ jobs:
         existing-cluster: [with-cluster, without-cluster]
         include:
           - existing-cluster: without-cluster
-            create-cluster: false
+            create-cluster: 'false'
+          - existing-cluster: with-cluster
+            create-cluster: 'true'
     permissions:
       contents: read
     steps:
--- a/.github/workflows/nightly-uds-core.yaml
+++ b/.github/workflows/nightly-uds-core.yaml
@@ -43,7 +43,7 @@ jobs:
       - name: Setup K3d
         uses: ./.github/actions/k3d
         with:
-          create-cluster: ${{ matrix.create-cluster }}
+          create-cluster: "${{ matrix.type == 'with-cluster' && 'false' || 'true'}}"
 
       - name: Run UDS Core smoke test
         run: build/uds run test:ci-uds-core-smoke-test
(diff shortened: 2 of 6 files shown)


.github/actions/setup-from-previous/action.yaml
```diff
@@ -4,12 +4,6 @@ description: grabs artifact from a previous job and sets up the env for tests
 runs:
   using: composite
   steps:
-    # Checkout the repo and setup the tooling for this job
-    - name: Checkout
-      uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
-      with:
-        fetch-depth: 0
-
```
catsby, line 12: I noticed that every invocation of `setup-from-previous` also ran the `Checkout` step, so we were doing checkout twice in every use of this action

.github/workflows/nightly-uds-core.yaml
```diff
@@ -16,6 +16,14 @@ concurrency:
...
   uds-core-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        existing-cluster: [with-cluster, without-cluster]
+        include:
+          - existing-cluster: without-cluster
+            create-cluster: 'false'
+          - existing-cluster: with-cluster
+            create-cluster: 'true'
```
catsby, line 26: Open to suggestions on naming here, but basically wanted to convey that one matrix test operates by installing k3d and creating a cluster, while the other specifically does not create the cluster.

.github/workflows/test-e2e-pr.yaml
```diff
@@ -1,4 +1,4 @@
-name: E2E Tests
+name: E2E PR Tests
```
catsby, line 1: Updated the name here because we had 2 workflow files with the same `name:` value

.github/workflows/nightly-uds-core.yaml
mjnagel: Would it be clearer if this were just existing-cluster as a true/false and then pass the inverse to the action below? I'm not a big fan of the second matrix parameter here when it's really the same as the first just different name and truthy vs descriptive string.

.github/actions/setup-from-previous/action.yaml
catsby: All **Steps** in a **Job** run in the same container, and IIRC they start in the root of the repo unless specified. I verified that all invocations of `setup-from-previous` were ran _after_ a previous step ran `checkout`, so it should be good. There's nothing codifying that though, so we could actually remove those instances of checkout and just put it in this file instead, but files like `release.yaml` and such would still need to call `checkout` themselves since they don't use `setup-from-previous`

.github/workflows/nightly-uds-core.yaml
catsby: > forgot that the matrix value gets used in the job name when viewing it, so true/false would be a bit weird and unhelpful there.

Maybe this depends on how it's named? In this case, the _label_ is being used, not the value:

with-cluster

But maybe I'm misunderstanding what you're saying here, I'm not fully versed in GitHub Actions these days

catsby: ah ok re-reading this I believe you're referring to `existing-cluster: [with-cluster, without-cluster]`, I thought you were referring to the `include:` part.

mjnagel, line 26:
```suggestion
    strategy:
      matrix:
        type: [with-cluster, without-cluster]
```

.github/workflows/nightly-uds-core.yaml
```diff
@@ -37,9 +45,10 @@ jobs:
         run: |
           chmod +x build/uds
 
-      - name: Install-k3d
-        run: "curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash"
-        shell: bash
+      - name: Setup K3d
+        uses: ./.github/actions/k3d
+        with:
+          create-cluster: ${{ matrix.create-cluster }}
```
mjnagel, line 51:
```suggestion
          create-cluster: "${{ matrix.type == 'with-cluster' && 'false' || 'true'}}"
```

.github/workflows/nightly-uds-core.yaml
mjnagel: Added suggestions that I _think_ would work? Would also need to do the same on the release test of course, but wanted to add for your thoughts.

.github/actions/setup-from-previous/action.yaml
mjnagel: > There's nothing codifying that though, so we could actually remove those instances of checkout and just put it in this file instead...

Fwiw you shouldn't even be able to call an action unless a checkout has occurred so I actually think this is guaranteed. You have to checkout the repo to be able to access a local action file, otherwise the action can't be found.

catsby: @mjnagel ah great point, thanks. I either forgot or just didn't know that definitively:

- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-pre-written-building-blocks-in-your-workflow#adding-an-action-from-the-same-repository

.github/workflows/nightly-uds-core.yaml
catsby: @mjnagel I applied your suggestions, thanks! At first run I found the logic was backwards (`with-` did not create a cluster, but `without-` did) so I swapped it in e1c460c which I _believe_ is correct now. Watching CI run..

mjnagel (approved):

UncleGedd (approved):
love the `matrix` strategy 🚀

PR #953, opened 2024-10-03
chore: update LICENSE with Defense Unicorns and year

catsby:
## Description

Update the `LICENSE` file with the right info

## Type of change

## Checklist before merging


mjnagel (changes requested):
Per guidance in slack we actually want to leave this as is since it is just an example for how to apply the license, it is not meant to be replaced with our company/year.

catsby:
@mjnagel apparently so! Thanks 👍

PR #954, opened 2024-10-03
chore: update CODEOWNERS

catsby:
## Description

Update CODEOWNERS to standardize some things

## Type of change

## Checklist before merging

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,5 @@
-* @UncleGedd @jeff-mccoy @decleaver @TristanHoladay @catsby
+* @defenseunicorns/uds-core
+
+# Additional privileged files
+/CODEOWNERS @jeff-mccoy @catsby @UncleGedd @daveworth
+/LICENSE @jeff-mccoy @austenbryan


CODEOWNERS
```diff
@@ -1 +1,5 @@
-* @UncleGedd @jeff-mccoy @decleaver @TristanHoladay @catsby
+* @defenseunicorns/uds-core
```
mjnagel, line 1: Does this include you as well currently @catsby? If not, should we add you to that one? (this will increase the emails from renovate for both sides 😅)

catsby: @mjnagel as of Friday it does indeed 👍 I created some renovate filters this morning :/

mjnagel (approved):

CODEOWNERS
mjnagel: Not sure my approval means anything since the old codeowners doesn't have me listed, but approved! And yeah might be good to note for the team when we merge this since we'll all get more emails for requested reviews.

UncleGedd (approved):

Issue #445, opened 2024-02-21
Review and Secure Curl Usage in GitHub Actions Workflow


naveensrinivasan:
In our GitHub Actions workflow defined in .github/actions/install-tools/action.yaml, we have a step that uses curl to download and execute a script from the internet:
https://github.com/defenseunicorns/uds-cli/blob/ef4b57782f6db38afbf74d735aaa294bc59ab973/.github/actions/install-tools/action.yaml#L11
This software installation method raises potential security concerns, as it directly executes a script fetched from the internet. While this approach is commonly used, it can expose our CI/CD pipeline to risks if the remote content is compromised or there is a man-in-the-middle attack.
### Potential Risks
- Remote Execution Vulnerability: If the repository or the specific commit referenced is compromised, malicious code could be executed in our pipeline environment.
- Man-in-the-Middle (MitM) Attacks: Without verifying the integrity of the script before execution, there's a risk that the content could be altered in transit.

https://about.codecov.io/security-update/

catsby:
Closed with #961

PR #961, opened 2024-10-08
fix: remove unused GitHub Action install-tools

catsby:
## Description

DRAFT: It's not clear to me how these tools are used, or where, or at all. I suspect they are vestigal from scripts/build process borrowed from our friends/ancestor Zarf.

## Related Issue

Relates to #445 

## Type of change

## Checklist before merging


mjnagel (approved):
LGTM, noting that all the installs here can be traced to specific things in zarf so likely were just copy/pasted at the start of the project.

PR #967, opened 2024-10-09
chore: refactor uds ui to use vendored runtime

TristanHoladay:
## Description
Refactors `uds ui` to vendor `uds-runtime` by downloading and embedding the [Runtime UI release artifact](https://github.com/defenseunicorns/uds-runtime/pull/418) and `uds.dev` certs. This means CLI will no longer have to download and include the Runtime binary. 

## Related Issue
https://github.com/defenseunicorns/uds-runtime/issues/417

## Type of change

## Checklist before merging


TristanHoladay:
@catsby some context / questions:

1. is it right to return an error from the cobra cmd here? i saw none of the other vendored commands were but the uds commands do.

1. Darcy had implemented handling for the interrupt signal and then cleaning up the temp files. since we don't need the temp files anymore i got rid of all that, but realized in the test that means we have to expect an Error when the interrupt occurs. Is that ok?

1. I thought it would be cleaner to have the build and cert assets under `cmd/assets` but because the `fs.Sub()` in the vendored `Setup()` is hardcoded to expect 'ui/build', it was unhappy with `assets/ui/build`. so i update the download script to place it all under `cmd/`.

src/cmd/ui.go
```diff
@@ -5,116 +5,31 @@
...
+//go:embed certs/cert.pem
+var localCert []byte
+
+//go:embed certs/key.pem
+var localKey []byte
+
+func startUI() error {
+	r, incluster, err := ui.Setup(&uiBuild)
+	if err != nil {
+		slog.Error("Failed to setup UI server", err)
```
catsby, line 26: Note from [log/slog docs](https://pkg.go.dev/log/slog): 
> Package slog provides structured logging, in which log records include a message, a severity level, and various other attributes expressed as key-value pairs.


```suggestion
		slog.Error("failed to setup UI server", "error", err)
```


nit: error messages should not be capitalized. I _think_ that's an optional golangci-list thing, so it's not a strict rule or anything.

src/cmd/ui.go
```diff
@@ -5,116 +5,31 @@
...
+		return err
+	}
+	err = ui.Serve(r, localCert, localKey, incluster)
+	if err != nil {
+		slog.Error("Failed to serve UI", err)
```
catsby, line 31:
```suggestion
		slog.Error("failed to serve UI", "error", err)
```

catsby (commented):
LGTM; I'll sync up with you about question 2 as I'm not 100% sure which test you're talking about, it's not clear to me from these changes here. Question 1 and 3 seem fine to me.

catsby (approved):

rjferguson21 (approved):

PR #982, opened 2024-10-18
fix: revert "fix: remove unused GitHub Action install-tools"

catsby:
Reverts defenseunicorns/uds-cli#961

It turns out these tools are used in the release process, albeit in a sort of hidden way (to me at least), by goreleaser.

### DIFF
--- a/.github/actions/install-tools/action.yaml
+++ b/.github/actions/install-tools/action.yaml
@@ -1,0 +1,14 @@
+name: install-tools
+description: "Install pipeline tools"
+
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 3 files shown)


koesbong (approved):

.github/actions/install-tools/action.yaml
```diff
@@ -0,0 +1,17 @@
+# Copyright 2024 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+name: install-tools
+description: "Install pipeline tools"
```
mjnagel, line 5: For the future might be good to add a comment or modify this description to note that they are used by goreleaser so that we don't nuke them again accidentally down the line.

PR #990, opened 2024-10-23
chore: batch renovate actions

catsby:
## Description

An attempt to batch/group our Renovate updates, so that all GitHub Actions updates are bundled together to reduce PR noise

## Type of change

## Checklist before merging

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -28,6 +28,10 @@
         "podinfo"
       ],
       "groupName": "podinfo"
+    },
+    {
+      "fileMatch": ["action.yaml"],
+      "groupName": "actions"
     }
   ],
   "regexManagers": [


renovate.json
```diff
@@ -28,11 +28,15 @@
...
+    },
+    {
+      "fileMatch": ["action.yaml"],
+      "groupName": "actions"
     }
   ],
   "regexManagers": [
     {
-      "fileMatch": ["uds-bundle.yaml", "action.yaml"],
+      "fileMatch": ["uds-bundle.yaml"],
```
mjnagel, line 39: Might be missing something else here but by removing from this list it will no longer be picked up by renovate as part of this regex match. I think you might want to leave this, but also add the packageRule/group?

catsby: I had thought that instruction would move them and I wouldn't want that there, but perhaps you're right and what I added was just grouping. I'll change that.

mjnagel: Yeah so `packageRules` is used for grouping (and some other purposes, but we mainly use it for grouping). `regexManagers` is used to target specific strings to identify dependencies to update. So in this case we want to target updates to `action.yaml` files but now want them grouped in an `actions` group, so I think we'll want both.

catsby: probably better now?

catsby: wait nvm I overwrote my changes with a rebase, just a sec

mjnagel (approved):

PR #1041, opened 2024-12-16
fix: update zarf to 0.45.0

catsby:
## Description

Update our included Zarf to [v0.45.0](https://github.com/zarf-dev/zarf/releases/tag/v0.45.0) to address a possible issue in the included `x/crypto` library. 

## Related Issue

Supersedes https://github.com/defenseunicorns/uds-cli/pull/1037

## Type of change

## Checklist before merging


src/test/packages/gitrepo/zarf.yaml
```diff
@@ -7,10 +7,10 @@ metadata:
   version: 0.0.1
 
 components:
-  - name: dubbd-repo
+  - name: mattermost
     required: true
     repos:
-      - https://github.com/defenseunicorns/uds-package-dubbd
+      - https://github.com/defenseunicorns/uds-package-mattermost@v10.2.0-uds.1
```
mjnagel, line 13: I might just be missing something but I'm a bit suspicious of this change given the test file name. Zarf has a feature to push a gitrepo to a git server "in the airgap" (similar to how it pushes images). I think this test was meant to validate that functionality, and I'm not sure `uds-package-mattermost` uses that?

At the same time, UDS is pretty opinionated to not ever use the gitrepo feature so there might be a route for dropping this test (although I'd be a bit wary of doing that).

catsby: I believe the intent of this test was/is "can we create a bundle that uses a `zarf.yaml` which contains a reference to another package via `https://`. I don't think this test is meant to verify the behavior you describe (there are other tests around pushing).

The reason for this change was Zarf was bailing trying to find the `dubbd` package. I believe it's archived and maybe that's a problem?

mjnagel: Yep - totally misread this the first time around, this pulls the repo NOT the zarf package.

mjnagel (approved):

PR #1045, opened 2024-12-18
chore: enable perfsprint and make changes

catsby:
## Description

Enable `perfsprint` linter and update usage of `fmt.Sprintf` accordingly 

## Type of change

## Checklist before merging


src/cmd/vendored.go
```diff
@@ -51,10 +50,10 @@ var runnerCmd = &cobra.Command{
 		if err = exec.RegisterCmdMutation("uds", executablePath); err != nil {
 			return err
 		}
-		if err = exec.RegisterCmdMutation("zarf", fmt.Sprintf("%s zarf", executablePath)); err != nil {
+		if err = exec.RegisterCmdMutation("zarf", executablePath+" zarf"); err != nil {
 			return err
 		}
-		if err = exec.RegisterCmdMutation("kubectl", fmt.Sprintf("%s zarf tools kubectl", executablePath)); err != nil {
+		if err = exec.RegisterCmdMutation("kubectl", executablePath+" zarf tools kubectl"); err != nil {
```
YrrepNoj, line 56: Learning question; is there a specific reason to avoid using `fmt.Sprintf()` here?

YrrepNoj (approved):
Approved with a quick learning question

src/cmd/vendored.go
catsby: it was flagged by the linter; I assume since it's simple string concatenation there's no formatting required like you might see with `%v` or `%q` or something.

PR #1056, opened 2025-01-16
feat: add dev command to extract Zarf Packages from UDS Bundles

YrrepNoj:
## Description

Adds a `uds dev extract` subcommand to extract `zarf-package-*.tar.zst` tarballs out of `uds-bundle-*.tar.zst` bundles.

Note: This is currently living as a dev command for utility and easing the development of the `tofu plan/apply` functionality. It is my expectation that this logic will eventually be moved into the `uds deploy` logic directly.

Note: I manually updated the `zarf-dev/zarf` dep to `v0.46.0`.

## Type of change

## Checklist before merging


src/pkg/bundle/dev.go
```diff
@@ -87,3 +95,57 @@ func (b *Bundle) SetDeploySource(srcDir string) {
...
+	err = archiveCmd.Execute()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (b *Bundle) Extract(path string) error {
+	if err := helpers.CreateDirectory(path, helpers.ReadWriteExecuteUser); err != nil {
```
catsby, line 139: nit: `b.extractPackage()` calls `zarfUtils.MakeTempDir()` which then calls `helpers.CreateDirectory()`, which itself calls `os.MkdirAll()` which makes me think we can remove this call here.

catsby (approved):

src/pkg/bundle/dev.go
YrrepNoj: The core difference between this call to `helpers.CreateDirectory()` is that the directory created here isn't purged when the command is finished (no call to `defer os.RemoveAll(...)`. 

This is a design implementation detail I am more than willing to budge on, but I intended to be able to run the command and get the following out:

`uds dev extract uds-bundle-my-bundle.tar.zst ./my-bundle-packages`
```
# tree

uds-bundle-my-bundle.tar.zst
my-bundle-packages/
├── zarf-package-my-package-1-arm64-0.13.1.tar.zst
├── zarf-package-my-package-2-arm64-0.13.1.tar.zst
└── zarf-package-my-package-3-arm64-0.13.1.tar.zst
```


And if a subdir wasn't desired:

`uds dev extract uds-bundle-my-bundle.tar.zst .`
```
# tree

uds-bundle-my-bundle.tar.zst
zarf-package-my-package-1-arm64-0.13.1.tar.zst
zarf-package-my-package-2-arm64-0.13.1.tar.zst
zarf-package-my-package-3-arm64-0.13.1.tar.zst
```

YrrepNoj: After thinking about it a second time, I might nit myself and put this under a flag, that might make the fact that this is a 'long lived' directory more obvious.

In my head, I am now comparing this to the `-C -> directory` flag in `tar`.

YrrepNoj: Third look, I will be able to clean this up. I was extracting the oci layers to this directory when I didn't need to be which is probably what added to the confusion.

YrrepNoj: @catsby does this look any better to you now?

catsby: LGTM 👍

catsby (approved):
LGTM

PR #1057, opened 2025-01-16
chore: embed tofu binary and expose 'plan' and 'apply'

YrrepNoj:
## Description

Working towards embedding the `tofu` binary within the UDS CLI. This embedded binary currently only exposes the `tofu plan (uds plan)` and `tofu apply (uds apply)` subcommands.

TODO: 
- [x] Implement a way to ensure the Tofu binary is in the correct location at build time so we can `go:embed` the binary into our code. Right now we only have a naive implementation that makes the following assumptions:
  1. `tofu` is somewhere on your path
  2. You are ONLY building the UDS CLI for your specific arch (since it is ONLY using the already existing tofu for your arch) 

## Type of change

## Checklist before merging

### DIFF
--- a/src/cmd/vendored.go
+++ b/src/cmd/vendored.go
@@ -4,9 +4,13 @@
 package cmd
 
 import (
+	"bytes"
 	"context"
+	_ "embed"
 	"fmt"
+	"io"
 	"os"
+	osExec "os/exec"
 	"runtime/debug"
 
 	runnerCLI "github.com/defenseunicorns/maru-runner/src/cmd"
@@ -23,6 +27,11 @@ import (
 
 	securityHub "github.com/defenseunicorns/uds-security-hub/cmd"
 )
+
+// NOTE: the bin directory needs to be within the directory or subdirectories of this file at compile time
+
+//go:embed bin/tofu
+var tofuCLI []byte
 
 var runnerCmd = &cobra.Command{
 	Use:     "run",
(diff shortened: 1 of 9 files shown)


tasks.yaml
```diff
@@ -54,6 +54,13 @@ tasks:
         setVariables:
           - name: BUILD_ARGS
 
+  # TODO: This is a naive (AND BROKEN!!!) solution. This will only work for a single architecture.
+  #       We need to create subcommands to install specific arch + OS versions of Tofu
+  - name: copy-tofu-bin
+    description: download tofu for go:embed
+    actions:
+      - cmd: cp -Lr $(which tofu) src/cmd/bin/
```
catsby, line 62: should we do a `mkdir -p src/cmd/bin` first here?

catsby: Also I think we need to adjust file permissions? I ran this and it did move the `tofu` binary but follow up `uds run build-` failed:

```
     cp: src/cmd/bin/tofu: Permission denied
     ERROR:  Failed to run action: command "cp -Lr $(which tofu) src/cmd/bin/" timed out after 0 seconds
```

The permissions I'm seeing are `555` (I'm using Homebrew for my `tofu` install)

YrrepNoj: Good call on the `mkdir`, I forgot that I manually made that myself.

The permission decided is another good catch. That is happening because the destination file already exists. I'll add the `-f` flag to the cp command.

tasks/docs.yaml
```diff
@@ -5,6 +5,8 @@ tasks:
   - name: update
     description: updates the CLI docs
     actions:
+      - cmd: mkdir -p src/cmd/bin/
+      - cmd: touch src/cmd/bin/tofu
```
YrrepNoj, line 9: @catsby I hate this but I'm not sure what the best way to handle this might be. Do you have any thoughts?

Context: 
The docs, schema, and unit tests error out if the file the `go:embedd` is reference doesn't exist. Here and elsewhere, I added these commands to ensure _something_ exists at that path.

I really don't like adding these series of commands in several places but at the same time I dont want to check-in a placeholder file at that path because it will make it SUPER easy to accidentally commit the binary itself and muckup our git history with a large file.

YrrepNoj: Synced in person.

tasks.yaml
```diff
@@ -54,6 +54,42 @@ tasks:
...
+      - cmd: rm src/cmd/bin/tofu_1.9.0_linux_arm64.zip
+
+  - name: download-tofu-linux-amd
+    description: download tofu for go:embed
+    actions:
+      - cmd: mkdir -p src/cmd/bin
+      - cmd: rm -f src/cmd/bin/tofu
+      - cmd: wget https://github.com/opentofu/opentofu/releases/download/v1.9.0/tofu_1.9.0_linux_amd64.zip -O src/cmd/bin/tofu_1.9.0_linux_amd64.zip
+      - cmd: unzip -p src/cmd/bin/tofu_1.9.0_linux_amd64.zip tofu > src/cmd/bin/tofu
+      - cmd: rm src/cmd/bin/tofu_1.9.0_linux_amd64.zip
```
YrrepNoj, line 91: Now that I know this works, all of this could be variableized more. However, given that I hope we move over to using a Make styled build system, I'm not too inclined to really spend the time to spend more time with this until we finis out more of our POC.

src/cmd/vendored.go
```diff
@@ -83,6 +92,50 @@ var zarfCli = &cobra.Command{
...
+}
+
+var planCmd = &cobra.Command{
+	Use:   "plan",
+	Short: lang.CmdBundlePlanShort,
+	// Args:  cobra.MaximumNArgs(0),
+	RunE: func(_ *cobra.Command, _ []string) error {
+		return useEmbeddedTofu()
+	},
+	DisableFlagParsing: true,
```
catsby, line 126: `DisableFlagParsing` messes with my feature flag implementation, but we'll cross that bridge later

src/config/lang/lang.go
```diff
@@ -51,6 +51,12 @@ const (
 	CmdBundleDeployFlagRetries  = "Specify the number of retries for package deployments (applies to all pkgs in a bundle)"
 	CmdBundleDeployFlagRef      = "Specify which zarf package ref you want to deploy. By default the ref set in the bundle yaml is used."
 
+	// bundle plan
+	CmdBundlePlanShort = "Generate a Tofu execution plan"
+
+	// bundle plan
```
catsby, line 57:
```suggestion
	// bundle apply
```

catsby (approved):
LGTM

PR #1060, opened 2025-01-23
feat: add 'dev tofu-create' command to build bundles from a uds-bundle.tf config

YrrepNoj:
## Description

Parses `uds-bundle.tf` files and generates a `uds-bundle-fake-metadata-0.0.0.tar.zst` tarball bundle.

There is (a lot) more work that will need to be implemented to ensure full feature parity with the standard `uds create` from a `uds-bundle.yaml` file, but I believe this PR offers a good starting point.

The changes in this PR add additional config booleans to check against so we do not change the behavior of a standard `uds create` command.

## Type of change

## Checklist before merging
😰😬🙈

### DIFF
--- a/src/pkg/bundle/create.go
+++ b/src/pkg/bundle/create.go
@@ -33,7 +33,6 @@ func (b *Bundle) Create() error {
 			return err
 		}
 
-		// Fake metadata for now
 		b.bundle.Kind = "UDSBundle"
 		b.bundle.Metadata = types.UDSMetadata{
 			Name:         tfConfig.Metadata.Name,
@@ -42,7 +41,7 @@ func (b *Bundle) Create() error {
 			Description:  *tfConfig.Metadata.Description,
 		}
 
-		// Parse each Package resoruce and convert it types.Package type
+		// Parse each tfparser.Packages resource and convert it to types.Package type
 		for _, pkg := range tfConfig.Packages {
 			newPackage := types.Package{
 				Name:       pkg.Name,
@@ -120,7 +119,7 @@ func (b *Bundle) Create() error {
 		}
 	}
 
-	// TODO: @JPERRY this dev eploy block has not been validated to work with tofu based bundles yet
+	// TODO: @JPERRY this dev deploy block has not been validated to work with tofu based bundles yet
 	// for dev mode update package ref for local bundles, refs for remote bundles updated on deploy
 	if config.Dev && len(b.cfg.DevDeployOpts.Ref) != 0 {
 		for i, pkg := range b.bundle.Packages {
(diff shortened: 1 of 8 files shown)


src/pkg/bundle/create.go
```diff
@@ -23,9 +24,38 @@ import (
...
+	// populate the b.bundle struct with information on the packages we are creating
+	if b.cfg.IsTofu {
+		// read the .tf data to determine which resources (packages) we are creating
+		// TODO: @JEPRRY consider making a helper function for this
+		tfConfig, err := tfparser.ParseFile(b.cfg.CreateOpts.BundleFile)
+		if err != nil {
+			return err
+		}
+
+		// Fake metadata for now
```
catsby, line 36: still fake?

src/pkg/bundle/create.go
```diff
@@ -23,9 +24,38 @@ import (
...
+		b.bundle.Kind = "UDSBundle"
+		b.bundle.Metadata = types.UDSMetadata{
+			Name:         tfConfig.Metadata.Name,
+			Version:      tfConfig.Metadata.Version,
+			Architecture: *tfConfig.Metadata.Architecture,
+			Description:  *tfConfig.Metadata.Description,
+		}
+
+		// Parse each Package resoruce and convert it types.Package type
```
catsby, line 45:
```suggestion
		// Parse each Package resoruce and convert it to types.Package type
```

src/pkg/bundle/create.go
```diff
@@ -88,6 +120,7 @@ func (b *Bundle) Create() error {
 		}
 	}
 
+	// TODO: @JPERRY this dev eploy block has not been validated to work with tofu based bundles yet
```
catsby, line 123:
```suggestion
	// TODO: @JPERRY this dev deploy block has not been validated to work with tofu based bundles yet
```

src/pkg/bundler/localbundle.go
```diff
@@ -107,17 +110,37 @@ func (lo *LocalBundle) create(signature []byte) error {
...
+		if err != nil {
+			return err
+		}
+	} else {
+		// push uds-bundle.yaml to OCI store
+		bundleManifestDesc, err = pushBundleYAMLToStore(store, bundle)
+		if err != nil {
+			return err
+		}
 	}
```
catsby, line 125: nit: style suggestion

```suggestion
	var bundleManifestDesc ocispec.Descriptor
	if lo.isTofu {
		// push uds-bundle.tf to OCI store
		bundleManifestDesc, err = pushBundleTFToStore(store, filepath.Join(lo.sourceDir, config.BundleTF))
	} else {
		// push uds-bundle.yaml to OCI store
		bundleManifestDesc, err = pushBundleYAMLToStore(store, bundle)
	}
	if err != nil {
		return err
	}
```

catsby (approved):
Overall looks solid, great job 👍

PR #1064, opened 2025-02-04
feat: add provider as layer to bundle

YrrepNoj:
## Description
Adds the provider as a layer with the OCI artifact that represents the UDS Bundle. When it comes time to deploy the bundle, the CLI writes a temporary `.terraformrc` file to the tmpdir that points to the provider that was added to the bundle.

Depends on https://github.com/defenseunicorns/terraform-provider-uds/pull/4.

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


src/cmd/common.go
```diff
@@ -83,6 +83,13 @@ func deploy(bndlClient *bundle.Bundle) error {
 		// Determine the location of the extracted *.tf files
 		chdirFlag := fmt.Sprintf("-chdir=%s", filepath.Dir(bndlClient.GetDefaultExtractPath()))
 
+		// Configure tofu to use the terraformrc file that came with the bundle so it knows where to locate the providers
+		customRCPath := filepath.Join(filepath.Dir(bndlClient.GetDefaultExtractPath()), config.TerraformRC)
+		os.Setenv("TF_CLI_CONFIG_FILE", customRCPath)
```
catsby, line 88: nit: we never unset these env vars. Could we build an input type for `udsEmbeddedTofu` with these values and then include them in the `osExec.Command` call by appending to the `cmd.Env`? 

Something like this (crude) example:

```
type tofuInput struct {
  Env map[string]string
}

input := tofuInput{}
input.Env["TF_CLI_CONFIG_FILE"] = customRCPath

useEmbeddedTofu(input)

// in useEmbeddedTofu

cmd := osExec.Command(// blah)
cmd.Env = os.Environ()
for key, val := range input.Env {
  cmd.Env[key]=val
}
cmd.Stdout = os.Stdout
cmd.Stderr = os.Stderr
err = cmd.Run()
return err

```

- https://pkg.go.dev/os/exec#example-Command-Environment

src/cmd/vendored.go
```diff
@@ -121,7 +125,7 @@ var planCmd = &cobra.Command{
 	Short: lang.CmdBundlePlanShort,
 	// Args:  cobra.MaximumNArgs(0),
 	RunE: func(_ *cobra.Command, _ []string) error {
-		return useEmbeddedTofu()
+		return useEmbeddedTofu(map[string]string{})
```
catsby, line 128: nit: can use `nil` here `return useEmbeddedTofu(nil)`

- https://play.golang.com/p/yU4MhgwI56y

src/pkg/bundle/deploy.go
```diff
@@ -293,6 +293,26 @@ func (b *Bundle) PreDeployValidationTF() (string, string, string, error) {
...
+
+	// Copy the contents of the terraform provider binary into a file in the tmpdir
+	terraformProvider, err := os.ReadFile(filepaths[config.TerraformProvider])
+	if err != nil {
+		return "", "", "", err
+	}
+	err = os.WriteFile(filepath.Join(b.tmp, config.TerraformProvider), terraformProvider, 0777) //nolint:gosec
+	if err != nil {
+		return "", "", "", err
+	}
```
catsby, line 314: nit: if the binary is large-ish then `io.Copy()` might be better here than reading the entire binary first, but that might not be a problem

src/pkg/bundler/localbundle.go
```diff
@@ -113,29 +113,45 @@ func (lo *LocalBundle) create(signature []byte) error {
...
-		// append uds-bundle.yaml layer to rootManifest and grab path for archiving
 		rootManifest.Layers = append(rootManifest.Layers, bundleConfigManifestDesc)
 		digest := bundleConfigManifestDesc.Digest.Encoded()
 		artifactPathMap[filepath.Join(lo.tmpDstDir, config.BlobsDir, digest)] = filepath.Join(config.BlobsDir, digest)
+
+		// append the terraform provider binary to the bundle
+		// TODO:: This is pulling from the local filesystem until we start publishing the provider externally
+		goBinPath := os.Getenv("GOBIN")
+		if goBinPath == "" {
+			return fmt.Errorf("GOBIN environment variable was not set, unable to locate the locally installed terraform-provuder-uds binary")
```
catsby, line 145:
```suggestion
			return fmt.Errorf("GOBIN environment variable was not set, unable to locate the locally installed terraform-provider-uds binary")
```

src/pkg/bundler/localbundle.go
```diff
@@ -113,29 +113,45 @@ func (lo *LocalBundle) create(signature []byte) error {
...
+		}
+		providerPath := filepath.Join(goBinPath, config.TerraformProvider)
```
catsby, line 147: We may want to pull this from the `.terraform` directory that gets created with `tofu init` instead. It's probably normal to not have the provider actually installed in the gobin and instead just pulled down in to the CWD `.terraform`. We could probably use `filepath.Walk()` to find it. 

- https://pkg.go.dev/path/filepath#Walk

For the sake of this PR we could leave it as-is though and make it a follow-up issue. Up to you.

catsby (approved):
There are some changes we want to make in the future but there's nothing blocking at the moment, so I think this is good-to-go for the sake of getting a viable demo.

src/pkg/bundler/localbundle.go
YrrepNoj: Why do today what you can do tomorrow?

> It's probably normal to not have the provider actually installed in the gobin and instead just pulled down in to the CWD.

I agree that looking in GOBIN is not a good solution, but we should follow up with the create flow to make sure we're on the same page. I think I'm operating off the mindset that the things `tofu init` would do should be handled automagically when people use the UDS CLI?

catsby: I agree `init` should likely have been called already, but `init` does not install the provider in `GOBIN` path, if that's what you're thinking?

PR #1066, opened 2025-02-05
chore: update zarf version ref to v0.48.1

YrrepNoj:
## Description

...

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


src/pkg/utils/sbom.go
```diff
@@ -104,7 +115,7 @@ func HandleSBOM(extractSBOM bool, SBOMArtifactPathMap map[string]string, bundleN
 		if err != nil {
 			return warns, err
 		}
-	} else if len(SBOMArtifactPathMap) > 0 {
+	} else if len(SBOMArtifactPathMap) > 0 && sbomArtifactCount > 0 {
```
catsby, line 118: nit: `sbomArtifactCount` cannot be greater than `0` if `len(SBOMArtifactPathMap)` is not also, so we could simplify this to just `else if sbomArtifactCount > 0` yeah?

catsby (approved):

PR #1076, opened 2025-02-07
chore: update go deps besides archiver

catsby:
## Description

Updates our go.mod deps besides the `archiver` dependency; that one is more complicated now that https://github.com/mholt/archiver is deprecated and we'll need to do some more wrangling to move to it's replacement. For now though this will update some other deps.

## Type of change

## Checklist before merging

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module github.com/defenseunicorns/uds-cli
 
-go 1.23.3
-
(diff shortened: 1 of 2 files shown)


catsby:
closing in favor of https://github.com/defenseunicorns/uds-cli/pull/1087 because `pterm` breaks things for reasons

PR #1096, opened 2025-03-01
fix!: use provided public signing key to build bundles with local signed zarf packages

YrrepNoj:
fixes issue where bundles that referenced a Signed Zarf Package on the local filesystem were unable to be built

**NOTE:**
I think this should be considered (and advertised) as a breaking change. Bundles that were utilizing signed Zarf Packages without providing the publickey are now going to give an error when being created instead of silently failing/continuing before.

### DIFF
--- a/src/pkg/bundler/fetcher/remote.go
+++ b/src/pkg/bundler/fetcher/remote.go
@@ -61,7 +61,10 @@ func (f *remoteFetcher) Fetch() ([]ocispec.Descriptor, error) {
 	f.pkg.PublicKey, err = getAbsKeyPath(f.pkg.PublicKey, f.cfg.CreateSrcDir)
 	if err != nil {
 		return nil, err
-	} else if f.pkg.PublicKey != "" {
+	}
+
+	// If the publicKey was provided, add the key flag to the args list
+	if f.pkg.PublicKey != "" {
 		cmdArgs = append(cmdArgs, "--key="+f.pkg.PublicKey)
 	}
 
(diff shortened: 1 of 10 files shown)


src/pkg/bundler/fetcher/remote.go
```diff
@@ -35,86 +36,55 @@ func (f *remoteFetcher) Fetch() ([]ocispec.Descriptor, error) {
...
+	}
 
-	// copy layers to local bundle
-	fetchSpinner.Updatef("Pushing package %s layers to bundle (package %d of %d)", f.pkg.Name, f.cfg.PkgIter+1, f.cfg.NumPkgs)
-	pkgDescs, err := f.copyRemotePkgLayers(layersToCopy)
+	// Add path to public key if provided
+	f.pkg.PublicKey, err = getAbsKeyPath(f.pkg.PublicKey, f.cfg.CreateSrcDir)
 	if err != nil {
 		return nil, err
+	} else if f.pkg.PublicKey != "" {
```
catsby, line 64: nit: remove the else and make this a separate `if` block
```suggestion
	}
	
	if f.pkg.PublicKey != "" {
```

> In the Go libraries, you'll find that when an if statement doesn't flow into the next statement—that is, the body ends in break, continue, goto, or return—the unnecessary else is omitted.


- https://go.dev/doc/effective_go#control-structures

src/pkg/bundler/fetcher/remote.go
```diff
@@ -35,86 +36,55 @@ func (f *remoteFetcher) Fetch() ([]ocispec.Descriptor, error) {
...
 
-		manifestConfigDesc, err := boci.ToOCIStore(f.pkgRootManifest.Config, zoci.ZarfConfigMediaType, f.cfg.Store)
-		if err != nil {
-			return nil, err
-		}
-		descsToBundle = append(descsToBundle, pkgManifestDesc, manifestConfigDesc)
+	matches, err := filepath.Glob(filepath.Join(outputDir, "zarf-*.tar.zst"))
+	if err != nil {
+		return nil, err
+	} else if len(matches) != 1 {
```
catsby, line 83: same comment on breaking out the `if` here

catsby (approved):
LGTM assuming tests pass. 1 suggestion in 2 places

catsby (approved):

PR #1099, opened 2025-03-05
chore: replace usage of archiver/v3 with archives

YrrepNoj:
## Description

Slight cleanup/QoL PR. We utilizing multiple versions of an old archiver library. This PR removes all usage of the unmaintained library and uses the new `mholt/archives` library that we were already using in most other places in this project.

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


src/config/config.go
```diff
@@ -97,6 +98,13 @@ var (
 
 	// Dev specifies if we are running in dev mode
 	Dev = false
+
+	// BundleArchiveFormat is the mholt/archives compression format for managing bundles on disk
+	BundleArchiveFormat = archives.CompressedArchive{
```
YrrepNoj, line 103: TODO: I believe there are other places that can use this config Format.

src/pkg/bundle/publish.go
```diff
@@ -41,8 +43,38 @@ func (b *Bundle) Publish() error {
...
+			return err
+		}
+		defer stream.Close()
+
+		fileBytes, err := io.ReadAll(stream)
+		if err != nil {
+			return err
+		}
+
+		return os.WriteFile(extractPath, fileBytes, 0644)
```
YrrepNoj, line 69: Here and elsewhere;

I kind of just smashed against my numpad when setting file & directory permissions. We should probably use a tiny bit of brainpower to make sure the perms we set make sense.

src/pkg/bundle/publish.go
```diff
@@ -41,8 +43,38 @@ func (b *Bundle) Publish() error {
 		return err
 	}
 
-	// unarchive bundle into empty tmp dir
-	err = av3.Unarchive(b.cfg.PublishOpts.Source, b.tmp) // todo: awkward to use old version of mholt/archiver
+	// construct a fileHandler that extracts all files from the archive with the same relative pathing
+	fileHandler := func(_ context.Context, file archives.FileInfo) error {
```
catsby, line 47: I see similar functions were added to `utils/utils.go`, should this go there as well?

catsby (approved):
LGTM, just 1 comment about adding to utils

src/pkg/bundle/publish.go
YrrepNoj: You are right, the `utils.ExtractFile()` helper basically does the exact same thing as this. I will edit that util function and use that instead.

src/pkg/utils/utils.go
```diff
@@ -119,6 +119,68 @@ func ExtractJSON(j any, expectedFilepath string) archives.FileHandler {
...
+	}
+}
+
+func ExtractAllFiles(outDirPath string) archives.FileHandler {
+	return ExtractFile("", outDirPath)
+}
+
+// ExtractFile returns an archives.FileHandler that extracts a file from an archive
+// NOTE: if expectedFilepath is an emtpy string, all files within the archive will be extracted
+func ExtractFile(expectedFilepath string, outDirPath string) archives.FileHandler {
```
YrrepNoj, line 151: @catsby do you have opinions on the `expectedFilepath` param having a behavior of "extract everything" if the input is an empty string?

catsby (approved):

PR #1101, opened 2025-03-06
chore(test): add zarf@main integration test nightly

brandtkeller:
## Description

Quick nightly workflow for testing the integration of zarf@main. 

Could use a team notification workflow - such as `create-or-update-comment` action but this PR is simply for some quick feedback. 

## Related Issue

No related issue yet

## Type of change

## Checklist before merging

### DIFF
--- a/.github/workflows/nightly-integration-test.yaml
+++ b/.github/workflows/nightly-integration-test.yaml
@@ -1,0 +1,17 @@
+# Copyright 2024 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+name: Nightly Build and Test with zarf@main
+on:
+  schedule:
+    - cron: '0 7 * * *' ## Every day at 0700 UTC
+  workflow_dispatch: # Allows manual triggering
+
+permissions:
+  contents: read
+
+jobs:
+  nightly-integration:
+    uses: ./.github/workflows/parallel-tests.yaml
+    with:
+      fetch_zarf: true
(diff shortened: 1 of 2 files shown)


.github/workflows/parallel-tests.yaml
```diff
@@ -4,6 +4,12 @@
 name: Helper workflow for E2E Tests
 on:
   workflow_call: # This is the event that triggers the workflow
+    inputs:
+        fetch_zarf:
```
YrrepNoj, line 8: What are the teams thoughts about making this input (or likely adding an additional input) a ref string?

The nightly cron can trigger with the ref of `main` but we could trigger a dispatch with a specific branch/ref if there's a specific branch we're interested in testing compatibility against.

YrrepNoj (approved):

.github/workflows/parallel-tests.yaml
catsby: I would support this enhancement but we can also do it ourselves later

.github/workflows/nightly-integration-test.yaml
```diff
@@ -0,0 +1,17 @@
...
+  schedule:
+    - cron: '0 7 * * *' ## Every day at 0700 UTC
+  workflow_dispatch: # Allows manual triggering
+
+permissions:
+  contents: read
+
+jobs:
+  nightly-integration:
+    uses: ./.github/workflows/parallel-tests.yaml
```
catsby, line 15: We'll likely need to use a different test suite or make a new one, as a few of them rely on packages we build during the initial tests, and somewhere in the chain we parse out the version of zarf from `go.mod` and then try to use that version in the `uds-bundle.yaml`. A typical zarf version upgrade PR updates this version automatically: 

- https://github.com/defenseunicorns/uds-cli/pull/1103/files#diff-a3cf35511c9c12330b39cc9d96833317ec7f1740866f98d14bc85f9c4e0dea9d

I can work on crafting a test that doesn't rely on that though

PR #1118, opened 2025-04-02
feat: migrate from message->slog

AustinAbro321:
In the next version of Zarf the `--legacy` flag for old logs is going to be deleted along with some of the functions in message. Here is an example of how you can change your logs to work with the new Zarf logger (you could also create your own slog implementation). 

Still to do is to change your calls to message to calls to the slogger. Technically it's not necessary, but things look a bit awkward with both
image

Feel free to edit this PR or create your own PR using this as inspiration. I will not be editing it further.


catsby:
superseded by https://github.com/defenseunicorns/uds-cli/pull/1120, thanks again @AustinAbro321

PR #1119, opened 2025-04-03
chore: minor io handling updates that avoid reading things in to memory

catsby:
## Description

I noticed there are some points where we read potentially large files in to memory, to then write them somewhere else (in part or in whole). In this PR we effectively stream the bytes from A->B instead of loading into memory first.

## Type of change

## Checklist before merging


catsby:
superseded (and cherry-picked) by https://github.com/defenseunicorns/uds-cli/pull/1129

PR #1147, opened 2025-05-14
fix: ignore eslint config for license

cmwylie19:
## Description

ESLint released a breaking change in v9.x which forced Pepr to also make a breaking change in order to update ESLint. Version 9.x of ESLint is configured differently and no longer use JSON files, instead now it is a `eslint.config.mjs`. 

We want to get ahead of any trouble in UDS Core and feel 💯 confident that this feature will not have impact on the product before the release so I am doing some testing on the nightly release of Pepr.

The problem is that the `addlicense` tool is [timing out after 0](https://github.com/defenseunicorns/uds-core/actions/runs/15022736270/job/42215747267?pr=1564#step:3:156) seconds on the new `eslint.config.mjs` in UDS Core CI.

Based on this error I would like to propose adding `eslint.config.mjs` to the ignore flags of the tool.

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/tasks/lint.yaml
+++ b/tasks/lint.yaml
@@ -37,6 +37,9 @@ tasks:
             ignore_flags+=("-ignore" "\"$line\"")
           done < .gitignore
 
+          # Ignore eslint config
+          ignore_flags+=("-ignore" "eslint.config.mjs")
+
           # Construct the command as a string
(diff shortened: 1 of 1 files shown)


cmwylie19:
wrong repo here. Closing this

PR #1267, opened 2025-12-01
chore: add broken link checker for docs

chance-coleman:
## Description
When we commit new docs or rename docs in our docs/ directory, sometimes they have links that are broken and are not caught until the docs site rebuilds and we notice the failure.

Also updated several docs links to use relative linking instead of URL links.

## Type of change

## Checklist before merging


mjnagel (approved):
LGTM

PR #1276, opened 2025-12-12
feat(list): add support for uds list command

### LINKED ISSUE #458: Add `list` to uds commands to display deployed bundles

uds currently supports commands `create`, `deploy` and `remove` but does not give the user the ability to `list` currently deployed uds bundles

When I deploy a uds bundle

I'm able to run `uds list` to display bundle name, version, zarf pkgs within bundle

So that I have an understanding of what's been deployed before applying upgrades, troubleshooting, removing, etc...

run `zarf package list` but this does not display uds bundle info

brandtkeller:
## Description

Adds a `uds list` command and mutates package annotations on deploy to make them able to be mapped back to a bundle natively with zarf.

This is solely for bundles that deploy to clusters but could be extended to support other state retrieval models. 

If a bundle is a collection of packages - the theoretically we should be able to map packages back to bundles if we have some identifying metadata. Using package annotations is one such approach - adding/overwriting with bundle name/version annotations. 

Zarf has standard logic to retrieve deployed packages that would then include these annotations and we simply need to map and sort for expected outputs. This uses zarf state natively without creating any additional persistent state to maintain. 

Consider this for the purpose of driving discussion on the required information.

## Related Issue

Fixes #458

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/bundle/deploy.go
+++ b/src/pkg/bundle/deploy.go
@@ -147,14 +147,14 @@ func deployPackages(ctx context.Context, packagesToDeploy []types.Package, b *Bu
 			IsInteractive:          !config.CommonOptions.Confirm,
 		}
 
-		bundleAnnotation := map[string]string{
-			AnnotationBundleName:    b.bundle.Metadata.Name,
-			AnnotationBundleVersion: b.bundle.Metadata.Version,
-		}
-		maps.Copy(bundleAnnotation, pkgLayout.Pkg.Metadata.Annotations)
+		// Merge package annotations with bundle annotations; bundle annotations take precedence
+		bundleAnnotations := make(map[string]string)
+		maps.Copy(bundleAnnotations, pkgLayout.Pkg.Metadata.Annotations)
+		bundleAnnotations[AnnotationBundleName] = b.bundle.Metadata.Name
+		bundleAnnotations[AnnotationBundleVersion] = b.bundle.Metadata.Version
 
 		// Set the merged annotations back on the package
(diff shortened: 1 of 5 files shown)


src/pkg/bundle/deploy.go
```diff
@@ -146,6 +147,15 @@ func deployPackages(ctx context.Context, packagesToDeploy []types.Package, b *Bu
 			IsInteractive:          !config.CommonOptions.Confirm,
 		}
 
+		bundleAnnotation := map[string]string{
+			AnnotationBundleName:    b.bundle.Metadata.Name,
+			AnnotationBundleVersion: b.bundle.Metadata.Version,
+		}
+		maps.Copy(bundleAnnotation, pkgLayout.Pkg.Metadata.Annotations)
```
mjnagel: If I'm not mistaken, this would allow the package annotations to win in a case of collision (package setting either bundle name or version annotations). This would ultimately have potential to mess up/confuse the list command (i.e. a package could end up in a different bundle or version). Seems worth overwriting imo, and in the UDS ecosystem we just treat these annotations as "reserved" (not meant to be used at package create time)?

mjnagel (commented):
I think this is great for a lightweight list implementation. Did some basic testing locally and everything worked how I'd expect, just had a few comments. One interesting thing I checked was the behavior with `--packages` if you upgrade part of a bundle to a different version. Tracking works great to show that as 2 separate bundles deployed, with the packages from their respective versions (what I expected, but figured I'd test it anyways).

Image

PR #1283, opened 2026-01-16
fix(docs): typo in bundle overrides ref

daniel-s-palmer:
## Description

Fixed a small typo in the bundle override reference doc

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/docs/reference/bundles/overrides.md
+++ b/docs/reference/bundles/overrides.md
@@ -242,7 +242,9 @@ There are 3 ways to override the `UI_COLOR` variable:
    uds deploy example-bundle --set helm-overrides-package.ui_color=green
    ```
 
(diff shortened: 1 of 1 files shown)


daniel-s-palmer:
Noticed as well just below this section that the warning doesn't render the emoji on the website. Happy to update that to a 
:::warning:::
if we think that makes sense 
image

chance-coleman (commented):
Yeah thats a good find, the astro config that we use, uses `::caution::` instead of `::warning::`. So if you want to update that, i can approve this

daniel-s-palmer:
@chance-coleman Awesome. Updated and used the syntax seen just a bit below in the same file.

PR #1293, opened 2026-01-21
Zarf values exploratory preview

michalszynkiewicz:
refs #1294 

To see how Zarf Values would look like, I tried them with Gitlab package.

* how a single var->value migration looks https://github.com/uds-packages/gitlab/pull/358 
* how migrating everything to zarf values looks: https://github.com/uds-packages/gitlab/pull/357


joelmccoy:
closing this. Zarf values will be incorporated in uds cli next which is being ported to this repository as we speak.  We will not be supporting zarf values for the current interation/logic of this cli

PR #1337, opened 2026-03-03
feat: added timeout override for zarf package deployments

### LINKED ISSUE #588: Support timeout as package parameter

Some packages take longer to deploy than others. Currently there is no mechanism to override the default timeout of 15 minutes, which can lead to very lengthy cycles on errors that should only take a few seconds to address.

The addition of a `timeout` field in the package definition could [be passed along as a deploy parameter](https://github.com/defenseunicorns/uds-cli/blob/df4aa095fcd5b928e025270f89b5d8dee45d1663/src/pkg/bundle/deploy.go#L119) and would allow a user to set a timeout for each zarf package contained in the bundle. 

As a bundle maintainer
When I define packages to be deployed
I would like the option to set a timeout for each individual package

This may look something like this:
```
kind: UDSBundle
metadata:
  name: some-bundle

packages:
  - name: core
    repository: ghcr.io/defenseunicorns/packages/uds/core
    ref: 0.20.0-registry1
    timeout: 10m
```

JoshuaFurman:
## Description
UDS doesn't currently expose the ability to override the deployment timeout of zarf packages in a UDS-bundle. This feature adds support to setting a timeout value override per zarf package in your uds-bundle.yaml file. 

## Related Issue

Fixes #588 

## Type of change

## Checklist before merging

## Misc
Following the contribution guidelines, I was unable to run the e2e test defined here: https://github.com/defenseunicorns/uds-cli/blob/main/CONTRIBUTING.md#running-tests as the `test:e2e-no-ghcr-write` task no longer exists but I ran all other automated tests and manual tests.

### DIFF
--- a/src/pkg/bundle/timeout_test.go
+++ b/src/pkg/bundle/timeout_test.go
@@ -81,6 +81,7 @@ func TestValidatePackageTimeouts(t *testing.T) {
 
 	err := validatePackageTimeouts([]types.Package{
 		{Name: "nginx", Timeout: "5m"},
+		{Name: "grafana", Timeout: ""},
 		{Name: "podinfo", Timeout: "not-a-duration"},
 	})
 


mjnagel (changes requested):
Few small changes here - overall I think this makes sense. Would also be good to have a second review to sanity check.

PR #1399, opened 2026-06-05
feat(publish): add force upload option for OCI publishes

cross-du:
## Description

Adds a `uds publish --force-upload` option to work around registries that incorrectly report blobs as already present during upload checks.

This is aimed at the GitLab-style failure mode where publish can appear to succeed, but a later `uds pull` or `uds inspect` fails because the target repository is missing one or more digests.

## What Changed

- Added `--force-upload` to `uds publish`
- Added `UDS_PUBLISH_FORCE_UPLOAD=true` support for pipeline-friendly opt-in usage
- Made the CLI flag take precedence over the environment variable
- Updated publish logic to wrap the ORAS destination repository and force `Exists()` to return `false` when enabled
- Updated CLI help/docs for the new option

## Tests

Added coverage at two levels:

- Unit-style coverage for:
  - flag registration
  - env var parsing
  - env var / flag precedence
  - invalid env var handling
- End-to-end regression coverage for the target failure mode:
  - simulates a registry that returns false-positive blob existence checks
  - shows publish without `--force-upload` leaves a missing blob behind
  - shows a later pull fails on the missing digest
  - shows publish with `--force-upload` uploads the missing blob and pull succeeds

## Verification

Validated with targeted Go tests and the new focused e2e regression test.

## Notes

This change is opt-in by default. Normal publish behavior is unchanged unless `--force-upload` or `UDS_PUBLISH_FORCE_UPLOAD=true` is used.

...

## Related Issue

Fixes CLI-206

## Type of change

## Checklist before merging


slaskawi:
@codex review

src/cmd/uds.go
```diff
@@ -22,6 +23,8 @@ import (
 	"github.com/zarf-dev/zarf/src/pkg/logger"
 )
 
+const publishForceUploadEnvVar = "UDS_PUBLISH_FORCE_UPLOAD"
```
slaskawi: nit: maybe it's worth moving this to `config.go` which seems to contain other env vars.

slaskawi (dismissed):
LGTM with 1 nit that might be worth addressing

emoskito:
@greptileai

src/cmd/uds.go
```diff
@@ -184,6 +190,25 @@ var publishCmd = &cobra.Command{
 	},
 }
 
+func applyPublishForceUploadEnv(cmd *cobra.Command) error {
```
emoskito: `applyPublishForceUploadEnv` is not necessary. You should be able to use Viper's configuration pattern instead to stay consistent with the other commands:

```go
// viper.go
const V_PUBLISH_FORCE_UPLOAD = "options.publish_force_upload"

// uds.go
publishCmd.Flags().BoolVar(
    &bundleCfg.PublishOpts.ForceUpload,
    "force-upload",
    v.GetBool(V_PUBLISH_FORCE_UPLOAD),
    lang.CmdPublishFlagForceUpload,
)
```

Refs: [`--skip-signature-validation`](https://github.com/defenseunicorns/uds-cli/blob/main/src/cmd/root.go#L109) (root boolean) and [`bundle pull --key`](https://github.com/defenseunicorns/uds-cli/blob/main/src/cmd/uds.go#L328)  (string) subcommand uses this same Viper-backed flag-default pattern.

src/test/e2e/commands_test.go
```diff
@@ -189,3 +197,292 @@ func removeZarfInit() {
...
+	}
+
+	trimmed := strings.TrimPrefix(path, "/v2/")
+	parts := strings.SplitN(trimmed, "/blobs/", 2)
+	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+		return "", "", false
+	}
+
+	return parts[0], parts[1], true
+}
```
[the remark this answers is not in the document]
emoskito: Non-blocking as this is limited to test-helper behavior and does not affect current ORAS upload flow; however simple change to tighten with `digest.Parse()`

emoskito (changes requested):
Use Viper configuration pattern instead like other commands instead of manually reading and parsing `UDS_PUBLISH_FORCE_UPLOAD`. `applyPublishForceUploadEnv` func and unit tests should go away.

A couple of non-blockers by greptile and @slaskawi  that may be worth addressing since code changes are required.

src/cmd/uds.go
cross-du: `applyPublishForceUploadEnv` has been removed and the code conforms to Viper's configuration pattern.

src/cmd/uds.go
cross-du: Addressed by following Viper's configuration pattern.

joelmccoy:
@greptileai review

src/cmd/viper.go
```diff
@@ -24,6 +24,7 @@ const (
 	V_TMP_DIR                   = "options.tmp_dir"
 	V_INSECURE                  = "options.insecure"
 	V_NO_COLOR                  = "options.no_color"
+	V_PUBLISH_FORCE_UPLOAD      = "options.publish_force_upload"
```
emoskito: NIT: Could you please rename to `V_BNDL_PUBLISH_FORCE_UPLOAD` and move this out from the root config keys since this is publish-specific. Just below the pull config keys, with an appropriate comment:

```go
//Bundle publish config keys
V_BNDL_PUBLISH_FORCE_UPLOAD      = "options.publish_force_upload"
```

I know functionally, this doesn't change anything, but it's code consistency for us.

cross-du: Totally fair, I've moved it into its own "section."

emoskito (approved):

PR #1410, opened 2026-07-07
fix: disable bundle format signature explicitly

### LINKED ISSUE #1409: Updates to Cosign require changes for parity

The bundle signing logic today creates a standard signature through the use of the zarf `signing` package. 

Cosign `sign_blob` has recently changed in behavior such that it does not produce both a standard signature _and_ the new bundle format signature. uds-cli handles package signing through zarf as expected but bundle signing relies on to production of both signatures (but largely ignored the new bundle signature). 

These changes are coming in the next Zarf release and can be mitigated by setting explicit bundle signing parameters. 

PR to follow.

brandtkeller:
## Description

This sets `NewBundleFormat` to false explicitly as changes in Cosign (consumed through the next Zarf update) would not produce a standard signature. 

This is to keep parity with current processes - it does not introduce any new support. 

Note that `verifyOptions` update is not strictly required but also sets explicit intent so as to not succumb to the same process in any future changes to cosign.

## Related Issue

Fixes #1409

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/bundle/common.go
+++ b/src/pkg/bundle/common.go
@@ -262,6 +262,7 @@ func ValidateBundleSignature(bundleYAMLPath, signaturePath, publicKeyPath string
 	verifyBlobOptions := signing.DefaultVerifyBlobOptions()
 	verifyBlobOptions.Signature = signaturePath
 	verifyBlobOptions.Key = publicKeyPath
+	verifyBlobOptions.CommonVerifyOptions.NewBundleFormat = false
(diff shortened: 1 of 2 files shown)


joelmccoy:
@greptileai review

joelmccoy (approved):
LGTM - Thanks for getting ahead of this.  I don't think realistically anybody are signing bundles right now but to be safe sticking with the legacy format is fine.  We will adopt the new format completely in CLI Next.

PR #1411, opened 2026-07-09
fix: error with deploying uds packages containing image archives via OCI

nicleary:
## Description

Previously, when deploying a bundle that contained a zarf package which used image archives via OCI, an error similar to `failed to deploy bundle: file images/blobs/sha256/060d0cbd1c16ff5ded94971e621bcf61330bc34dbca76e29bb9329319a52797a from checksum missing in layout` would occur.

This was due to UDS CLI only looking at the images contained in the packages, not the images and image archives. Used Zarf's `GetImages()` helper to handle this.

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


src/pkg/utils/boci/oci.go
```diff
@@ -301,7 +301,7 @@ func FilterImageIndex(components []v1alpha1.ZarfComponent, imgIndex ocispec.Inde
 	manifestIncludeMap := map[string]ocispec.Descriptor{}
 	for _, manifest := range imgIndex.Manifests {
 		for _, component := range components {
-			for _, imgName := range component.Images {
+			for _, imgName := range component.GetImages() {
```
brandtkeller, line 304: This is the first instance - for sake of completeness i'd expect similar adaptation to src/pkg/bundle/inspect.go

uds-bundle.yaml
```diff
@@ -0,0 +1,53 @@
+kind: UDSBundle
```
brandtkeller, line 1: assuming this is for use of testing - albeit I don't have the required access

brandtkeller (changes requested):
minor addition requested for completeness

uds-bundle.yaml
brandtkeller: (now outdated -resolved)

PR #1427, opened 2026-07-28
feat: mask sensitive zarf vars

### LINKED ISSUE #901: Handle Sensitive Zarf Vars
Zarf Variables can be specified as sensitive in a Zarf package yaml. Ensure that we check if any Zarf vars are sensitive and if so mask them in uds output.

Some considerations:
This requires loading the zarf package metadata, which we can use `loadPackage()` from `inspect.go` for. That means we're now loading packages at two points in the deployment flow -- pre-deploy views and actual deployment. 

Alternate Options:
We don't display any Zarf vars and their values in the pre-deploy view. When the bundle deploys, each package definition with its vars and their current values are displayed and Zarf already handles masking the sensitive values in that output.

### LINKED ISSUE #1417: uds inspect v0.34.x can fail on valid pre-v0.34 bundles due to tar entry ordering
### Environment

Device and OS: Reproduced on a Kairos RHEL9-Common VM and a Linux development host
App version: UDS CLI v0.34.0
Other:

- The affected bundle was created with UDS CLI v0.32.0.
- The same prefetch implementation remains in UDS CLI v0.34.1 and current `main`.
- The inspected Zarf package uses keyless verification.

### Steps to reproduce

1. Create a bundle with a pre-v0.34 UDS CLI version whose package OCI blobs have the following relative tar order:

   1. `checksums.txt`
   2. Package manifest
   3. `zarf.yaml`

   UDS CLI v0.32.0 did not sort tar entries before writing them, so entry order varied between bundle builds.

2. Inspect the bundle using UDS CLI v0.34.0 or v0.34.1 with the default signature-validation behavior:

   ```shell
   uds inspect uds-bundle-bootstrap-amd64-0.7.3.tar.zst \
     --no-progress \
     --no-color \
     --no-log-file
   ```

3. Observe that inspection fails because `checksums.txt` is absent from the prefetched package metadata directory, even though the blob exists in the bundle.

### Expected result

`uds inspect` should successfully inspect valid bundles created by earlier UDS CLI versions regardless of tar entry order.

If the single-pass metadata prefetcher cannot capture all metadata referenced by a package manifest, it should return an error and allow `Inspect()` to use its existing per-package extraction fallback.

This is also the backward-compatibility behavior described by PR #1403:

> New CLI reading old bundle | Works. Read-side logic doesn't assume order.

https://github.com/defenseunicorns/uds-cli/pull/1403

### Actual Result

Inspection fails with:

```text
ERROR: failed to inspect bundle: package "init": stat /tmp/.../pkgmeta-9a7fb42ea822165fa94a081d668b9821ed1dfb4a445ea0d984854ee9ea59368a/checksums.txt: no such file or directory
```

The referenced `checksums.txt` blob exists in the bundle. It was encountered before the package manifest and therefore was not captured by the single-pass prefetcher.

The prefetcher initially knows only the package manifest digest. When it later parses the manifest, it adds the package metadata digests to `needed`. Metadata already passed in the stream cannot be recovered.

After archive extraction completes, the implementation does not check whether `needed` is non-empty:

https://github.com/defenseunicorns/uds-cli/blob/v0.34.1/src/pkg/bundle/tarball_prefetch.go#L91-L227

The result-building code explicitly reports a missing `zarf.yaml`, but it silently skips other missing metadata files. If `zarf.yaml` occurs after the manifest while `checksums.txt` occurs before it, prefetch returns a partial cache as a successful result. Package verification later fails while loading that cache.

The exact same UDS CLI v0.34.0 binary was tested against two bundles created with v0.32.0 and containing the same `init` package manifest digest.

Binary SHA-256:

```text
967b8a09abe4b59116041c120a2de3bf9738ec0edc40c65fdae15fac4dd692cb
```

Package manifest digest:

```text
9a7fb42ea822165fa94a081d668b9821ed1dfb4a445ea0d984854ee9ea59368a
```

| Bundle | `checksums.txt` position | Manifest position | `zarf.yaml` position | Result |
|---|---:|---:|---:|---|
| Failing bundle | 6 | 26 | 44 | Missing `checksums.txt` |
| Successful bundle | 63 | 109 | 102 | Prefetch reports missing `zarf.yaml`, falls back, and succeeds |

The successful archive produces the intended fallback:

```text
WARNING package metadata prefetch failed, falling back to per-package extraction:
zarf.yaml not found for package "init" in bundle
Verified OK
```

This explains why the problem appears intermittent across bundle builds. It is deterministic for a particular archive, but pre-v0.34 bundle entry ordering was nondeterministic.

The existing prefetch test covers the case where all metadata precedes the manifest, causing the explicit missing-`zarf.yaml` error. It does not cover partial ordering where `checksums.txt` precedes the manifest but `zarf.yaml` follows it:

https://github.com/defenseunicorns/uds-cli/blob/v0.34.1/src/pkg/bundle/tarball_prefetch_test.go

### Severity/Priority

Suggested priority: P2.

This is a released backward-compatibility regression that prevents inspecting some otherwise valid bundles created by older UDS CLI versions. Rebuilding with a newer CLI avoids the archive ordering, but existing bundles remain affected.

`--skip-signature-validation` avoids this code path, but it weakens package verification and is not an appropriate general workaround.

### Additional Context

PR #1403 added the single-pass prefetcher and deterministic bundle ordering together:

https://github.com/defenseunicorns/uds-cli/pull/1403

A minimal correctness fix would be to check `len(needed)` after archive extraction. If any required digests remain, return a prefetch error so `Inspect()` invokes the existing slow fallback.

A regression test should use this ordering:

```text
checksums.txt -> package manifest -> zarf.yaml
```

The test should verify that prefetch reports an incomplete result rather than returning a partial cache. An integration-level test could additionally verify that `Inspect()` falls back and successfully inspects the bundle.

rodneyxr:
## Description

The deploy preview previously displayed resolved override values without checking the underlying Zarf package metadata. A variable marked sensitive by a package author could therefore appear in terminal output.

This keeps the preview useful for confirming which variables are applied while preserving the boundary that secret values must not reach the terminal.

This PR makes deployment previews now honor each Zarf package variable’s `sensitive: true` declaration.

* Sensitive Zarf variable values are shown as `****`; names remain visible.
* Redaction applies to values supplied through `uds-config.yaml` and `--set` overrides, in addition to existing environment-variable masking.
* Sensitive-variable matching is case-normalized to match override resolution.
* If package metadata cannot be loaded, the preview masks all Zarf variable values for that package rather than risking a secret leak.
* Package references without a manifest digest now return a clear metadata-loading error instead of panicking. This *should* never happen since its added in a bundle create however it showed up in synthetic tests.

### Tests

Tests cover declared-sensitive masking for config and CLI sources, retain visibility for non-sensitive variables, and verify the conservative masking behavior when package metadata is unavailable.

## Related Issue

Fixes defenseunicorns/uds-cli#901  <br>Fixes defenseunicorns/uds-cli#1417

## Type of change

## Checklist before merging

- [ ] [Contributor Guide Steps](<https://github.com/defenseunicorns/uds-cli/blob/main/CONTRIBUTING.md>) followed

### DIFF
--- a/src/pkg/bundle/deploy.go
+++ b/src/pkg/bundle/deploy.go
@@ -492,7 +492,17 @@ func formPkgViews(b *Bundle) []PkgView {
 			}
 		}
 
-		variables = addZarfVars(variableData, variables)
+		// Load the package metadata to get the list of sensitive variables for masking their values in the view
+		zarfPkg, err := b.getMetadata(pkg)
+		sensitiveVars := sensitiveZarfVariableNames(zarfPkg.Variables)
+		if err != nil {
+			message.WarnErr(err, fmt.Sprintf("unable to load metadata for package %q; masking all Zarf variable values", pkg.Name))
+			for name := range variableData {
+				sensitiveVars[name] = true
+			}
+		}
+
+		variables = addZarfVars(variableData, variables, sensitiveVars)
 		pkgViews = append(pkgViews, PkgView{meta: formPkgMeta(pkg), overrides: map[string]interface{}{"overrides": variables}})
 	}
 	return pkgViews
@@ -514,7 +524,17 @@ func formPkgMeta(pkg types.Package) map[string]string {
 	return pkgMeta
 }
 
-func addZarfVars(pkgVars map[string]overrideData, variables []interface{}) []interface{} {
+func sensitiveZarfVariableNames(variables []v1alpha1.InteractiveVariable) map[string]bool {
+	sensitiveVars := make(map[string]bool)
+	for _, variable := range variables {
+		if variable.Sensitive {
+			sensitiveVars[strings.ToUpper(variable.Name)] = true
+		}
+	}
+	return sensitiveVars
+}
+
+func addZarfVars(pkgVars map[string]overrideData, variables []interface{}, sensitiveVars map[string]bool) []interface{} {
 	// Built-in sensitive variables that should be sanitized
 	sensitiveBuiltInVars := map[string]bool{
 		config.RegistryPushUsername: true,
@@ -534,7 +554,7 @@ func addZarfVars(pkgVars map[string]overrideData, variables []interface{}) []int
 		// "CONFIG" refers to "UDS_CONFIG" which is not a Zarf variable or override so we skip it
 		if key != "CONFIG" {
 			// Mask potentially secret ENV vars or built-in sensitive variables
-			if fv.source == valuesources.Env || sensitiveBuiltInVars[key] {
+			if fv.source == valuesources.Env || sensitiveBuiltInVars[key] || sensitiveVars[key] {
 				fv.value = hiddenVar
 			}
 			variables = append(variables, map[string]interface{}{key: fv.value})
--- a/src/pkg/bundle/deploy.go
+++ b/src/pkg/bundle/deploy.go
@@ -89,6 +89,10 @@ func deployPackages(ctx context.Context, packagesToDeploy []types.Package, b *Bu
 			}
 			b.bundle.Packages[i] = pkg
 		}
+		sha, err := packageManifestDigest(pkg)
+		if err != nil {
+			return err
+		}
 		pkgTmp, err := zarfUtils.MakeTempDir(config.CommonOptions.TempDirectory)
 		if err != nil {
 			return err
@@ -114,8 +118,6 @@ func deployPackages(ctx context.Context, packagesToDeploy []types.Package, b *Bu
 		if err != nil {
 			return err
 		}
-
-		sha := strings.Split(pkg.Ref, "@sha256:")[1] // using appended SHA from create!
 
 		source, err := sources.NewFromLocation(*b.cfg, pkg, pkgTmp, verifyOpts, config.CommonOptions.SkipSignatureValidation, sha, nsOverrides)
 		if err != nil {
@@ -412,9 +414,12 @@ func (b *Bundle) PreDeployValidation() (string, string, string, error) {
 	return bundleName, string(bundleYAML), source, err
 }
 
-// ConfirmBundleDeploy prompts the user to confirm bundle creation
-func (b *Bundle) ConfirmBundleDeploy() (confirm bool) {
-	pkgviews := formPkgViews(b)
+// ConfirmBundleDeploy prompts the user to confirm bundle deployment
+func (b *Bundle) ConfirmBundleDeploy() (bool, error) {
+	pkgViews, err := formPkgViews(b)
+	if err != nil {
+		return false, err
+	}
 
 	message.HeaderInfof("🎁 BUNDLE DEFINITION")
 
@@ -434,7 +439,7 @@ func (b *Bundle) ConfirmBundleDeploy() (confirm bool) {
 
 	message.Title("Packages:", "definition of packages this bundle deploys, including variable overrides")
 
-	for _, pkg := range pkgviews {
+	for _, pkg := range pkgViews {
 		if err := zarfUtils.ColorPrintYAML(pkg.meta, nil, false); err != nil {
 			message.WarnErr(err, "unable to print package metadata yaml")
 		}
@@ -447,17 +452,21 @@ func (b *Bundle) ConfirmBundleDeploy() (confirm bool) {
 
 	// Display prompt if not auto-confirmed
 	if config.CommonOptions.Confirm {
-		return config.CommonOptions.Confirm
+		return config.CommonOptions.Confirm, nil
 	}
 
 	prompt := &survey.Confirm{
 		Message: "Deploy this bundle?",
 	}
 
-	if err := survey.AskOne(prompt, &confirm); err != nil || !confirm {
-		return false
+	var confirm bool
+	if err := survey.AskOne(prompt, &confirm); err != nil {
+		return false, err
 	}
-	return true
+	if !confirm {
+		return false, nil
+	}
+	return true, nil
 }
 
 type PkgView struct {
@@ -466,7 +475,11 @@ type PkgView struct {
 }
 
 // formPkgViews creates a unique pre deploy view of each package's set overrides and Zarf variables
-func formPkgViews(b *Bundle) []PkgView {
+func formPkgViews(b *Bundle) ([]PkgView, error) {
+	return formPkgViewsWithMetadata(b, b.getMetadata)
+}
+
+func formPkgViewsWithMetadata(b *Bundle, getMetadata func(types.Package) (v1alpha1.ZarfPackage, error)) ([]PkgView, error) {
 	var pkgViews []PkgView
 	for _, pkg := range b.bundle.Packages {
 		variables := make([]interface{}, 0)
@@ -476,7 +489,10 @@ func formPkgViews(b *Bundle) []PkgView {
 		valuesOverrides, _, _ := b.loadChartOverrides(pkg, variableData)
 
 		// Load the package metadata to get the list of sensitive variables for masking their values in the view
-		zarfPkg, err := b.getMetadata(pkg)
+		zarfPkg, err := getMetadata(pkg)
+		if err != nil {
+			return nil, fmt.Errorf("unable to load metadata for package %q: %w", pkg.Name, err)
+		}
 		sensitiveVars := sensitiveZarfVariableNames(zarfPkg.Variables)
 
 		for compName, component := range pkg.Overrides {
@@ -498,17 +514,10 @@ func formPkgViews(b *Bundle) []PkgView {
 			}
 		}
 
-		if err != nil {
-			message.WarnErr(err, fmt.Sprintf("unable to load metadata for package %q; masking all Zarf variable values", pkg.Name))
-			for name := range variableData {
-				sensitiveVars[name] = true
-			}
-		}
-
 		variables = addZarfVars(variableData, variables, sensitiveVars)
 		pkgViews = append(pkgViews, PkgView{meta: formPkgMeta(pkg), overrides: map[string]interface{}{"overrides": variables}})
 	}
-	return pkgViews
+	return pkgViews, nil
 }
 
 func formPkgMeta(pkg types.Package) map[string]string {
--- a/src/pkg/bundle/deploy.go
+++ b/src/pkg/bundle/deploy.go
@@ -406,6 +406,8 @@ func (b *Bundle) PreDeployValidation() (string, string, string, error) {
 		return "", "", "", err
 	}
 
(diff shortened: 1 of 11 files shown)


mjnagel:
@greptileai review

src/pkg/bundle/deploy.go
mjnagel: Maybe a separate bug, but seems worth fixing...on name collisions between overrides and sensitive zarf vars this would end up printing out the (potentially) sensitive var during the chart/override printing.

I think if you shift the existing fix above this and pass `sensitiveVars` as a 3rd param here we could cover this case? Would need to update `extractValues` hidden var handling to also read that param, but otherwise relatively straightforward.

src/pkg/bundle/deploy.go
```diff
@@ -492,7 +492,17 @@ func formPkgViews(b *Bundle) []PkgView {
 			}
 		}
 
-		variables = addZarfVars(variableData, variables)
+		// Load the package metadata to get the list of sensitive variables for masking their values in the view
+		zarfPkg, err := b.getMetadata(pkg)
```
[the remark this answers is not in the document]
mjnagel: I believe this is accurate and can borrow some code from the inspect path to resolve? Essentially [this prefetch](https://github.com/defenseunicorns/uds-cli/blob/e480c70849b377f5b8aecd0c71d6eaa4398c60d5/src/pkg/bundle/inspect.go#L83-L92) would ensure metadata is cached?

mjnagel (commented):
Two small comments - functionally this seems to work.

src/pkg/bundle/deploy.go
rodneyxr: So if there happens to be a collision you think we should always mask to be safe?

mjnagel: I think that would make sense? Otherwise you may risk printing a sensitive value (meant for the zarf var) just because it was also used for an override? Don't feel strongly on this one and would welcome other opinions - but think this would err more on the side of caution?

(maybe separate considerations around whether collisions should be allowed but that's definitely not in scope of your changes here 😅)

src/pkg/bundle/deploy.go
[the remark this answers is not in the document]
rodneyxr: I've addressed the issue for local tarballs by reusing the code from `inspect.go` but as the greptile bot pointed out, remote OCI will still fetch per package. Not sure if that is in scope for this issue but I do worry about redundant fetches during deployments for large bundles.

src/pkg/bundle/deploy.go
rodneyxr: I've made updates to mask zarf vars even if a uds-config.yaml marks a colliding name as `sensitive: false`

joelmccoy:
@greptileai review

src/pkg/bundle/deploy.go
```diff
@@ -473,6 +475,10 @@ func formPkgViews(b *Bundle) []PkgView {
 		_, variableData := b.loadVariables(pkg, nil)
 		valuesOverrides, _, _ := b.loadChartOverrides(pkg, variableData)
 
+		// Load the package metadata to get the list of sensitive variables for masking their values in the view
+		zarfPkg, err := b.getMetadata(pkg)
```
joelmccoy: We should include an error guard here. Could we return this error instead of just issuing a warning and continuing? When metadata loading fails, `sensitiveVars` becomes empty while the chart preview is being built. Consequently, a configuration or CLI value that shares a name with a sensitive Zarf variable might be printed. Please return the error from `formPkgViews` and propagate it through `ConfirmBundleDeploy` before rendering the preview. The later masking fallback can then be removed.

```suggestion
		zarfPkg, err := b.getMetadata(pkg)
		if err != nil {
            return nil, fmt.Errorf("unable to load metadata for package %q: %w", pkg.Name, err)
        }
```

src/pkg/bundle/deploy.go
```diff
@@ -484,15 +490,22 @@ func formPkgViews(b *Bundle) []PkgView {
...
 			}
 		}
 
-		variables = addZarfVars(variableData, variables)
+		if err != nil {
+			message.WarnErr(err, fmt.Sprintf("unable to load metadata for package %q; masking all Zarf variable values", pkg.Name))
+			for name := range variableData {
+				sensitiveVars[name] = true
+			}
+		}
```
joelmccoy: Instead of issuing a warning and over-masking, we should return an error earlier here (see earlier comment). If we can't read the package metadata, something is wrong.

src/pkg/bundle/inspect.go
```diff
@@ -258,7 +261,10 @@ func (b *Bundle) getMetadata(pkg types.Package) (v1alpha1.ZarfPackage, error) {
 			return v1alpha1.ZarfPackage{}, fmt.Errorf("package %q: %w", pkg.Name, err)
 		}
 
-		sha := strings.Split(pkg.Ref, "@sha256:")[1] // using appended SHA from create!
+		_, sha, ok := strings.Cut(pkg.Ref, "@sha256:")
```
chance-coleman: If I’m following this correctly, the new missing-digest handling only covers the preview metadata path. The PR description says refs without a digest now return a clear error instead of panicking, but `deployPackages` still indexes `strings.Split(...)[1]` at deploy.go:118. It seems like a malformed or legacy bundle could still panic after confirmation. I think we should either share the safe parser with deployment, or remove this from the PR’s scope and track it separately.

joelmccoy:
@greptileai review

src/pkg/bundle/deploy.go
joelmccoy: I have this return errors all the way through now

src/pkg/bundle/deploy.go
joelmccoy: removed this and return errors instead

joelmccoy:
@greptileai review

src/pkg/bundle/inspect.go
joelmccoy: added a helper function here to check missing manifest and also validate it

src/pkg/bundle/tarball_prefetch.go
```diff
@@ -160,14 +160,29 @@ func (tp *tarballBundleProvider) prefetchPackageMetadata(ctx context.Context, pa
...
+		defer srcFile.Close()
+		return config.BundleArchiveFormat.Extract(ctx, srcFile, handler)
 	}
-	defer srcFile.Close()
-	if err := config.BundleArchiveFormat.Extract(ctx, srcFile, handler); err != nil {
+
+	if err := runPass(); err != nil {
 		return nil, err
 	}
+	// Pulled or older bundles may place metadata before its package manifest.
```
joelmccoy, line 175: Note: The deployment preview now requires package metadata to identify sensitive variables. Metadata prefetch reads the compressed tarball sequentially, but older bundles may place metadata before their package manifest. In such cases, the first pass cannot capture it. We retry once when required blobs remain, preserving the one-pass fast path while supporting valid bundles regardless of tar entry order.

We should improve our pull behavior to ensure it pulls in an ordered format. I believe this issue: https://linear.app/defense-unicorns/issue/CLI-238/uds-inspect-v034x-can-fail-on-valid-pre-v034-bundles-due-to-tar-entry is the same bug.  This addition should fix that issue too

joelmccoy: Also created this issue to make sure pull preserves the expected format: https://linear.app/defense-unicorns/issue/CLI-262/make-uds-pull-preserve-optimized-bundle-tar-entry-ordering

src/pkg/bundle/tarball_prefetch.go
```diff
@@ -160,14 +160,29 @@ func (tp *tarballBundleProvider) prefetchPackageMetadata(ctx context.Context, pa
 		return nil
 	}
 
-	srcFile, err := os.Open(tp.src)
-	if err != nil {
-		return nil, err
+	runPass := func() error {
```
slaskawi, line 163: nit: since this is used two times, maybe it's worth to promote it to a standalone function?

src/pkg/bundle/deploy_test.go
```diff
@@ -682,13 +677,85 @@ func TestFormPkgViews(t *testing.T) {
...
+	}
+
+	for _, source := range []valuesources.Source{valuesources.Config, valuesources.CLI} {
+		t.Run(string(source), func(t *testing.T) {
+			variableData := map[string]overrideData{
+				"SENSITIVE_VAR": {value: "sensitive-value", source: source},
+				"PUBLIC_VAR":    {value: "public-value", source: source},
+			}
+
+			variables := addZarfVars(variableData, nil, sensitiveZarfVariableNames(zarfVariables))
```
slaskawi: If we are in hurry, please disregard this comment but I think it would be fantastic to write a dedicated "table-driven" test for `addZarfVars`. We should test not only replacing public vars with sensitive ones but also other cases like checking "sensitiveBuiltInVars".

slaskawi (dismissed):
Added two minor comments and approved.

src/pkg/bundle/tarball_prefetch.go
joelmccoy: Keeping this local for now since it’s only used twice and closes over the surrounding state. Extracting it would add parameters without improving reuse or clarity.

src/pkg/bundle/deploy_test.go
joelmccoy: updated this

src/pkg/bundle/tarball_prefetch.go
mjnagel: I think this is relatively in line with idiomatic go for an anonymous func like this given the usage internal to this func and the amount of params? But certainly not an expert here...

Edit: slow on my response, but plus one to Joel!

mjnagel (approved):

src/pkg/bundle/deploy.go
```diff
@@ -473,6 +488,13 @@ func formPkgViews(b *Bundle) []PkgView {
 		_, variableData := b.loadVariables(pkg, nil)
 		valuesOverrides, _, _ := b.loadChartOverrides(pkg, variableData)
 
+		// Load the package metadata to get the list of sensitive variables for masking their values in the view
+		zarfPkg, err := getMetadata(pkg)
```
chance-coleman, line 492: remote `uds dev deploy --ref` builds the preview from the package ref in the bundle. If the replacement package marks a variable sensitive but the original does not, the config or CLI value could be printed in clear text.

chance-coleman (approved):

src/pkg/bundle/deploy.go
chance-coleman: closing in favor of a follow on issue

PR #1430, opened 2026-07-30
feat: introduce nfpm release artifacts

wcrum:
## Description

This PR adds an nfpm.yaml and a new series of release artifacts. Specifically variants for amd64/arm64 and RPMs and Debs.

## Related Issue

None.

## Type of change

## Checklist before merging


wcrum:
Should we release nightly?

mjnagel:
@codex review

.github/workflows/release.yaml
```diff
@@ -80,3 +80,58 @@ jobs:
...
+        arch: [amd64, arm64]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Setup golang
+        uses: ./.github/actions/golang
+
+      - name: Install nFPM
+        run: go install github.com/goreleaser/nfpm/v2/cmd/nfpm@v2.47.0
```
mjnagel: Don't know much about this, but any reason to not use [goreleaser to do this](https://goreleaser.com/customization/package/nfpm/)? This would address the codex comment too I think potentially and be a bit more directly in line with existing release process.

mjnagel: Sidenote - one thing that also jumped to mind here was immutable releases which we don't have enabled, but the way this attaches artifacts would be incompatible with _if_ we enabled that.

.goreleaser.yaml
```diff
@@ -34,6 +34,24 @@ builds:
...
+    vendor: Defense Unicorns
+    homepage: https://github.com/defenseunicorns/uds-cli
+    maintainer: Defense Unicorns <support@defenseunicorns.com>
+    description: CLI for creating, publishing, and deploying UDS bundles.
+    license: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+    formats:
+      - deb
+      - rpm
+    bindir: /usr/bin
+    section: utils
```
wcrum, line 51: Section is the reference for debian policy.

https://www.debian.org/doc/debian-policy/ch-archive.html#sections

joelmccoy:
@greptileai review

.goreleaser.yaml
```diff
@@ -34,6 +34,24 @@ builds:
       - arm64
     binary: uds
 
+# Build native Linux packages from the release binaries
+nfpms:
+  - id: linux-packages
+    package_name: uds-cli
+    file_name_template: "{{ .PackageName }}-v{{ .Version }}-{{ .Arch }}"
```
joelmccoy, line 41: Is there a reason why we don't use the conventional filename template here? I don't think we cover OS here. see: https://goreleaser.com/customization/package/nfpm/
```suggestion
    # Default: '{{ .PackageName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ with .Arm }}v{{ . }}{{ end }}{{ with .Mips }}_{{ . }}{{ end }}{{ if not (eq .Amd64 "v1") }}{{ .Amd64 }}{{ end }}'.
    file_name_template: "{{ .ConventionalFileName }}"
```

wcrum: `"{{ .PackageName }}-v{{ .Version }}-{{ .Arch }}"` renders as `myapp-v1.2.3-amd64.deb`

While the default, renders as `myapp_1.2.3_linux_amd64.deb`

https://github.com/defenseunicorns/uds-remote-agent/releases/tag/v0.36.0

I am copying the structure of UDS Remote Agent. 

Is there a naming standard that we follow?

joelmccoy: Good context. I don't think we have strong opinions on the naming convention.  Just was curious about why there was a deviation from the default that nfpm goreleaser.  I guess OS doesn't really disambiguate rpm/deb packages so the conevntionalfilename template doesn't make the most sense.  Closing this out.

joelmccoy (approved):

PR #1449, opened 2026-08-06
fix: zarf manifest digest mangling in bundles

Racer159:
## Description

This fixes digest mangling when Zarf packages are included in bundles.

Zarf v0.80.0 included new functionality to help with state tracking of Zarf packages that are deployed to the cluster that is currently in use by fleet for their "baselines" feature.  UDS CLI currently mangles these manifests and causes an incorrect package digest to show in the cluster that is not correlatable to the originating Zarf package.

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging


joelmccoy:
@greptileai review

src/pkg/bundler/pusher/remote.go
```diff
@@ -43,16 +42,14 @@ func NewPkgPusher(pkg types.Package, cfg Config) RemotePusher {
 
 // Push pushes a Zarf pkg to a remote bundle
 func (p *RemotePusher) Push() (ocispec.Descriptor, error) {
-	zarfManifestDesc, err := p.PushManifest()
+	zarfManifestDesc, err := boci.ToOCIRemote(p.cfg.PkgRootManifest, layout.ZarfLayerMediaTypeBlob, p.cfg.RemoteDst.OrasRemote)
```
joelmccoy, line 45: I think ToOCIRemote still unmarshals json and can mandle the manfiest: https://github.com/defenseunicorns/uds-cli/blob/89bd8234f920e33d6eb1b83d586667dff0f69a30/src/pkg/utils/boci/oci.go#L38

Maybe worth changing those helpers to write the raw bytes?

src/pkg/bundler/fetcher/remote.go
```diff
@@ -92,39 +92,29 @@ func (f *remoteFetcher) copyRemotePkgLayers(layersToCopy []ocispec.Descriptor) (
...
-		}
-
-		// save pkg manifest to bundle root manifest
-		pkgManifestDesc.MediaType = layout.ZarfLayerMediaTypeBlob // force media type to Zarf blob
-		f.cfg.BundleRootManifest.Layers = append(f.cfg.BundleRootManifest.Layers, pkgManifestDesc)
-
-		manifestConfigDesc, err := boci.ToOCIStore(f.pkgRootManifest.Config, layout.ZarfConfigMediaType, f.cfg.Store)
+		// no layers to pull but need to store the package manifest because we didn't use oras.Copy()
+		var err error
+		manifestLayerDesc, err = boci.ToOCIStore(f.pkgRootManifest, layout.ZarfLayerMediaTypeBlob, f.cfg.Store)
```
joelmccoy, line 111: I also think to OCIStore marshalls the json here: https://github.com/defenseunicorns/uds-cli/blob/89bd8234f920e33d6eb1b83d586667dff0f69a30/src/pkg/utils/boci/oci.go#L55

I wonder if it is worth just passing the raw bytes into those functions?

src/pkg/bundler/fetcher/remote_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/zarf-dev/zarf/src/pkg/packager/layout"
+	"oras.land/oras-go/v2/content"
+	ocistore "oras.land/oras-go/v2/content/oci"
+)
+
+func TestRemoteFetcherPreservesCachedPackageManifest(t *testing.T) {
+	ctx := context.Background()
+	manifest := oci.Manifest{Manifest: ocispec.Manifest{MediaType: ocispec.MediaTypeImageManifest}}
+	manifestBytes, err := json.Marshal(&manifest)
```
joelmccoy, line 22: don't we not want to marshal? technically could alter the manifest bytes representation?

Racer159: It definitely can and I actually had it that way before moving around bytes (I will look at the specific functions you had to see if it can be simplified) but for Zarf, package manifests are created "virtually" at times https://github.com/zarf-dev/zarf/blob/36e2779931059765de191ff68c78c25d37a25dd8/src/pkg/packager/layout/oci.go#L85 - so _technically_ it is already relying (and has tests around) the stability of marshaling.

joelmccoy: hmmmm but we have no guarantees that the way zarf marshals is the same that we marshal? right now for the remote stuff we are not calling zarf functions.... and technically could cause a delta. should we be writing the raw bytes here?

Racer159: For more information the main thing that could break it is using a different type (than oci.Manifest) to marshal into since that could reorder keys in the structs - `json.Marshal` (what ORAS uses internally as well) in go will sort things like map keys to produce the same bytes for the same struct input.

In this case the struct ordering and naming comes from the ocispec directly and while that _is_ an external library we also wrap it and could fix it if it diverged. https://github.com/defenseunicorns/pkg/blob/b143f68c76250fd2e7ed09348b6d828da95a4468/oci/manifest.go#L17

We also have tests here and in Zarf that would signal a change like that breaking things.

joelmccoy: ahhh thanks for adding the context there. I think this and my other comments can be dismissed then.

src/pkg/bundler/fetcher/local_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/zarf-dev/zarf/src/pkg/packager"
+	"github.com/zarf-dev/zarf/src/pkg/packager/filters"
+	"github.com/zarf-dev/zarf/src/pkg/packager/layout"
+	"oras.land/oras-go/v2/content"
+	ocistore "oras.land/oras-go/v2/content/oci"
+)
+
+func TestLocalFetcherPreservesZarfManifest(t *testing.T) {
+	ctx := context.Background()
+	packagePath, err := filepath.Abs(filepath.Join("..", "..", "..", "test", "packages", "no-cluster", "real-simple", "zarf-package-real-simple-amd64-0.0.1.tar.zst"))
```
joelmccoy: i don't think this is created anywhere? fails locally for me

src/pkg/sources/common_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/opencontainers/go-digest"
+	"github.com/stretchr/testify/require"
+	"github.com/zarf-dev/zarf/src/pkg/packager/filters"
+	"github.com/zarf-dev/zarf/src/pkg/packager/layout"
+)
+
+func TestLoadPackageFromDirPreservesManifestDigest(t *testing.T) {
+	ctx := context.Background()
+	packagePath, err := filepath.Abs(filepath.Join("..", "..", "test", "packages", "no-cluster", "real-simple", "zarf-package-real-simple-amd64-0.0.1.tar.zst"))
```
joelmccoy: this file is not present and failing unit tests

joelmccoy (commented):
Functionally looks good to me. Will just want to fix the unit tests and wait for CI to be able to run for this

src/pkg/sources/common_test.go
Racer159: Ah yeah - it is made by e2e tests - it was very tiny so I just committed a version of it to the repo for these unit tests.

src/pkg/bundler/fetcher/local_test.go
Racer159: same

chance-coleman (commented):
Reviewed and agree with @joelmccoy this looks good, i dont have any additional findings. Just wanted to note that i looked at this. Will approve on successful CI 👍🏼

joelmccoy:
@Racer159 it looks like github actions are back to normal. recommend updating this branch to trigger CI

src/pkg/bundler/fetcher/local.go
```diff
@@ -157,129 +153,93 @@ func (f *localFetcher) toBundle() ([]ocispec.Descriptor, string, error) {
...
 	}
 
-	// create a file store in the same tmp dir as the Zarf pkg (used to create descs + layers)
-	src, err := file.New(pkgTmp)
+	rootManifestDesc, err := pkgLayout.Resolve(ctx, pkgLayout.Digest())
 	if err != nil {
-		return nil, pkgTmp, err
+		return nil, err
+	}
+	// TODO: Use pkgLayout.Manifest() once it is available in a released Zarf SDK.
```
daveworth, line 184: I'd love an associated Linear cited here for this `TODO` so it doesn't fall off of the end of the human-memory-queue

joelmccoy: captured here: I am working this right now anyways so this will be a fast follow before we release with the new version of zarf: https://linear.app/defense-unicorns/issue/CLI-277/refactor-delegate-zarf-package-oci-layer-selection-to-zoci

chance-coleman (approved):

PR #1450, opened 2026-08-06
add docs for uds zarf

bradclawsie:
## Description

https://docs.defenseunicorns.com/cli/reference/commands/uds/ does not have an entry for `uds zarf`.

I figured this was not intentional, just an oversight. I added a new page for the subcommand with links out to the official zarf documentation for full reference.

If the lack of documentation for this subcommand was intentional, I'll just close this.

This was tested by running `uds run -f tasks.yaml dev-docs` and inspecting the pages. 

Additionally,  `uds run -f tasks.yaml uds-docs-validate` was also completed.
...

## Type of change

## Checklist before merging

### DIFF
--- a/docs/reference/commands/uds.md
+++ b/docs/reference/commands/uds.md
@@ -41,4 +41,5 @@ uds COMMAND [flags]
 * [uds remove](/reference/commands/uds_remove/)	 - Remove a bundle that has been deployed already
 * [uds run](/reference/commands/uds_run/)	 - Run a task using maru-runner
 * [uds version](/reference/commands/uds_version/)	 - Shows the version of the running UDS-CLI binary
+* [uds zarf](/reference/commands/uds_zarf) - Run a zarf command
 
--- a/docs/reference/commands/uds_zarf.md
+++ b/docs/reference/commands/uds_zarf.md
@@ -1,0 +1,43 @@
+---
+title: uds zarf
+description: UDS CLI command reference for <code>uds zarf</code>.
+---
+## uds zarf
+
+Run a zarf command
+
+### Synopsis
+
+Run a zarf command. See [zarf command documentation](https://docs.zarf.dev/commands/zarf/)
+
+```
+uds zarf COMMAND [flags]
+```
+
+### Options
+
+```
+  -h, --help   help for zarf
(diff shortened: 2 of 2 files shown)


docs/reference/commands/uds_zarf.md
joelmccoy, line 1: we auto generate these docs. I think we need to remove this. zarf is just a pass through for upstream and we can not generate the command tree with our program? https://github.com/defenseunicorns/uds-cli/blob/bfde6550b696c2978ee077a7ae39a570dc2e9758/src/cmd/internal.go#L57

docs/reference/commands/uds.md
```diff
@@ -41,4 +41,5 @@ uds COMMAND [flags]
 * [uds remove](/reference/commands/uds_remove/)	 - Remove a bundle that has been deployed already
 * [uds run](/reference/commands/uds_run/)	 - Run a task using maru-runner
 * [uds version](/reference/commands/uds_version/)	 - Shows the version of the running UDS-CLI binary
+* [uds zarf](/reference/commands/uds_zarf) - Run a zarf command
```
joelmccoy, line 44: maybe we can just point upstream for this instead of trying to generate in our docs?

joelmccoy (commented):
Appreciate you calling this out.  Left some comments. We have an auto-generated flow to do this and adding this in manually doesn't really make sense IMO.  Also I am historically not a fan of these command docs (they were just inherited).  When we do some more docs work in the future I would probably like to get rid of these all together.

All that being said, you can maybe just update the `uds.md` to point upstream and remove `uds_zarf.md`?

docs/reference/commands/uds.md
bradclawsie: Yes I think that is a better idea, I'll amend and put this back in to _Draft_ so you aren't pestered with updates. Thanks for the feedback @joelmccoy !

docs/reference/commands/uds_zarf.md
bradclawsie: Yup, I didn't understand the workflow when I opened up this branch. Closing this as a gap in my understanding. Sorry to chew up time!

Issue #1456, opened 2026-08-12
Object value substitution corrupts quotes and backslashes


slaskawi:
### Environment

Device and OS: Any
App version: Current main
Kubernetes distro being used: Any because failure occurs before deployment
Other: Helm object or list override containing ${VAR}

### Steps to reproduce

1. Create this fictional UDS bundle:
   ```yaml
   kind: UDSBundle
   metadata:
     name: example-package
     version: 0.0.1
   packages:
     - name: example-package
       path: ./example-package
       ref: 0.0.1
       overrides:
         example-component:
           example-chart:
             values:
               - path: application.settings
                 value:
                   notifications:
                     enabled: true
                   externalService:
                     endpoint: https://example.invalid
                     credential: ${EXTERNAL_SERVICE_CREDENTIAL}
   ```
2. Run the same preview or deployment command with each value:
   ```sh
   UDS_EXTERNAL_SERVICE_CREDENTIAL='example"credential' uds deploy <bundle> --confirm
   UDS_EXTERNAL_SERVICE_CREDENTIAL='example\q' uds deploy <bundle> --confirm
   UDS_EXTERNAL_SERVICE_CREDENTIAL='example\1' uds deploy <bundle> --confirm
   ```

### Expected result

The complete substituted value is preserved, safely JSON encoded, and passed to Helm without affecting sibling values.

### Actual Result

Object and list values are JSON encoded before substitution. Quotes and backslashes can invalidate the JSON, the whole object fails, preview may omit confirmation because the parsing error is ignored, and the error can include the resolved value.

### Visual Proof (screenshots, videos, text, etc)

The reproduction above should be captured as an automated regression test.

### Severity/Priority

High. Valid credential characters can block deployment, preview can mislead, and sensitive values may enter logs.

### Additional Context

Root cause:

Object and list values are encoded before recursive variable substitution, so substituted quotes and backslashes are interpreted as JSON syntax after encoding.

Proposed implementation:

1. Recursively substitute templated string leaves in maps and lists before JSON encoding.
2. Preserve existing scalar behavior and variable precedence.
3. Propagate preview parsing errors and redact resolved values from all errors.

Documentation clarification:

1. Values are static creation time configuration.
2. Deploy time configuration must use variables.
3. Sensitive runtime values must use `sensitive: true`.
4. Exact literal preservation, including backslashes, should use a file variable with `type: file`.

Fictional migration example:

```yaml
kind: UDSBundle
metadata:
[...]

PR #1457, opened 2026-08-12
docs: clarify bundle values and variables

slaskawi:
## Description

Clarify the intended bundle override model:

* Use `values` for static configuration set at bundle creation.
* Use `variables` for configuration supplied or changed at deploy time.
* Mark credentials with `sensitive: true`.
* Use `type: file` when exact literal contents, including backslashes, must be preserved.

Includes a migration example. No code changes.

## Related Issue

Related to defenseunicorns/uds-cli#1456

## Validation

* `uds run docs:test`
* `git diff --check origin/main...HEAD`


docs/how-to-guides/use-bundle-overrides.mdx
```diff
@@ -64,6 +64,8 @@ overrides:
 
    The `path` uses dot notation to locate the value in the chart's `values.yaml`.
 
+   Use `values` for static configuration. Template expressions in `values` remain supported for compatibility and bundle imports, but use `variables` for new configuration that is supplied at deploy time.
```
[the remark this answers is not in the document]
slaskawi: Clarified this in the guide: [`values` templates remain compatibility support, while new deploy-time configuration should use `variables`](https://github.com/defenseunicorns/uds-cli/blob/agent/object-override-substitution/docs/how-to-guides/use-bundle-overrides.mdx#L67).

src/pkg/bundle/overrides.go
```diff
@@ -140,7 +152,8 @@ func convertOverridesMap(overrideMap map[string]map[string]*values.Options) (pac
 			// Merge the chart values with Helm
 			data, err := chart.MergeValues(getter.Providers{})
 			if err != nil {
-				return nil, err
+				// Do not wrap the Helm error because it may include resolved sensitive values.
+				return nil, fmt.Errorf("unable to process Helm values overrides for component %q chart %q", componentName, chartName)
 			}
```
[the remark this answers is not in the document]
slaskawi: The Helm parser uses a generic wrapper type for these failures, so exposing its type would not identify the bad override. The existing component and chart context keeps the error actionable without risking resolved credential exposure.

src/pkg/bundle/overrides_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+import (
+	"testing"
+
+	"github.com/defenseunicorns/uds-cli/src/types"
+	"github.com/stretchr/testify/require"
+	"github.com/zarf-dev/zarf/src/api/v1alpha1"
+)
+
+func TestLoadChartOverridesPreservesTicketCredentialValues(t *testing.T) {
```
chance-coleman: Was this func name a typo? `TestLoadChartOverridesPreservesTicketCredentialValues` wondering if that was supposed to be `TestLoadChartOverridesPreservesTemplateCredentialValues`

src/pkg/bundle/overrides_test.go
```diff
@@ -0,0 +1,151 @@
...
+	}}}
+
+	_, err := formPkgViewsWithMetadata(&bundle, func(types.Package) (v1alpha1.ZarfPackage, error) {
+		return v1alpha1.ZarfPackage{}, nil
+	})
+
+	require.NoError(t, err)
+}
+
+func TestFormPkgViewsPropagatesTicketOverrideErrorWithoutCredential(t *testing.T) {
```
chance-coleman: similar typo here maybe?

src/pkg/bundle/overrides_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+		_, _, err := (&Bundle{}).loadChartOverrides(pkg, bOverridesData{"KEY": {value: "duplicate"}})
+
+		require.ErrorContains(t, err, "templated map keys resolve to the same value")
+		require.ErrorContains(t, err, `package "example package" component "component" chart "chart"`)
+		require.ErrorContains(t, err, `path "object"`)
+	})
+}
+
+func TestFormPkgViewsLeavesUnavailableImportedTemplatedKeysUnresolved(t *testing.T) {
```
chance-coleman: Could this test assert the generated override instead of only checking `NoError`? Before this change, missing imports were stored with `nil` values, causing both `${FIRST}` and `${SECOND}` to resolve to `<nil>`.

src/pkg/bundle/overrides.go
```diff
@@ -48,8 +49,19 @@ func (b *Bundle) loadVariables(pkg types.Package, bundleExportedVars map[string]
...
 	// imported vars
 	for _, imp := range pkg.Imports {
-		pkgVars[strings.ToUpper(imp.Name)] = bundleExportedVars[imp.Package][imp.Name]
-		overVarsData[strings.ToUpper(imp.Name)] = overrideData{bundleExportedVars[imp.Package][imp.Name], valuesources.Bundle}
+		exportedVars, found := bundleExportedVars[imp.Package]
+		if !found {
+			continue
+		}
+		value, found := exportedVars[imp.Name]
+		if !found {
```
chance-coleman: nit: i think there is a test gap here. i believe the existing test uses lowercase keys which succeeds and never exercises this fallback.

chance-coleman (commented):
few small comments, generally looking good

src/pkg/bundle/overrides_test.go
slaskawi: Yes, that name was unclear. We removed this test when narrowing the PR to documentation and one verified variable migration test.

src/pkg/bundle/overrides_test.go
slaskawi: Removed with the production logic changes. The remaining test is `TestBundleOverrideVariableMigration`.

src/pkg/bundle/overrides_test.go
slaskawi: Agreed. We removed the missing import behavior change entirely. The remaining migration test asserts the generated override for both raw and file variables.

src/pkg/bundle/overrides.go
slaskawi: Agreed, but this fallback was part of the production logic change. We removed it and kept current behavior unchanged, per the agreed documentation only scope.

bradclawsie (approved):

chance-coleman (approved):

joelmccoy (approved):

PR #1458, opened 2026-08-13
refactor(CLI-269)!: prepare repository for CLI modes

slaskawi:
## Description

Prepare the repository for complete Legacy and Next command trees without changing the default CLI behavior.

This is a breaking change for Go library consumers. Legacy public packages move from the `src` namespace to the `pkg/legacy` namespace. Customers importing the public Go library must update their imports before upgrading.

The migration is mechanical. Package names and exported symbols remain unchanged, so call sites do not need to change.

```go
// Before
import "github.com/defenseunicorns/uds-cli/src/config"
import "github.com/defenseunicorns/uds-cli/src/pkg/bundle"
import "github.com/defenseunicorns/uds-cli/src/types"

// After
import "github.com/defenseunicorns/uds-cli/pkg/legacy/config"
import "github.com/defenseunicorns/uds-cli/pkg/legacy/bundle"
import "github.com/defenseunicorns/uds-cli/pkg/legacy/types"
```

Additional examples include:

```text
github.com/defenseunicorns/uds-cli/src/pkg/utils
  -> github.com/defenseunicorns/uds-cli/pkg/legacy/utils

github.com/defenseunicorns/uds-cli/src/config/lang
  -> github.com/defenseunicorns/uds-cli/pkg/legacy/config/lang
```

CLI invocations continue to use the Legacy command tree by default. Next commands are not copied or exposed by this change.

The PR title and commits follow [Conventional Commits 1.0.0](https://www.conventionalcommits.org/en/v1.0.0/), with the breaking change marked using `!`.

## Type of change

- [ ] Bug fix (non breaking change which fixes an issue)
- [ ] New feature (non breaking change which adds functionality)
- [x] Other, repository structure and build preparation, breaking library import migration

## Validation

- Repository lint and secret scanning
- Unit, router, architecture, and Legacy API compatibility tests
- Schema and command documentation generation
- Native and cross platform builds
- GoReleaser configuration and snapshot release build
- Local allowlisted Next import proof, kept outside this branch

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
refactor!: prepare repository for CLI modes

BREAKING CHANGE: Legacy public Go packages moved from `github.com/defenseunicorns/uds-cli/src/...` to `github.com/defenseunicorns/uds-cli/pkg/legacy/...`. Customers must update their imports before upgrading.
END_COMMIT_OVERRIDE


slaskawi:
@greptileai review

cmd/uds/main.go
```diff
@@ -0,0 +1,56 @@
...
+// Package main is the entrypoint for the uds binary.
+package main
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
+	legacycli "github.com/defenseunicorns/uds-cli/internal/legacy/cli"
+	"github.com/defenseunicorns/uds-cli/internal/mode"
```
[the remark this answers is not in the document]
slaskawi: Verified against Go 1.26 initialization rules and the current graph: internal/mode is initialized before Zarf src/cmd, so mode.init strips bootstrap arguments before Zarf reads os.Args. No code change is needed for this thread.

internal/mode/mode_test.go
```diff
@@ -0,0 +1,98 @@
...
+		{"Zarf feature", []string{"zarf", "--features=values=false", "version"}, "", Legacy, "NextMode=false,values=false", []string{"zarf", "--features=values=false", "version"}},
+		{"double dash", []string{"create", "--", "--features=NextMode=true"}, "", Legacy, "NextMode=false", []string{"create", "--", "--features=NextMode=true"}},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			lookup := func(string) (string, bool) { return tt.env, tt.env != "" }
+			got, features, args, err := Resolve(tt.args, lookup)
+			if err != nil || got != tt.mode || features.String() != tt.want || !slices.Equal(args, tt.wantArgs) {
+				t.Fatalf("Resolve() = %q, %q, %q, %v", got, features.String(), args, err)
+			}
```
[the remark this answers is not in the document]
slaskawi: Addressed with the requested formatting cleanup. The Go printf analyzer accepts %q for []string, so this was not a correctness issue, but the assertion now uses %v for clearer diagnostics.

internal/legacy/zarfexec/zarfexec_test.go
```diff
@@ -0,0 +1,32 @@
...
+	originalArgs := append([]string(nil), os.Args...)
+	wantArgs := []string{"zarf", "version"}
+	wantError := errors.New("test error")
+
+	err := execute(context.Background(), []string{"version"}, func(context.Context) error {
+		if !slices.Equal(os.Args, wantArgs) {
+			t.Fatalf("os.Args = %q, want %q", os.Args, wantArgs)
+		}
+		return wantError
+	})
```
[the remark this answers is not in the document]
slaskawi: Verified as a false positive. Go accepts %q for []string and the printf analyzer produces no diagnostic. This moved file does not require a code change.

.github/workflows/nightly-release.yaml
```diff
@@ -47,12 +47,13 @@ jobs:
...
           echo ARCHIVES_VERSION=$(go list -f '{{.Version}}' -m github.com/mholt/archives) >> $GITHUB_ENV
           echo HELM_VERSION=$(go list -f '{{.Version}}' -m helm.sh/helm/v4) >> $GITHUB_ENV
+          echo ZARF_VERSION=$(go list -f '{{.Version}}' -m github.com/zarf-dev/zarf) >> $GITHUB_ENV
 
       - name: Build release artifacts
         uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94 # v7.2.3
         with:
           distribution: goreleaser
-          version: latest
+          version: v2.15.3
```
slaskawi: We need a renovate tag on this and upgrade to the latest version supported as of today

.github/workflows/test.yaml
```diff
@@ -50,11 +50,11 @@ jobs:
       - name: Setup development tools
         uses: ./.github/actions/mise
 
-      - name: Run unit tests
+      - name: Run preparation tests
```
slaskawi: Let's call it "run tests"

.github/workflows/test.yaml
```diff
@@ -91,7 +91,11 @@ jobs:
 
       - name: Fetch zarf@main
         if: inputs.fetch_zarf
-        run: go get github.com/zarf-dev/zarf@main
+        run: |
+          go get github.com/zarf-dev/zarf@main
```
slaskawi, line 95: We need Renovate here as well.

docs/adr/0010-legacy-next-coexistence.md
```diff
@@ -0,0 +1,78 @@
...
+
+UDS CLI Legacy and UDS CLI Next must ship from one Go module and one `uds` binary. This change prepares the repository without copying or exposing Next commands. The future Next integration remains blocked until this preparation is merged.
+
+Legacy is the Alpha default. Its command implementation is private under `internal/legacy`, and its exported library is public under `pkg/legacy`. The canonical root `internal` and `pkg` paths are reserved for Next. Neutral composition packages remain at the root.
+
+## Decision
+
+Use this target structure:
+
+```text
```
slaskawi: We just need a small table that shows what moves where. Something similar to this:

```
main.go                    → cmd/uds/main.go
src/cmd/**                 → internal/legacy/cli/**
src/config/**              → pkg/legacy/config/**
src/types/**               → pkg/legacy/types/**
src/pkg/**                 → pkg/legacy/**
src/test/e2e/**                        → tests/legacy/e2e/**
src/test/common.go                     → internal/legacy/testutil/testutil.go
src/test/{bundles,packages,tasks}/**   → testdata/legacy/{same area}/**
uds-cli-next/internal/**   → internal/**
uds-cli-next/pkg/**        → pkg/**
uds-cli-next/tests/**      → tests/**
```

docs/adr/0010-legacy-next-coexistence.md
```diff
@@ -0,0 +1,78 @@
...
+internal/legacy/testutil
+internal/{artifact,bundle,cli,logger,oci,printer,testutil,zarf}
+pkg/legacy/{config,types,bundle,bundler,cache,engine,interactive,message,sources,style,utils}
+pkg/{bundle,iostreams}
+tests/legacy/e2e
+tests/{integration,cluster,library,smoke,test_data,testutil}
+testdata/legacy/{bundles,packages,tasks}
+```
+
+The canonical Next package is `internal/bundle`. This reflects the final Next source even though the earlier design used `internal/bundlehcl`.
```
slaskawi: Remove all the specific information below.

docs/adr/0010-legacy-next-coexistence.md
```diff
@@ -0,0 +1,78 @@
...
+
+Do not copy the Next entrypoint, documentation, vendor tree, module files, tasks, workflows, release files, lint configuration, Renovate configuration, policy files, or other repository control files. Rewrite the Next module prefix to `github.com/defenseunicorns/uds-cli`. Adapt its composition into `cmd/uds`.
+
+The root lint configuration retains the existing Legacy linters. The pinned Next source is validated with its canonical lint set, so `gosec`, `perfsprint`, and `revive` are excluded only for the allowlisted Next import roots. New composition code remains covered by the complete root lint set.
+
+The preparation owns `internal/mode` and the single `internal/version` metadata seam. The integration must merge, not overwrite, these files. Legacy cannot import canonical Next implementation packages. The integration may compose Legacy command constructors at the root.
+
+Use the existing module with Go 1.26.5. The direct dependency versions shared by both repositories already agree. Helm v3 and Helm v4 coexist through distinct major module paths. Merge Next requirements, run one root `go mod tidy`, and keep one dependency, vulnerability, license, and release graph.
```
slaskawi: Remove this.

.github/workflows/release.yaml
```diff
@@ -71,7 +72,7 @@ jobs:
         uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94 # v7.2.3
         with:
           distribution: goreleaser
-          version: latest
+          version: v2.15.3
```
slaskawi: Add Renovate tag and pin to the latest version available as of today

docs/adr/0010-legacy-next-coexistence.md
slaskawi: The table the above is the most important information from this ADR. Trim down all the rest.

hack/check-architecture.sh
```diff
@@ -0,0 +1,22 @@
...
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+set -eu
+
+for path in internal/legacy pkg/legacy; do
+	if [ ! -d "$path" ]; then
+		echo "Architecture path does not exist: $path" >&2
+		exit 1
+	fi
+	if matches=$(grep -R -n -E 'github\.com/defenseunicorns/uds-cli/internal/(artifact|bundle|cli|logger|oci|printer|testutil|version|zarf)(/|\")' "$path"); then
```
slaskawi, line 12: Ensure this check is invoked as part of the CI

mise.toml
```diff
@@ -5,6 +5,7 @@
 go = "1.26.5"
 hk = "1.54.0"
 golangci-lint = "2.12.2"
+"github:goreleaser/goreleaser" = "2.15.3"
```
slaskawi: Pin to the latest Go releaser version

slaskawi:
Confirmed. The initialization dependency is explicit: the Legacy CLI imports mode before its Zarf dependency graph initializes.

.github/workflows/nightly-release.yaml
```diff
@@ -47,12 +47,13 @@ jobs:
...
           echo ARCHIVES_VERSION=$(go list -f '{{.Version}}' -m github.com/mholt/archives) >> $GITHUB_ENV
           echo HELM_VERSION=$(go list -f '{{.Version}}' -m helm.sh/helm/v4) >> $GITHUB_ENV
+          echo ZARF_VERSION=$(go list -f '{{.Version}}' -m github.com/zarf-dev/zarf) >> $GITHUB_ENV
 
       - name: Build release artifacts
         uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94 # v7.2.3
         with:
           distribution: goreleaser
-          version: latest
+          version: v2.17.1
```
joelmccoy, line 57: can we add a renovate tag for this?

.github/workflows/release.yaml
```diff
@@ -71,7 +72,7 @@ jobs:
         uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94 # v7.2.3
         with:
           distribution: goreleaser
-          version: latest
+          version: v2.17.1
```
joelmccoy, line 76: can we add a renovate tag for this?

hack/check-architecture.sh
joelmccoy, line 1: I am confused, what is the intention with this hack script? we run it as a test?

.github/workflows/test.yaml
```diff
@@ -50,11 +50,11 @@ jobs:
       - name: Setup development tools
         uses: ./.github/actions/mise
 
-      - name: Run unit tests
+      - name: Run tests
         shell: bash
         run: |
           set -o pipefail
-          uds run test:unit 2>&1 | tee "${RUNNER_TEMP}/uds-command-unit.log"
+          uds run test:all 2>&1 | tee "${RUNNER_TEMP}/uds-command-unit.log"
```
joelmccoy: I'm confused by the name of this task. when i see test:all I think this would include integration/e2e tests too... but that is not the case? is this just unit tests? and if so can we name this back to unit tests??

docs/adr/0010-legacy-next-coexistence.md
```diff
@@ -0,0 +1,35 @@
...
+
+Accepted
+
+## Context
+
+UDS CLI Legacy and UDS CLI Next must coexist in one Go module and one `uds` binary. Legacy is the existing CLI and library surface, while Next introduces the canonical package layout. The repository therefore reserves canonical `internal` and `pkg` paths for Next and namespaces Legacy implementation and library packages under `internal/legacy` and `pkg/legacy`.
+
+## Decision
+
+Use the following durable source to destination mapping:
```
joelmccoy: I'm not sure what the actual decision is here "Use the following durable source to destination mapping" ? do we need an ADR? this just seems to document the moved code in this PR. 

We could probably update this to capture the decision to move CLI next code into certain directories for code separation and eventual removal.

And maybe we should call out how we are switching the modes in here?

cmd/uds/main.go
```diff
@@ -0,0 +1,61 @@
...
+		return err
+	}
+	if version.Version != "unset" {
+		legacyconfig.CLIVersion = version.Version
+	}
+	root, err := newRootCommand(selected)
+	if err != nil {
+		return err
+	}
+	root.PersistentFlags().String("features", "", "Features, comma separated name=true or name=false pairs. UDS_FEATURES is also supported.")
```
joelmccoy: I think UDS_FEATURES will have a collision with env varible things we have for CLI.  Anything prefixed with `UDS_` gets translated into variables passed through so using this would prevent someone from sending the `FEATURES` variable through...

Maybe this should have a different prefix so we don't block this? maybe `CLI_FEATURES` so there is no collision?

tasks/tests.yaml
```diff
@@ -1,66 +1,94 @@
...
   - name: unit
     description: run all the unit tests
     actions:
       - cmd: go test ./... -failfast -v -timeout 5m
-        dir: src/pkg
+        dir: pkg/legacy
       - cmd: go test ./... -failfast -v -timeout 5m
-        dir: src/cmd
+        dir: internal/legacy
+      - cmd: go test ./cmd/uds -failfast -v -timeout 5m
```
joelmccoy: nit: could we just run all these with 1 command? or is there a reason we break this out?

bradclawsie (dismissed):
whew a long read, but lgtm

joelmccoy (commented):
Generally lgtm! a couple small comments

docs/adr/0010-legacy-next-coexistence.md
bradclawsie: Do we have a policy on deleting ADRs? This seems worthwhile to keep around during the (expectedly long) time people might expect to see `legacy/`. If we expect ADRs to be evergreen, then I agree, it might be better to delete and document this elsewhere.

.github/workflows/nightly-release.yaml
slaskawi: Added the Renovate directive for [goreleaser/goreleaser](https://github.com/defenseunicorns/uds-cli/blob/6c54602/.github/workflows/nightly-release.yaml#L56).

.github/workflows/release.yaml
slaskawi: Added the shared Renovate directive [here](https://github.com/defenseunicorns/uds-cli/blob/6c54602/.github/workflows/release.yaml#L75); mise.toml does not manage this workflow input.

hack/check-architecture.sh
slaskawi: It is a dependency-boundary test for coexistence: Legacy must remain independent of the canonical Next packages. The Unit Tests CI job runs it in the separate [Check architecture step](https://github.com/defenseunicorns/uds-cli/blob/6c54602/.github/workflows/test.yaml#L59).

.github/workflows/test.yaml
slaskawi: You were right: this is unit coverage plus the architecture boundary check, not the E2E suite. The workflow now runs them as [separate, named steps](https://github.com/defenseunicorns/uds-cli/blob/6c54602/.github/workflows/test.yaml#L53-L60).

cmd/uds/main.go
slaskawi: Agreed. Reserving UDS_FEATURES prevented the existing UDS_ override mechanism from passing FEATURES through. The mode setting is now [CLI_FEATURES](https://github.com/defenseunicorns/uds-cli/blob/6c54602/internal/mode/mode.go#L19), and the reservation is removed.

docs/adr/0010-legacy-next-coexistence.md
slaskawi: I kept this ADR because it records the durable coexistence decision: Legacy is namespaced, Next owns the canonical paths, and NextMode selects the command tree. The [mapping](https://github.com/defenseunicorns/uds-cli/blob/6c54602/docs/adr/0010-legacy-next-coexistence.md#L15-L35) is the implementation record; ADR-retention policy can be decided separately when Legacy removal is planned.

tasks/tests.yaml
slaskawi: Consolidated the same [unit-test roots](https://github.com/defenseunicorns/uds-cli/blob/89499da/tasks/tests.yaml#L5-L8) into one command; E2E remains outside this task.

bradclawsie (dismissed):
checked the two new commits ( https://github.com/defenseunicorns/uds-cli/pull/1458/commits/89499da2aed869208dac1c0dc78234ce4164e5d9 and https://github.com/defenseunicorns/uds-cli/pull/1458/commits/1511d4e246864c83dda14f6b6307bfadb546e7a2 ) since my last approval, these look fine, re-approving

docs/adr/0010-legacy-next-coexistence.md
joelmccoy: That makes sense. Can we state the decision directly before the mapping?

The decision is not about the file move itself. The decision is:

* Legacy code will reside under `internal/legacy` and `pkg/legacy` to keep it isolated for future removal.
* Next code will utilize the standard `internal` and `pkg` paths, ensuring new development does not rely on legacy paths.
* NextMode will determine which command tree runs.

The mapping table can remain as an implementation detail, illustrating how this PR applies that decision.

joelmccoy (dismissed):
LGTM - responded to one of my comments about the wording in the ADR, but really non blocking

docs/adr/0010-legacy-next-coexistence.md
slaskawi: Agreed. I’ll state those three coexistence decisions directly before the mapping and keep the table as this PR’s implementation detail.

bradclawsie (approved):
checked commits since last approved; lgtm!

joelmccoy (approved):

PR #1459, opened 2026-08-14
chore: update go to 1.26.6

bradclawsie:
## Description

Update Go to 1.26.6.

## Type of change

## Checklist before merging

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
 
 [tools]
-go = "1.26.5"
+go = "1.26.6"
 hk = "1.54.0"
 golangci-lint = "2.12.2"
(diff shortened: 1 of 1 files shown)


mise.toml
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
 
 [tools]
-go = "1.26.5"
```
chance-coleman, line 5: should we also be updating the `go.mod` for Go 1.26.6?

chance-coleman, line 5: I was wondering why this was being created manually instead of renovate opening a pr so i looked at renovate logs and it would have opened a PR for this but its waiting for the three day stability window. I probably missed something that determined we needed to update this sooner than renovate would?

bradclawsie: ah no! I didn't know about that! thanks Chance, I'll just close this and let renovate do its thing.

bradclawsie:
apparently renovate takes care of these

PR #1462, opened 2026-08-20
chore: add go mod tidy to hk/lint checks

joelmccoy:
## Description

Adds `go mod tidy` check to hk/lint checks.

## Type of change

## Checklist before merging

### DIFF
--- a/hk.pkl
+++ b/hk.pkl
@@ -36,6 +36,11 @@ local checks = new Mapping<String, Step> {
         fix = "golangci-lint run --fix ./..."
         glob = List("**/*.go", "go.mod", "go.sum")
     }
+    ["go-mod-tidy"] {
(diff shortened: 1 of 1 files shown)


bradclawsie (commented):
seems fine but what is our intent for updating pkgs? I usually do `go mod tidy` to its own PR, done only periodically

joelmccoy:
> seems fine but what is our intent for updating pkgs? I usually do `go mod tidy` to its own PR, done only periodically

renovate already does this in it's prs (runs go mod tidy). however if you add a feature that adds a dep/bumps it we have no lint check as it currently stands to do this (i caught this with my import work). this should be a check and we should never have any non tidy go mod/sum files checked in

bradclawsie (approved):

PR #1460, opened 2026-08-18
chore(deps): update to zarf 0.84.0

AustinAbro321:
## Description

Prepares for the next version of Zarf 0.84.0, which introduces rather significant breaking changes. 

CI failures seem mostly related to using a non-version

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Other (security config, docs update, etc)

## Checklist before merging

- [ ] Test, docs, adr added or updated as needed
- [ ] [Contributor Guide Steps](https://github.com/defenseunicorns/uds-cli/blob/main/CONTRIBUTING.md) followed

PR #1468, opened 2026-08-21
chore(deps): update zarf to v0.84.0

slaskawi:
## Description

Updates Zarf to v0.84.0 and adapts the legacy and next CLI modes to Zarf's package-definition API. Updates next-mode init test fixtures.

## Related Issue

Relates to #1460

## Type of change

## Checklist before merging


go.mod
```diff
@@ -23,8 +23,8 @@ require (
 	github.com/spf13/cobra v1.10.2
 	github.com/spf13/pflag v1.0.10
 	github.com/spf13/viper v1.21.0
-	github.com/stretchr/testify v1.11.1
-	github.com/zarf-dev/zarf v0.83.0
+	github.com/stretchr/testify v1.12.0
+	github.com/zarf-dev/zarf v0.84.0
```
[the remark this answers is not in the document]
slaskawi: Fixed by aligning the schema generator and checked-in Zarf schema with v0.84.0: https://github.com/defenseunicorns/uds-cli/blob/31e4220/hack/generate-schema.sh#L36

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/defenseunicorns/uds-cli
 
-go 1.26.5
+go 1.26.6
```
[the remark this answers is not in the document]
slaskawi: Fixed the local toolchain mismatch in [mise.toml](https://github.com/defenseunicorns/uds-cli/blob/31e4220/mise.toml#L5).

slaskawi:
@greptileai review

slaskawi:
@codex review

pkg/legacy/bundler/fetcher/common.go
```diff
@@ -19,19 +19,18 @@ import (
 	"github.com/zarf-dev/zarf/src/api/v1alpha1"
 	"github.com/zarf-dev/zarf/src/pkg/packager/layout"
 	"github.com/zarf-dev/zarf/src/pkg/zoci"
+	zarfTypes "github.com/zarf-dev/zarf/src/types"
 )
 
-func NewZarfOCIRemote(ctx context.Context, url string, platform ocispec.Platform, mods ...oci.Modifier) (*zoci.Remote, error) {
+func NewZarfOCIRemote(ctx context.Context, url string, platform ocispec.Platform, _ ...oci.Modifier) (*zoci.Remote, error) {
```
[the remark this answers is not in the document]
slaskawi: Fixed by removing the unused variadic modifier parameter from [NewZarfOCIRemote](https://github.com/defenseunicorns/uds-cli/blob/43c7daf/pkg/legacy/bundler/fetcher/common.go#L24), so callers cannot silently assume modifiers are applied.

briantwatson (commented):
Thinking we might want to also bump these doc locations:

https://github.com/defenseunicorns/uds-cli/blob/c769944375df9d175da0418a652dd1ae6130fc16/docs/getting-started/quickstart.mdx?plain=1#L44
https://github.com/defenseunicorns/uds-cli/blob/c769944375df9d175da0418a652dd1ae6130fc16/docs/how-to-guides/verify-keyless-package-signatures.mdx?plain=1#L68
https://github.com/defenseunicorns/uds-cli/blob/c769944375df9d175da0418a652dd1ae6130fc16/README.md?plain=1#L31

joelmccoy:
@greptileai review

bradclawsie:
@codex review

pkg/bundle/zarf.go
```diff
@@ -136,14 +137,15 @@ func fromZarfPackageLayout(pkgLayout *layout.PackageLayout) *ZarfPackageLayout {
 	if pkgLayout == nil {
 		return nil
 	}
+	zarfPkg := pkgLayout.AsV1alpha1()
```
chance-coleman: `PackageDefinition` is converted to `v1alpha1` and rebuilt, which loses `v1beta1` fields such as `Image.Source: daemon`, multiple imports, and some service metadata.  Is this an expected behavior? This probably affects some pre deploy hooks and other public loaders.

chance-coleman: Also applicable further down in this doc around line #200.

pkg/legacy/bundler/fetcher/common.go
```diff
@@ -11,27 +11,25 @@ import (
...
 	goyaml "github.com/goccy/go-yaml"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/zarf-dev/zarf/src/api/v1alpha1"
 	"github.com/zarf-dev/zarf/src/pkg/packager/layout"
 	"github.com/zarf-dev/zarf/src/pkg/zoci"
+	zarfTypes "github.com/zarf-dev/zarf/src/types"
 )
 
-func NewZarfOCIRemote(ctx context.Context, url string, platform ocispec.Platform, mods ...oci.Modifier) (*zoci.Remote, error) {
+func NewZarfOCIRemote(ctx context.Context, url string, platform ocispec.Platform) (*zoci.Remote, error) {
```
chance-coleman: The old code sent `uds-cli/<version>`. The new Zarf client sends `zarf/<version>`, likely `zarf/unset-development-only` during normal UDS execution. Won't this affect UDS Identity which could have some unfortunate side affects? Also audit logs would change i think

pkg/bundle/zarf_test.go
```diff
@@ -24,6 +25,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func newPackageLayout(pkg v1alpha1.ZarfPackage) *layout.PackageLayout {
```
chance-coleman, line 29: Noticed there is no v1beta1 regression tests for adapters. Existing tests construct only v1alpha1 layouts. Probably worth adding a v1beta1 round-trip test covering image source, service metadata, and multiple imports.

bradclawsie (dismissed):
reviewed, codex reviewed, codex reviewed from local cli

chance-coleman (commented):
first pass comments before i take off.

pkg/bundle/zarf.go
slaskawi: Good catch. This is a real bug: the adapter drops v1beta1-only fields during the `AsV1alpha1()` round-trip before rebuilding the layout ([`zarf.go#L140`](https://github.com/defenseunicorns/uds-cli/blob/chore/update-zarf-0.84.0/pkg/bundle/zarf.go#L140), [`zarf.go#L200`](https://github.com/defenseunicorns/uds-cli/blob/chore/update-zarf-0.84.0/pkg/bundle/zarf.go#L200)). I’ll preserve the original definition and apply only the supported hook mutations, with regression tests for these fields.

pkg/legacy/bundler/fetcher/common.go
slaskawi: Good catch. This is a real compatibility regression: the new path can emit `zarf/unset-development-only` instead of `uds-cli/<version>`. I’ll retain the UDS override and add a header regression test.

pkg/bundle/zarf_test.go
slaskawi: Agree. The adapter currently normalizes layouts through v1alpha1, so beta-only fields can be lost. I’ll add the requested round-trip coverage and preserve the original definition version in [`zarf.go`](https://github.com/defenseunicorns/uds-cli/blob/chore/update-zarf-0.84.0/pkg/bundle/zarf.go#L196).

pkg/bundle/zarf.go
```diff
@@ -20,9 +22,10 @@ import (
 type ZarfPackageLayout struct {
 	Directory string
 	// IsPartial indicates that the layout may omit files referenced by its checksums.
-	IsPartial      bool
-	Pkg            ZarfPackage
-	registryDigest string
+	IsPartial         bool
+	Pkg               ZarfPackage
```
joelmccoy: do we need pkg here anymore then? Upstream now uses `PackageDefinition`, so keeping `Pkg` here feels like we are preserving the old layout model in our public API. Can we refactor this to mirror the new type shape instead: https://github.com/zarf-dev/zarf/blob/97a1b4d0e9ca6095450f87a9c8690d5c76f71702/src/pkg/packager/layout/package.go#L35-L41

go.mod
chance-coleman, line 27: Does this upgrade require updating the legacy YAML readers too? They still decode into v1alpha1, so valid v1beta1 packages can fail or lose required/optional semantics. I'm thinking it might make sense here to use Zarf’s version-aware parser or reject v1beta1 explicitly.

pkg/bundle/zarf.go
```diff
@@ -238,9 +249,47 @@ func applyPublicPackageLayout(dst *layout.PackageLayout, src *ZarfPackageLayout)
 		}
 		components[i] = private
 	}
-	dst.Pkg.Components = components
+	zarfPkg.Components = components
+	if dst.PackageDefinition.OriginalAPIVersion() == v1beta1.APIVersion {
```
joelmccoy: Can we avoid this reconstruction path by having the hook-facing layout carry `api.PackageDefinition` directly?

The v1beta1 preservation logic seems to indicate that the public `Pkg` shape is lossy. We expose images as `[]string`, but v1beta1 images contain additional data, such as `Source`. This requires the adapter to rebuild the actual package definition and manually preserve fields that the public type cannot represent.

Since the upstream `layout.PackageLayout` now stores `PackageDefinition api.PackageDefinition`, can we mirror that here instead of exposing `Pkg ZarfPackage`? This way, hooks can modify the actual package definition shape for the schema they care about, eliminating the need for custom v1alpha1/v1beta1 reconstruction logic in this adapter.

joelmccoy (commented):
Mostly looking good! Noticed one thing we may want to consider here for CLI next

go.mod
slaskawi: Fixed the legacy metadata readers to preserve beta package semantics through the compatibility path: [decoder integration](https://github.com/defenseunicorns/uds-cli/blob/902888b/pkg/legacy/sources/tarball.go#L57).

pkg/bundle/zarf.go
slaskawi: Implemented the schema-aware [PackageDefinition](https://github.com/defenseunicorns/uds-cli/blob/902888b/pkg/bundle/zarf.go#L20) hook surface, removing the lossy reconstruction path.

pkg/bundle/zarf.go
slaskawi: Implemented the native [PackageDefinition](https://github.com/defenseunicorns/uds-cli/blob/902888b/pkg/bundle/zarf.go#L20) hook API and removed the lossy reconstruction path.

bradclawsie (commented):
lgtm but I will await approval from reviewers who brought up the last round of issues.

chance-coleman (dismissed):

pkg/legacy/utils/utils.go
```diff
@@ -27,12 +27,41 @@ import (
...
+	"github.com/zarf-dev/zarf/src/api/v1beta1"
 	"github.com/zarf-dev/zarf/src/pkg/packager"
 	"github.com/zarf-dev/zarf/src/pkg/packager/layout"
 	"github.com/zarf-dev/zarf/src/pkg/signing"
 )
 
+// ReadPackageYAML decodes either supported Zarf package schema and returns
+// the legacy v1alpha1 compatibility view through Zarf's public converter.
+func ReadPackageYAML(path string) (v1alpha1.ZarfPackage, error) {
+	data, err := os.ReadFile(path)
```
joelmccoy, line 41: we should use upstream zarf method `load.PackageDefinition` here instead of having custom conversion logic here.

joelmccoy: Welp, it seems this isn't working smoothly right now because the legacy bundle paths are reading metadata from already-built Zarf package artifacts instead of source package definitions.

I attempted to switch to `upstream load.PackageDefinition`, which is our long-term goal since Zarf should manage alpha/beta schema handling, conversion, imports, validation, defaults, and more. However, `load.PackageDefinition` appears to be designed for source `zarf.yaml` definitions and performs package schema linting. This caused failures with older built package metadata that Zarf still accepts through package artifact/layout paths.

Here are two specific issues I encountered:

1. We have an unmanaged remote package fixture used in E2E tests, `ghcr.io/defenseunicorns/uds-cli/nginx:0.0.2`, which was built manually with older tooling and includes legacy build metadata like:

    ```yaml
      build:
        lastNonBreakingVersion: v0.27.0
    ```

    `load.PackageDefinition` rejects this with:

    ```text
      Additional property lastNonBreakingVersion is not allowed
    ```

    We should either remove this fixture, rebuild/manage it from this repository, or document it intentionally as a backwards-compatibility fixture: https://linear.app/defense-unicorns/issue/CLI-309/replace-unmanaged-remote-zarf-package-test-fixture-in-legacy-bundle.

2. There is a broader API/architecture question here. We should prefer upstream Zarf metadata loading whenever possible, especially now that Zarf manages both `v1alpha1` and `v1beta1` schemas. However, we need to differentiate between source-definition loading and built-artifact metadata loading, using the correct upstream Zarf API for each path. I captured that follow-up here: https://linear.app/defense-unicorns/issue/CLI-308/investigate-upstream-zarf-metadata-loading-in-legacy-bundle-paths.

So, I’m backing out this specific change for now instead of forcing `load.PackageDefinition` into a path where it rejects package artifacts that Zarf can still pull and load.

bradclawsie:
@codex review

slaskawi:
Reopened to hopefully trigger a new CI run. Github Actions don't seem to be in the mood today.

pkg/legacy/utils/boci/oci.go
```diff
@@ -511,13 +511,28 @@ func getFilteredComponents(ctx context.Context, remote *oci.OrasRemote, manifest
...
-		filters.ForDeploy(strings.Join(optionalComponents, ","), false),
-	)
-	filteredComponents, err := createFilter.Apply(zarfPkg)
+	componentViews := make([]filters.ComponentView, 0, len(zarfPkg.Components))
+	for _, component := range zarfPkg.Components {
+		componentViews = append(componentViews, filters.ComponentView{
+			Name:        component.Name,
+			Description: component.Description,
+			Optional:    !component.IsRequired(),
+			Default:     component.Default,
```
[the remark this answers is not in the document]
slaskawi: Fixed by routing both [remote filtering](https://github.com/defenseunicorns/uds-cli/blob/chore/update-zarf-0.84.0/pkg/legacy/utils/boci/oci.go#L498) and [remote loading](https://github.com/defenseunicorns/uds-cli/blob/chore/update-zarf-0.84.0/pkg/legacy/sources/remote.go#L43) through the API-version-aware compatibility decoder.

pkg/bundle/zarf.go
```diff
@@ -178,9 +139,7 @@ func toZarfPackageLayout(ctx context.Context, pkgLayout *ZarfPackageLayout, isPa
 		if err != nil {
 			return nil, err
 		}
-		if err := applyPublicPackageLayout(result, pkgLayout); err != nil {
-			return nil, err
-		}
+		result.PackageDefinition = pkgLayout.PackageDefinition
 		return result, nil
```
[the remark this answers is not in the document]
slaskawi: Fixed by preserving the caller-supplied deployed digest when rebuilding the staged layout in [zarf.go](https://github.com/defenseunicorns/uds-cli/blob/chore/update-zarf-0.84.0/pkg/bundle/zarf.go#L142).

bradclawsie (approved):

joelmccoy (approved):

PR #1469, opened 2026-08-21
feat(next)!: artifact support for remove

eoghanriley:
## Description

Removal now accepts local tarballs and OCI references

Originally from https://github.com/defenseunicorns/uds-cli-next/pull/302

## Related Issue

Fixes CLI-299

## Type of change

## Checklist before merging


internal/zarf/remove.go
```diff
@@ -247,11 +256,12 @@ func (r *ZarfRemover) RemovePackage(ctx context.Context, pkg *spec.Package, opts
 		OCIConcurrency: opts.Config.Options.Concurrency,
 	}
 
-	// Load metadata from pkg.Source to discover the Zarf metadata.name. For OCI
-	// sources, only the metadata layer is pulled (a few KB), not the full package.
-	zarfPkg, err := packager.GetPackageFromSourceOrCluster(ctx, c, pkg.Source, pkg.Namespace, loadOpts)
+	// Artifact-backed removal supplies the deployed Zarf name so this loads
+	// package state from the cluster. Source removal discovers the name from the
+	// author-provided package source; OCI sources fetch only the metadata layer.
+	zarfPkg, err := packager.GetPackageFromSourceOrCluster(ctx, c, packageSource, pkg.Namespace, loadOpts)
```
chance-coleman, line 277: Is this expected? If a package is already gone from the cluster, removal now returns an error instead of marking it as skipped. Could we check whether it is deployed before loading it from the cluster?

pkg/bundle/remove.go
```diff
@@ -65,6 +73,68 @@ func Remove(ctx context.Context, source *DeploySource, opts RemoveOptions) (*Rem
...
+	case artifact.IsTarZst(source.BundlePath):
+		var err error
+		source, err = PrepareDeploySource(
+			ctx,
+			s,
+			source.BundlePath,
+			opts.Config.Options.TmpDir,
+			opts.Config.Options.Architecture,
+		)
+		if err != nil {
```
chance-coleman: Could we avoid using `source.BundlePath` after `PrepareDeploySource` fails? That function can return a nil source, which would cause this error path to panic. The OCI error path has the same issue.

pkg/bundle/remove.go
```diff
@@ -65,6 +73,68 @@ func Remove(ctx context.Context, source *DeploySource, opts RemoveOptions) (*Rem
...
+		}
+		pulled, err := Pull(ctx, source.BundlePath, workspace, PullOptions{
+			Config:                    opts.Config,
+			Verification:              policy,
+			SkipSignatureVerification: opts.SkipSignatureVerification,
+			Streams:                   s,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("%w %q: %w", ErrRemoveBundle, source.BundlePath, err)
+		}
```
chance-coleman, line 97: Should we call `source.Close()` after preparing the pulled artifact? The prepared source creates its own temporary directory, but the current cleanup only removes the pull directory.

pkg/bundle/remove.go
```diff
@@ -65,6 +73,68 @@ func Remove(ctx context.Context, source *DeploySource, opts RemoveOptions) (*Rem
 
 	s := logger.Bind(opts.Streams, opts.Config.Options.LogLevel)
 
+	switch {
```
chance-coleman, line 78: I might be missing something here so definitely double check me.  An OCI tag can validly end in .tar.zst, but this new branch would treat that reference as a local file and fail to pull it. Maybe we should check for an OCI reference before checking for .tar.zst?

chance-coleman (commented):
generally looking good, few first pass comments.

pkg/bundle/remove.go
eoghanriley: fixed

pkg/bundle/remove.go
eoghanriley: fixed

pkg/bundle/remove.go
eoghanriley: Didn't realize IsTarZst is just a string.contains, fixed

eoghanriley:
@greptile-apps

internal/zarf/remove.go
eoghanriley: Fixed and added a test

pkg/bundle/remove.go
chance-coleman: If i'm understanding this correctly, this resolves the library path, but the CLI still calls `ValidateBundlePath` first, where `.tar.zst` is checked before OCI references. Because bundle remove enables both artifact types, `oci://registry/repo:tag.tar.zst` may still be treated as a local file and rejected before reaching this branch. Could we reorder that validation or add a regression test for this case?

eoghanriley: good catch I forgot about that function. Fixed and I added regression tests for Remove and ValidateBundlePath

chance-coleman (approved):

PR #1476, opened 2026-08-26
build: vendor Go dependencies

slaskawi:
## Description

- Add and commit the Go vendor tree.
- Remove the `zarf@main` CI override; builds use the version declared in `go.mod`.
- Renovate manages `github.com/zarf-dev/zarf` through the Go module manager and the existing `zarf` package rule.

## Related Issue

N/A

## Type of change

## Checklist before merging


bradclawsie (approved):
vendoring lgtm

chance-coleman (approved):
lgtm. with one question: Do we need a CI check that regenerates the vendored dependencies and verifies the committed vendor/ tree matches?

bradclawsie:
> lgtm. with one question: Do we need a CI check that regenerates the vendored dependencies and verifies the committed vendor/ tree matches?

I would think it would be enough to see if `go.sum` changed?

slaskawi:
> lgtm. with one question: Do we need a CI check that regenerates the vendored dependencies and verifies the committed vendor/ tree matches?

@chance-coleman This check is not strictly necessary but I believe it's helpful. Go will not error out if you modify some of the files by hand (and just keep `vendor/modules.txt` untouched). I checked locally, at it seems this check costs us ~3–5 minutes (previous run was 149 seconds actually), whereas the E2E matrix takes up to 15 minutes. 

So I think it's worth to have another safeguard in place and ensure we always commit the correct content of the vendor directory. 

> I would think it would be enough to see if go.sum changed?

@bradclawsie Actually no - go.sum tracks the checksums of the downloaded modules. If there were no changes to the go.mod, the go.sum should stay the same.

joelmccoy:
I got to this one late, but we will also want to update renvoate to ignore vendored path and also run go mod vendor on update: https://github.com/defenseunicorns/uds-cli-next/blob/main/renovate.json#L11

https://github.com/defenseunicorns/uds-cli-next/blob/main/renovate.json#L15

PR #1477, opened 2026-08-26
feat(next)!: artifact support for remove

eoghanriley:
## Description

Allows users to call the Remove function on cli command with a tar.zst or oci artifact.

## Related Issue

Fixes CLI-299

## Type of change

## Checklist before merging


internal/artifact/inspect.go
```diff
@@ -225,39 +211,22 @@ func inspectPackageSignature(ctx context.Context, idx ocispec.Index, pkg spec.Pa
...
-		return nil, UnsupportedSchemaVersionError{Artifact: "package manifest", Version: manifest.SchemaVersion}
+		return nil, err
 	}
-	if manifest.MediaType != "" && !udsoci.IsImageManifestMediaType(manifest.MediaType) {
-		return nil, UnsupportedMediaTypeError{Artifact: "package manifest", MediaType: manifest.MediaType}
+	if !found {
+		return summary, nil
 	}
 
-	zarfLayer, ok := findLayerByTitleOptional(manifest, "zarf.yaml")
```
joelmccoy, line 243: nice cleanup here!

internal/cli/bundle/remove.go
```diff
@@ -130,23 +142,50 @@ func (o *RemoveOptions) Complete(cmd *cobra.Command, args []string) error {
...
+		}
+		parsedBundle = result.Bundle
+		digest = result.ArtifactDigest
+	} else {
+		bundlePath := resolveBundlePath(o.BundlePath)
+		parsedBundle, err = bundleinternal.NewHCLParser(o.Config.Options.Architecture, s).ParseBundleFile(ctx, bundlePath)
+		if err != nil {
+			return fmt.Errorf("%w %q: %w", ErrParseBundle, bundlePath, err)
+		}
+	}
```
joelmccoy, line 167: this does too mcuh in the Cobra options validation path. The `Validate()` function now pulls and inspects OCI references or extracts and inspects local artifacts to parse the bundle. However, the `Run()` function later calls `bundle.Remove()`, which prepares the artifact again. This means that the command `uds bundle remove oci://...` can pull and process the same artifact twice, and the `.tar.zst` file can be extracted twice.

Can we restructure this to align with the deploy process instead? Let's keep `Validate()` focused on simple argument and flag checks such as path and reference shape, local file existence, and verification policy. Then, in `Run()`, we can prepare the source once, parse the bundle, validate package selection and dependency safety, prompt the user, and call remove with that same prepared source.

internal/cli/bundle/remove.go
```diff
@@ -130,23 +142,50 @@ func (o *RemoveOptions) Complete(cmd *cobra.Command, args []string) error {
...
-	}
-
 	// Bind a logger so the parse + safety-check diagnostics below honor --log-level
 	// and Streams.ErrOut, consistent with the rest of the CLI.
 	ctx := context.Background()
 	s := logger.Bind(o.IOStreams, o.Config.Options.LogLevel)
-	bundlePath := resolveBundlePath(o.BundlePath)
-	parsedBundle, err := bundleinternal.NewHCLParser(o.Config.Options.Architecture, s).ParseBundleFile(ctx, bundlePath)
+
+	err := ValidateBundlePath(o.BundlePath, AllowArtifactBundlePath(), AllowOCIReferenceBundlePath())
```
joelmccoy: I know this wasn't originally in scope... but I think we should consider it looking at this pr and I would be ok with this breaking change

Could we simplify the input model and align `remove` with `deploy`? Currently, `uds bundle deploy` is artifact-only, while source-based deployment is under `uds bundle dev deploy`. It seems that `uds bundle remove` should follow the same structure: artifact removal here (`.tar.zst` / `oci://...`) and source-based removal under `uds bundle dev remove`.

This change would allow the command to use the same artifact-reference validation pattern as `deploy`, avoid expanding `ValidateBundlePath()` to support every source type, and maintain a consistent user experience.

pkg/bundle/remove.go
```diff
@@ -6,17 +6,26 @@ package bundle
...
 type RemoveOptions struct {
-	Config   *UDSBundleConfig
-	Packages []string
+	Config                    *UDSBundleConfig
+	Packages                  []string
+	Verification              VerificationPolicy
+	SkipSignatureVerification bool
+	// ExpectedArtifactDigest, when set, requires the artifact used for removal
+	// to match the artifact previously inspected by the caller.
+	ExpectedArtifactDigest string
```
joelmccoy: What was the use case you imagined callers needing this for? I would remove this all the way through the chain as I am not seeing the immediate value of checking this?

pkg/bundle/remove.go
```diff
@@ -65,6 +75,97 @@ func Remove(ctx context.Context, source *DeploySource, opts RemoveOptions) (*Rem
...
+		workspace, err := os.MkdirTemp(opts.Config.Options.TmpDir, "uds-bundle-remove-*")
+		if err != nil {
+			return nil, fmt.Errorf("%w %q: creating workspace: %w", ErrRemoveBundle, source.BundlePath, err)
+		}
+		defer func() { _ = os.RemoveAll(workspace) }()
+		policy := opts.Verification
+		if !policy.configured() && opts.Config.SignatureVerification != nil {
+			policy = *opts.Config.SignatureVerification
+		}
+		pulled, err := Pull(ctx, source.BundlePath, workspace, PullOptions{
```
joelmccoy: nit: it would be way cooler/efficient if we could do the remove by reading the minimal amount of OCI information instead of pulling down the whole artifact. don't need to do for initial implementation, but maybe capture a ticket for this if you don't want to do in this pr

joelmccoy: We would need to make `func readPackageZarfName` storage agnostic in `package_metadata.go`. We can continue to develop the artifact metadata extraction in this direction, ensuring it is independent of where the OCI content resides. This way, local artifact extraction and remote `oci://` removal can utilize the same package-name reader. Additionally, a remote removal path can fetch only the necessary package manifests and `zarf.yaml` blobs instead of pulling the entire artifact.

joelmccoy (commented):
Overall looking pretty good! Had a couple comments for consideration. Some of them might change approach a bit, but can always be deferred if you feel it is out of scope for this PR

internal/cli/bundle/remove.go
eoghanriley: That makes, I definitely think scope wise that should be a separate PR as this is already decently large.

internal/cli/bundle/remove.go
eoghanriley: Reworked this, the bundle should no longer be pulled/extracted twice anymore.

pkg/bundle/remove.go
eoghanriley: Removed

pkg/bundle/remove.go
eoghanriley: Did a lot of reworking so that reads should be pretty minimal.

pkg/bundle/remove.go
```diff
@@ -65,6 +72,19 @@ func Remove(ctx context.Context, source *DeploySource, opts RemoveOptions) (*Rem
 
 	s := logger.Bind(opts.Streams, opts.Config.Options.LogLevel)
 
+	if udsoci.IsOCIReference(source.BundlePath) || artifact.IsTarZst(source.BundlePath) {
```
chance-coleman, line 75: Can we use the bundle we already read before the prompt here? Reading it again means the file could change after the user confirms, so we might show one bundle and remove a different one.

eoghanriley: fixed, now prepareArtifactRemoveSource only performs a read if the provided bundle is nil

internal/cli/bundle/remove.go
joelmccoy: can you get a ticket created to track this

eoghanriley: forgot to link it https://linear.app/defense-unicorns/issue/CLI-320/split-cli-next-remove-into-uds-remove-commands

chance-coleman (approved):
looks good to me

joelmccoy (approved):

PR #1520, opened 2026-08-26
feat: add Zarf package disassembly command

Racer159:
## Description

This PR adds a Zarf package disassembly command to `uds-cli` - the goal is to take a built Zarf package and disassemble it into re-creatable source code that can be modified entirely offline.  This is possible today manually with `zarf tools archiver` commands but this makes the process a bit more streamlined and intentional.

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging


internal/mode/disassemble/beta.go
Racer159, line 1: We will likely want to talk to the Zarf team about how better to handle translations like this - there is a "safety" check test whenever the schemas (either beta or alpha) shift on us but this conversion should be owned more by Zarf than us in the future.

internal/mode/disassemble/disassemble_test.go
```diff
@@ -0,0 +1,439 @@
...
+	assert.True(t, component.Manifests[0].EnableTemplating)
+	assert.Empty(t, component.Selector.Flavor)
+
+	reassembled, err := assemble.AssemblePackage(t.Context(), generated, outputDir, assemble.AssembleOptions{SkipSBOM: true})
+	require.NoError(t, err)
+	defer func() { require.NoError(t, reassembled.Cleanup()) }()
+	assert.Equal(t, v1beta1.APIVersion, reassembled.PackageDefinition.OriginalAPIVersion())
+}
+
+func TestPackageSchemaCoverage(t *testing.T) {
```
Racer159: This is that safety check test

internal/legacy/cli/dev_disassemble.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/defenseunicorns/uds-cli/pkg/legacy/config"
+	"github.com/defenseunicorns/uds-cli/pkg/legacy/message"
+	"github.com/spf13/cobra"
+	"github.com/zarf-dev/zarf/src/pkg/packager/layout"
+)
+
+func newDevDisassembleCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "disassemble <source> <output-dir>",
+		Short: "[beta] Convert a Zarf package into rebuildable offline source",
```
joelmccoy: Does ti make sense to maybe start this off with an alpha description? there is a bunch of other places where we call this beta too that would need to be fixed if so.

Racer159:
since I know is'a comin' - @greptileai review

Racer159:
Want to improve UX a little so moved back to draft - specifically thinking about how a chart edit might be needed...  the zarf yaml I think can also use Zarf's helpers to be formatted a little cleaner and I want to explore that too.

PR #1521, opened 2026-08-27
feat(release): align snapshot releases with CLI Next

slaskawi:
## Description

Implements CLI-265 with unique snapshot releases built and tested only from `main`.

### How it works

1. Scheduled and manual runs check out and test `main`.
2. The workflow derives the latest stable semantic version and generates `v<version>-snapshot+<UTC timestamp>-<8-char SHA>`.
3. Release tests complete before the tag is pushed.
4. The generated tag is pushed and published with the snapshot release configuration.
5. Snapshot releases publish prerelease binaries, DEB/RPM packages, checksums, and SBOMs without updating Homebrew or `latest`.
6. Publication verifies the Linux amd64 checksum and both CLI modes report the generated tag.
7. Scheduled cleanup retains the three newest snapshot releases and tags.

### Safety boundaries

- Snapshot releases always build reviewed `main`; arbitrary branches and SHAs cannot reach release-test write permissions.
- Stable publishing skips `-snapshot+` tags.
- Duplicate generated tags fail before publication.
- Test failure prevents tag creation and publication.

## Related Issue

Relates to [CLI-265](https://linear.app/defenseunicorns/issue/CLI-265/align-uds-cli-nightly-and-ad-hoc-releases-with-uds-cli-next)

## Type of change

## Checklist before merging

## Validation

- YAML parsing passed for changed workflows and GoReleaser configurations.
- `.goreleaser.snapshot.yaml` passed `mise exec -- goreleaser check --config .goreleaser.snapshot.yaml`.
- `.goreleaser.yaml` retains a pre-existing deprecated `brews` property, so its GoReleaser check exits non-zero.
- `git diff --check` passed.
- PR CI is running.


.github/workflows/unstable-release.yaml
```diff
@@ -1,77 +1,140 @@
...
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          if [[ "$RELEASE_REF" =~ ^[0-9a-fA-F]{7,40}$ ]]; then
+            git rev-parse --verify "$RELEASE_REF^{commit}" >/dev/null
+          elif ! git ls-remote --exit-code --heads origin "refs/heads/$RELEASE_REF" >/dev/null 2>&1; then
+            echo "ref must be an existing branch or commit SHA: $RELEASE_REF" >&2
+            exit 1
+          fi
```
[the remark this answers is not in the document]
slaskawi: Fixed: [the prepare job now requires the selected ref to contain the unstable release config before tagging](https://github.com/defenseunicorns/uds-cli/blob/1c842dc23e7c51738200393b947681f145068d17/.github/workflows/nightly-release.yaml#L56).

.github/workflows/release-publish.yaml
```diff
@@ -0,0 +1,115 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG_NAME: ${{ inputs.tag_name }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          mkdir verify
+          gh release download "$TAG_NAME" --dir verify --pattern '*Linux*' --pattern 'checksums.txt'
+          (cd verify && sha256sum --ignore-missing -c checksums.txt)
+          binary=$(find verify -type f -name '*Linux*' ! -name '*.sbom' ! -name '*.deb' ! -name '*.rpm' | head -n 1)
```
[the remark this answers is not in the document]
slaskawi: Fixed: [verification now requires exactly one Linux amd64 binary](https://github.com/defenseunicorns/uds-cli/blob/1c842dc23e7c51738200393b947681f145068d17/.github/workflows/release-publish.yaml#L109).

.github/workflows/release-publish.yaml
```diff
@@ -0,0 +1,115 @@
...
+          test -n "$binary"
```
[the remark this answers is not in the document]
slaskawi: Fixed: [verification now requires exactly one Linux amd64 binary](https://github.com/defenseunicorns/uds-cli/blob/1c842dc23e7c51738200393b947681f145068d17/.github/workflows/release-publish.yaml#L109).

slaskawi:
@codex review

.github/workflows/nightly-release.yaml
```diff
@@ -1,77 +1,200 @@
...
+          mkdir verify
+          gh release download "$TAG_NAME" --dir verify --pattern '*Linux*' --pattern 'checksums.txt'
+          (cd verify && sha256sum --ignore-missing -c checksums.txt)
+          binary_count=$(find verify -maxdepth 1 -type f -name '*_Linux_amd64' ! -name '*.sbom' | wc -l)
+          test "$binary_count" -eq 1
+          binary=$(find verify -maxdepth 1 -type f -name '*_Linux_amd64' ! -name '*.sbom')
+          chmod +x "$binary"
+          "$binary" version | tee verify/legacy-version.txt
+          "$binary" --features=NextMode=true version | tee verify/next-version.txt
+          grep -F "uds version $TAG_NAME" verify/legacy-version.txt
```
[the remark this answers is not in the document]
slaskawi: Fixed: Legacy now validates its bare-tag output, while Next retains its `uds version` assertion.

.github/workflows/release.yaml
```diff
@@ -10,13 +10,14 @@ on:
...
+      - "!v*-nightly*"
+      - "!v*-adhoc*"
 
 jobs:
   test:
     uses: ./.github/workflows/test-release.yaml
     permissions:
       contents: read
-      id-token: write
       packages: write
```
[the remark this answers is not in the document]
slaskawi: Fixed by restoring OIDC permission to the stable release test job.

.github/workflows/unstable-release.yaml
```diff
@@ -1,77 +1,200 @@
...
   schedule:
-    - cron: "0 3 * * *" ## nightly at 3am UTC
+    - cron: "0 3 * * *" # nightly at 3am UTC
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: "Branch or commit SHA to release"
+        required: false
+        default: main
+        type: string
```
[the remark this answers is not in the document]
slaskawi: Fixed: maintainer workflow guidance and consumer-facing nightly prerelease details are now documented.

.github/workflows/snapshot-release.yaml
```diff
@@ -1,77 +1,200 @@
...
+          while read -r candidate; do
+            if [[ "$candidate" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+              stable_tag="$candidate"
+              break
+            fi
+          done < <(git tag --list 'v*' --sort=-version:refname)
+          test -n "$stable_tag"
+          timestamp=$(date -u +%Y%m%d%H%M%S)
+          short_sha=$(git rev-parse --short=8 HEAD)
+          tag_name="${stable_tag}-${RELEASE_TYPE}+${timestamp}-${short_sha}"
```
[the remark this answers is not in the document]
slaskawi: Fixed: ADR 0024 now records the immutable release model and its retention constraints.

docs/adr/0024-unique-unstable-release-tags.md
joelmccoy, line 1: I would remove the word "immutable" from everywhere in this PR... technically these are not immutable unless you use the immutable github releases feature: https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases. I don't think we need to switch to immutable releases, just don't present this as that.

These releases can still technically be overwritten.  Maybe call these `unique unstable releases`

.github/workflows/snapshot-release.yaml
joelmccoy, line 1: should we name this file `unstable-release` based on the workflow name change?

.goreleaser.snapshot.yaml
joelmccoy, line 1: I think we need to update this to ignore the new tags like `v*-nightly*` and `v*-adhoc`: https://github.com/defenseunicorns/uds-cli/blob/4eb05e5299c592daa263017eebd9e7f5b09920ec/.goreleaser.yaml#L141-L142

joelmccoy (commented):
A couple comments here

.github/workflows/snapshot-release.yaml
```diff
@@ -1,77 +1,200 @@
...
+    with:
+      ref: ${{ needs.prepare.outputs.source_sha }}
     permissions:
       contents: read
       id-token: write
       packages: write
     secrets: inherit
 
-  push:
+  tag:
```
bradclawsie, line 73: I don't have deep expertise in GH Actions syntax so I asked Codex to assist with a review. Hopefully this is relevant:

"_P1 — approval gate is bypassed before the release is published. The new tag job has
    contents: write and pushes the immutable remote tag, but it does not use the protected
    release-nightly environment._"

.github/workflows/snapshot-release.yaml
```diff
@@ -1,77 +1,200 @@
...
+        with:
+          ref: ${{ needs.prepare.outputs.source_sha }}
+          fetch-depth: 0
+          fetch-tags: true
+      - name: Push unstable tag
+        env:
+          TAG_NAME: ${{ needs.prepare.outputs.tag_name }}
+        run: |
+          git tag "$TAG_NAME" "${{ needs.prepare.outputs.source_sha }}"
+          git push origin "refs/tags/$TAG_NAME"
```
bradclawsie, line 93: Codex also cited this:

_"P2 — a publishing failure leaves an uncleaned orphan tag."_

Not sure if this is relevant because there is no contract that a tag must correspond to a release (or is there?), but just leaving that in case it is helpful

AGENTS.md
```diff
@@ -71,6 +71,12 @@ Keep user, contributor, and design docs synchronized with code changes.
 - If you change generated docs, run the corresponding docs generation or docs test task when practical.
 - If you make a significant design decision, add a new ADR.
 
+## Release workflows
+
+- Ask before dispatching `.github/workflows/nightly-release.yaml`. It writes to GHCR during validation and uses repository/package write permissions plus the `release-nightly` environment to create remote tags and prereleases.
```
bradclawsie: I would be interested in seeing some guidance on effective prompting here, just to be sure I never lead Codex astray

docs/adr/0024-unique-unstable-release-tags.md
slaskawi: The release tags are unique rather than GitHub immutable releases: [the ADR now uses accurate terminology](https://github.com/defenseunicorns/uds-cli/blob/48c107c5/docs/adr/0024-unique-unstable-release-tags.md#L1).

.github/workflows/snapshot-release.yaml
slaskawi: Leaving the tag behavior unchanged here: the workflow intentionally creates the tag before publishing, and cleanup is only defined for scheduled nightly releases.

AGENTS.md
slaskawi: This PR documents the release workflow and invariants; general Codex prompting guidance is separate scope and should be tracked in a follow-up issue.

.github/workflows/snapshot-release.yaml
slaskawi: Renamed as suggested: [`.github/workflows/unstable-release.yaml`](https://github.com/defenseunicorns/uds-cli/blob/48c107c5/.github/workflows/unstable-release.yaml).

.goreleaser.snapshot.yaml
slaskawi: Added the generated unstable tag patterns to GoReleaser's stable configuration exclusions: [`.goreleaser.yaml`](https://github.com/defenseunicorns/uds-cli/blob/48c107c5/.goreleaser.yaml#L143).

.github/workflows/snapshot-release.yaml
slaskawi: Added the `release-nightly` environment to the tag job so approval is required before the remote tag is created: [workflow](https://github.com/defenseunicorns/uds-cli/blob/48c107c5/.github/workflows/unstable-release.yaml#L96).

bradclawsie (dismissed):
let's give it a shot!

.github/workflows/snapshot-release.yaml
```diff
@@ -0,0 +1,201 @@
...
+
+          if git ls-remote --exit-code --refs origin "refs/tags/$tag_name" >/dev/null 2>&1; then
+            echo "tag already exists: $tag_name" >&2
+            exit 1
+          fi
+
+          echo "source_sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+          echo "tag_name=$tag_name" >> "$GITHUB_OUTPUT"
+
+  test:
```
chance-coleman, line 62: The workflow runs any selected branch or SHA with `packages: write` and `id-token: write` before reaching the release environment. A malicious branch could use these permissions. Can we remove the write permissions?

.github/workflows/unstable-release.yaml
```diff
@@ -0,0 +1,201 @@
...
+    permissions:
+      contents: write
+    steps:
+      - name: Remove older nightly releases
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          mapfile -t nightly_tags < <(gh release list --limit 100 --order desc --json tagName,isPrerelease --jq '.[] | select(.isPrerelease and (.tagName | test("^v[0-9]+\\.[0-9]+\\.[0-9]+-nightly\\+"))) | .tagName')
```
chance-coleman: After more than 100 releases exist, older nightlies will not be listed and will never be cleaned up. Maybe we should paginate the release list or query all nightly releases directly?

bradclawsie: I thought they got auto-deleted at 3?

chance-coleman: hmm yea i misunderstood what was happening here. There is a super edge case where nightlies could fall outside the 100 release count but i dont think it'll happen. Thanks for challenging me to understand this better, im going to resolve this comment.

.github/workflows/snapshot-release.yaml
joelmccoy: i believe these permissions are needed for some e2e tests (ghcr write/and bundle signature on cli next for token).

Instead I think we should lock down this workflow to only be run on the main branch. should shouldn't be cutting any unstable releases from any place other then there

chance-coleman: Ahh okay, yeah agreed if they're needed than thats a good solution.

bradclawsie: not to put words in @slaskawi 's mouth, but my impression of the `unstable` release was that you specifically desired something from a branch or a hash other than `main`/`HEAD`...if `main`/`HEAD` is always "stable"

joelmccoy: yeah unstable sounds a bit confusing here... but really we should NEVER cut a release artifact off something that is not merged into main (bypasses a lot of security and circumvents our peer review process).  The goal/intent of these nightlies/on-demand releases are pre-releases of main.

slaskawi: Fixed: snapshot releases now build only from `main`, so the test job’s required write permissions cannot run against an arbitrary branch or SHA.

slaskawi:
@codex review
@copilot review

.github/workflows/snapshot-release.yaml
```diff
@@ -0,0 +1,181 @@
...
+        with:
+          distribution: goreleaser
+          # renovate: datasource=github-releases depName=goreleaser/goreleaser
+          version: v2.17.1
+          args: release --clean --verbose --config .goreleaser.snapshot.yaml
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GORELEASER_CURRENT_TAG: ${{ needs.prepare.outputs.tag_name }}
+
+      - name: Verify published Linux amd64 artifact
```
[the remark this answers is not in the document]
slaskawi: Fixed: failed snapshot publication now cleans up the release and tag generated by that run.

.github/workflows/snapshot-release.yaml
```diff
@@ -0,0 +1,181 @@
...
+        with:
+          ref: ${{ needs.prepare.outputs.source_sha }}
+          fetch-depth: 0
+          fetch-tags: true
+      - name: Push snapshot tag
+        env:
+          TAG_NAME: ${{ needs.prepare.outputs.tag_name }}
+        run: |
+          git tag "$TAG_NAME" "${{ needs.prepare.outputs.source_sha }}"
+          git push origin "refs/tags/$TAG_NAME"
```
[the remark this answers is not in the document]
slaskawi: Fixed: a failed publish now removes only the snapshot tag created by that run.

.github/workflows/snapshot-release.yaml
```diff
@@ -0,0 +1,181 @@
...
+    permissions:
+      contents: write
+    steps:
+      - name: Remove older snapshot releases
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          mapfile -t snapshot_tags < <(gh release list --limit 100 --order desc --json tagName,isPrerelease --jq '.[] | select(.isPrerelease and (.tagName | test("^v[0-9]+\\.[0-9]+\\.[0-9]+-snapshot\\+"))) | .tagName')
```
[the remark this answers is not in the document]
slaskawi: Fixed: cleanup now explicitly selects the workflow repository.

.github/workflows/snapshot-release.yaml
```diff
@@ -0,0 +1,181 @@
...
+          fi
+
+          echo "source_sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+          echo "tag_name=$tag_name" >> "$GITHUB_OUTPUT"
+
+  test:
+    needs: prepare
+    uses: ./.github/workflows/test-release.yaml
+    with:
+      ref: main
```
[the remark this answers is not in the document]
slaskawi: Fixed: the release-test workflow now declares and forwards `ref`, and the snapshot workflow passes the prepared SHA.

.goreleaser.yaml
```diff
@@ -139,4 +139,4 @@ git:
   #
   # Templates: allowed.
   ignore_tags:
-    - nightly-unstable
+    - v*-snapshot*
```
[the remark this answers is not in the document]
slaskawi: Fixed the copyright range in `.goreleaser.yaml`.

.github/workflows/snapshot-release.yaml
```diff
@@ -0,0 +1,203 @@
...
+    steps:
+      - name: Remove failed snapshot release and tag
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG_NAME: ${{ needs.prepare.outputs.tag_name }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          gh release delete "$TAG_NAME" --yes || true
+          git push origin --delete "refs/tags/$TAG_NAME"
```
chance-coleman: Am i misunderstanding anything? trying to track through this action:  `cleanup-failed-snapshot` runs on a fresh runner without checking out the repository and without setting `GH_REPO` or passing `--repo`. Both `gh release delete` and `git push
  origin --delete` lack repository context, so wouldn't that failed publication still leave an orphaned snapshot tag?

bradclawsie:
Is there a forked repo in play where we can see this working? The PR feels quite complicated at this point (mostly due to the nature of Github Actions, nothing on you as a dev) and it seems difficult to assert behavior by just reading code.

.github/workflows/snapshot-release.yaml
joelmccoy: I believe this is assessment is correct and needs fixed.

.github/workflows/snapshot-release.yaml
```diff
@@ -0,0 +1,203 @@
...
+      - name: Remove older snapshot releases
+        env:
+          GH_REPO: ${{ github.repository }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          mapfile -t snapshot_tags < <(gh release list --limit 100 --order desc --json tagName,isPrerelease --jq '.[] | select(.isPrerelease and (.tagName | test("^v[0-9]+\\.[0-9]+\\.[0-9]+-snapshot\\+"))) | .tagName')
+          for tag_name in "${snapshot_tags[@]:3}"; do
+            gh release delete "$tag_name" --cleanup-tag --yes
```
joelmccoy: nit: i wonder if it is worth keeping the git tag around (these are cheap) and allows us to trace when a snapshot was made (and allows people to reference this snapshot via tag in the future)

```suggestion
            gh release delete "$tag_name" --yes
```

joelmccoy (commented):
Mostly looking good to me. 1 small non blocking nit comment, but Chance's comment should probably be addressed

joelmccoy:
this also might require a careful merge/rebase since we merged in the release-please pr

slaskawi:
We used the same workflow in UDS CLI Next; its published releases are available at https://github.com/defenseunicorns/uds-cli-next/releases. The full flow here creates an upstream tag and prerelease, so I won’t dispatch it without maintainer approval.

slaskawi:
Merged current `main` and resolved the release workflow conflict, preserving the snapshot-tag exclusion in [release.yaml](https://github.com/defenseunicorns/uds-cli/blob/091a60c4/.github/workflows/release.yaml#L8).

.github/workflows/snapshot-release.yaml
slaskawi: Good call. Scheduled cleanup will remove older prereleases while preserving their tags so snapshots remain referenceable.

.github/workflows/snapshot-release.yaml
slaskawi: Good catch. This cleanup job has neither a checkout nor repository context, so the failed tag can remain orphaned. I’ll use explicit `GH_REPO` and GitHub API-based tag deletion so both cleanup operations target the PR repository.

slaskawi: Yes, corrected this with the latest patch.

chance-coleman (approved):

joelmccoy (approved):

PR #1523, opened 2026-08-27
chore(docs): add cli next docs how to guides

chance-coleman:
## Description

Add end-user documentation for UDS CLI Next mode, covering bundle authoring, deployment, publishing, pulling, inspection, verification, and reconfiguration.

The guides explain when to use Next versus Legacy mode, keep unsigned workflows limited to local alpha testing, and require signature verification for registry workflows.

## Related Issue

Relates to CLI-221

## Type of change

Live k3d, OCI registry, and Cosign workflows were not run because they require environment-specific resources and credentials.

## Checklist before merging

- [x] [Contributor Guide](https://github.com/defenseunicorns/uds-cli/blob/main/CONTRIBUTING.md) followed


chance-coleman:
@greptileai @codex review these changes

.agents/skills/documentation-authoring/SKILL.md
chance-coleman, line 1: i took the liberty of adding the `.ai` directive for doc writing that is mostly the same as the uds-core directive. My thinking was that since we want a unified doc site it would make sense to have this in place earlier than later as we build out more docs.

docs/getting-started/overview.mdx
```diff
@@ -20,23 +20,33 @@ Choose your path based on your goal.
     <LinkCard title="Install UDS CLI" href="/cli/getting-started/installation/" />
   </Card>
 
-  <Card title="Quickstart" icon="open-book">
```
chance-coleman, line 23: I think because we have two different paths at the moment with legacy and next we should maintain the legacy quickstart. However because we want to encourage adoption i reordered this so that the next quick start shows up before legacy.

.agents/skills/documentation-authoring/SKILL.md
bradclawsie: Just learned about `.ai` and how it works, seems like a great idea and a good time to introduce it.

docs/getting-started/next-quickstart.mdx
```diff
@@ -0,0 +1,128 @@
...
+     version     = "0.1.0"
+   }
+
+   package "uds_k3d_dev" {
+     source = "oci://ghcr.io/defenseunicorns/packages/uds-k3d:0.20.2"
+     signature_verification { verify = false }
+   }
+
+   package "init" {
+     source = "oci://ghcr.io/zarf-dev/packages/init:v0.83.0"
```
bradclawsie: Do we want to feature 0.84, or is it still considered too new?

bradclawsie (commented):
looks good, just my one question

docs/getting-started/next-quickstart.mdx
chance-coleman: good catch, updated!

.agents/skills/documentation-authoring/SKILL.md
briantwatson: `.ai` a standard we're adopting?  We recently took a pass at moving everything to AGENTS.md and `.agents`. .  https://github.com/defenseunicorns/uds-core/commit/e00b9127e6cfb992398041ef487316fbf80e6ca9

Technically I think a lot of the loose md files we had across several repos made sense to structure as skills like we did there.

docs/getting-started/next-quickstart.mdx
```diff
@@ -0,0 +1,128 @@
...
+   }
+
+   metadata {
+     name        = "next-quickstart"
+     description = "Next mode quickstart bundle"
+     version     = "0.1.0"
+   }
+
+   package "uds_k3d_dev" {
+     source = "oci://ghcr.io/defenseunicorns/packages/uds-k3d:0.20.2"
```
briantwatson, line 49: Do we have a mechanism that updates versions in our docs?

docs/getting-started/next-quickstart.mdx
briantwatson, line 1: If next will become default, does it make sense for this doc to take the `quickstart.mdx` and the pre-existing one to become `legacy-quickstart.mdx`?  Maybe this can come later, but seeing the re-ordering in linkcards has me thinking 'next' should become the default.

docs/how-to-guides/next-create-a-bundle.mdx
```diff
@@ -0,0 +1,124 @@
...
+       }
+     }
+   }
+   ```
+
+   Create the values file referenced by the `api` package. Replace the example field with a value supported by the package's charts:
+
+   ```yaml title="values/api.yaml"
+   # Replace this example with values supported by the API package.
+   replicaCount: {{ .vars.replica_count }}
```
briantwatson: Does CLI support templated Values files like so?  I actually don't know, but just had me wondering if a raw value would be clearer or no.

docs/how-to-guides/next-create-a-bundle.mdx
```diff
@@ -0,0 +1,124 @@
...
+   ```
+
+    The package ID is the label after `package`. Use `depends_on` to control deployment order. Independent packages deploy in parallel.
+
+    Each package needs one verification method: `public_key`, `keyless`, or an explicit `signature_verification { verify = false }` bypass. Use the bypass only for local alpha workflows. It produces a warning.
```
briantwatson: Package ID comment seemed out of place or unclear.  Rereading I believe now it's relevant for `depends_on`, the linkage might not be clear.  So maybe something like:
 'Use the package ID with `depends_on` to control deployment order by referencing, for example...'

I wonder if these lines would be better served at top of section, and the verification method maybe in a Note or Important callout, then the code block reference below. Defer to your judgement here!

docs/how-to-guides/next/create-a-bundle.mdx
```diff
@@ -0,0 +1,124 @@
...
+   ```
+
+   For a local unsigned test artifact:
+
+   ```bash
+   CLI_FEATURES=NextMode=true uds bundle create . --unsigned
+   ```
+
+    Use `--keyless` to sign through an OIDC identity. The command writes a local `.tar.zst` containing the definition, defaults, values, and package content.
+
```
briantwatson, line 148: This might benefit from an example like the two above?

slaskawi (dismissed):

docs/how-to-guides/next/deploy-a-bundle.mdx
```diff
@@ -0,0 +1,120 @@
+---
+title: "Deploy a bundle in Next mode"
+description: Deploy a bundle in Next mode from its HCL definition or a local or OCI artifact.
+sidebar:
+  order: 3.600
+---
+
+import { Steps } from '@astrojs/starlight/components';
+
+Use `bundle dev deploy` for a definition and `bundle deploy` for a created artifact.
```
briantwatson, line 10: Not sure on this, but do we want to have `uds` in the commands here?  I see they're in the Reference commands.  Maybe makes sense to not have here.

docs/how-to-guides/next/deploy-a-bundle.mdx
```diff
@@ -0,0 +1,120 @@
...
+- Deploy a local or OCI artifact.
+- Select packages for an intentional partial deployment.
+
+## Prerequisites
+
+- [UDS CLI installed](/cli/getting-started/installation/)
+- A directory containing `bundle.uds.hcl`, or a created `.tar.zst` artifact
+- A matching bundle public key such as `./cosign.pub` for signed artifact examples
+- Access to the target Kubernetes cluster
+- Registry access when deploying an OCI artifact
```
briantwatson, line 25:
```suggestion
- OCI access to each package source
```
Might just align wording with pre-req on doc above

docs/how-to-guides/next-deploy-a-bundle.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+- Deploy from a bundle definition.
+- Deploy a local or OCI artifact.
+- Select packages for an intentional partial deployment.
+
+## Prerequisites
+
+- [UDS CLI installed](/cli/getting-started/installation/)
+- A directory containing `bundle.uds.hcl`, or a created `.tar.zst` artifact
+- A matching bundle public key such as `./cosign.pub` for signed artifact examples
```
briantwatson:
```suggestion
- The matching bundle public key used to signed the artifact
```
Maybe simpler and clearer?

docs/how-to-guides/next-inspect-verify.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+- Inspect bundle metadata.
+- Verify a signature with a public key or keyless constraints.
+- Emit JSON or YAML for automation.
+
+## Prerequisites
+
+- [UDS CLI installed](/cli/getting-started/installation/)
+- A local `.tar.zst` bundle artifact or an OCI bundle reference
+- The public key or keyless certificate constraints for a signed artifact
```
briantwatson:
```suggestion
- The matching bundle public key used to signed the artifact
```
Same as my above, your preference on wording here, just might be clean to align them

docs/how-to-guides/next-publish-pull.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+- Publish a signed artifact.
+- Pull it into a local directory.
+- Deploy the pulled artifact.
+
+## Prerequisites
+
+- [UDS CLI installed](/cli/getting-started/installation/)
+- A bundle directory containing `bundle.uds.hcl` and its referenced files, or an existing `.tar.zst` bundle artifact
+- OCI access to each package source and any public keys referenced by `bundle.uds.hcl` when creating an artifact
```
briantwatson:
```suggestion
- OCI access to each package source
```

docs/how-to-guides/next-publish-pull.mdx
```diff
@@ -0,0 +1,107 @@
...
+- A signing key or keyless identity for the create step. The example uses `./cosign.key`.
+- A matching public key or keyless certificate constraints for pull and deploy. The example uses `./cosign.pub`.
+- Authentication and write access to the destination OCI registry
+- Read access to the registry when pulling the artifact
```
briantwatson: This whole section seems much more verbose than others. I think the simpler pre-reqs is more approachable.  Can probably be distilled down to at least 2 fewer lines I'm thinking with shorter text.

docs/how-to-guides/next-publish-pull.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+Replace `<ARCH>` with the architecture in the path printed by `bundle pull`, such as `amd64` or `arm64`. The output filename includes the bundle name, architecture, and version.
+
+## Troubleshooting
+
+### Problem: The registry rejects the push
+
+**Symptom:** `bundle push` cannot authenticate or write the destination artifact.
+
+**Solution:** Authenticate to the registry, confirm that the repository exists or that your account can create it, and retry the push.
```
briantwatson: This had me reflecting on when we might want to offer troubleshooting steps and I think as long as it pertains to UDS behavior it might make sense.  This is just generic OCI Registry troubleshooting and I think is worth considering dropping altogether.

docs/how-to-guides/next-reconfigure-a-bundle.mdx
```diff
@@ -0,0 +1,121 @@
...
+- Create a derivative in a local directory or OCI registry.
+- Keep the package content unchanged.
+
+## Prerequisites
+
+- [UDS CLI installed](/cli/getting-started/installation/)
+- A local `.tar.zst` artifact or an OCI bundle reference. Use a signed artifact outside local alpha workflows.
+- A `defaults.uds.hcl` file containing the replacement variables
+- A public key or keyless verification policy for the source artifact
+- A signing key or keyless signing identity for a signed derivative
```
briantwatson: Again here, would just suggest we align exact wording across how-to pre-reqs if makes sense.  I think these align with publish-pull doc pre-reqs

briantwatson (commented):
Great set of doc additions here -- lot's of good content.  I've got some questions and suggestions for clarity among several.

bradclawsie (dismissed):
lgtm

.agents/skills/documentation-authoring/SKILL.md
chance-coleman: yeah @briantwatson thats a good point. I should've looked at what we adopting for a standard with skills instead of going off memory of what we did previously. I put zero thought into what/where this should live look like. I'm going to update this to be more similar to the pattern we're using in other places.

docs/getting-started/next-quickstart.mdx
chance-coleman: Added a Renovate custom manager in renovate.json to track versioned OCI source references in README.md and docs/**/*.mdx

docs/how-to-guides/next-create-a-bundle.mdx
chance-coleman: Yea, `values_files` supports Go templates. Added guidance explaining that `{{ .vars.* }}` values are resolved at deploy time from defaults or config, and missing variables fail deployment.

docs/how-to-guides/next-create-a-bundle.mdx
chance-coleman: Reworded this to connect the package ID directly to `depends_on` and added an explicit example showing the dependency reference.

docs/how-to-guides/next/deploy-a-bundle.mdx
chance-coleman: My thinking was that these are copy-paste command examples. The command group alone would not be executable. Could be swayed in that opinion if you or others feel strongly

docs/how-to-guides/next/create-a-bundle.mdx
chance-coleman: Added a concrete locals example showing how to reuse registry and version values in package sources.

docs/how-to-guides/next/deploy-a-bundle.mdx
chance-coleman: Updated the prerequisite to call out OCI access to each package source

docs/how-to-guides/next-deploy-a-bundle.mdx
chance-coleman: Updated this to use matching bundle public-key or keyless certificate wording

docs/how-to-guides/next-inspect-verify.mdx
chance-coleman: Aligned this prerequisite with the deploy guide using matching bundle public-key wording

docs/how-to-guides/next-publish-pull.mdx
chance-coleman: Updated this to call out OCI access to each package source when creating an artifact.

docs/how-to-guides/next-publish-pull.mdx
chance-coleman: took this from seven lines to five and combined the signing, verification, and registry access requirements.

docs/how-to-guides/next-publish-pull.mdx
chance-coleman: Removed the generic OCI registry troubleshooting section. The remaining troubleshooting guidance is focused on UDS-specific behavior.

docs/how-to-guides/next-reconfigure-a-bundle.mdx
chance-coleman: Aligned the reconfigure prerequisites with the other bundle guides

docs/getting-started/next-quickstart.mdx
chance-coleman: Yeah i see what you mean here.. Regardless of what we do, when we deprecate and remove the legacy quickstart guide we will have to update things here. So I'd probably vote to just kick the can down the road.

.agents/skills/documentation-authoring/SKILL.md
joelmccoy: do we also want the templates for this repo? https://github.com/defenseunicorns/uds-core/tree/main/.agents/skills/documentation-authoring/references

chance-coleman: @joelmccoy thats a good call, i added those templates. Also updated them a bit to remove any core specific pieces.

briantwatson (dismissed):
Appreciate all the changes -- nice additions to docs!

docs/how-to-guides/next/create-a-bundle.mdx
joelmccoy, line 1: I wonder if we should put all these next mode how to guides under a sub folder for next mode.  If not we should probably order/prefix the titles better. See sidebard snapshot: The verify keyless package... guide is for legacy and mixed in the order with the next docs

Image

joelmccoy: If we don't want a subfolder maybe start every doc title with `Next -` like `Next - Create a Bundle`

docs/how-to-guides/next/create-a-bundle.mdx
```diff
@@ -0,0 +1,138 @@
...
+         certificate_identity    = "https://github.com/my-org/api/.github/workflows/release.yml@refs/heads/main"
+         certificate_oidc_issuer = "https://token.actions.githubusercontent.com"
+       }
+     }
+   }
+   ```
+
+   Create the values file referenced by the `api` package. Values files support Go templates. The `{{ .vars.* }}` expressions read variables available at deploy time, including values from `defaults.uds.hcl` or `config.uds.hcl`; missing variables cause deployment to fail. Replace the example field with a value supported by the package's charts:
+
+   ```yaml title="values/api.yaml"
```
joelmccoy, line 71: we should probably call out that these are inputs to zarf values on the zarf package.  I think we are missing bridging the gap here.  These values get mapped based on the values mapping in zarf.  I think we should include an example of what that looks like on the zarf side so we can see how it maps to actual helm chart values

joelmccoy (commented):
Looking pretty good!  A couple small comments and a couple things that I might request as follow on work if you can create tickets for them:

1. A how-to guide for signing a bundle (including generating a cosign key with zarf)
2. A reference doc for `config.uds.hcl`, this should include all things that can be set in the config (we don't have a good doc for this).
3. A reference doc for `bundle.uds.hcl` this should include all the things/fields we can set in the bundle per package

I think as a developer have a single one page reference for the spec for config and bundle definitions would be SUPER helpful

docs/how-to-guides/next/create-a-bundle.mdx
chance-coleman: i ended up going with a single `next/` subfolder for now. this keeps the existing Legacy URLs stable, gives the Next docs a clear group in the sidebar, and keeps the future transition open. once Next is stable, we can move those docs to the top level and handle redirects and latest-version behavior then.

docs/how-to-guides/next/create-a-bundle.mdx
chance-coleman: i added an example showing the full mapping from the `replica_count` bundle variable, to the rendered replicas value in the bundle values file, to Zarf’s `sourcePath`, and finally to the Helm chart’s `targetPath` of `replicaCount`.

joelmccoy (approved):

PR #1525, opened 2026-08-28
fix!: keep CLI features out of Zarf tools

slaskawi:
## Description

Prevent normalized `CLI_FEATURES` values from being reinserted into `zarf tools` callback arguments. This fixes Zarf action callbacks such as `uds zarf tools kubectl ...`, which reject the root `--features` flag.

## Related Issue

Fixes CLI-317

## Type of change

## Checklist before merging


slaskawi:
@greptileai review

slaskawi:
@codex review

slaskawi:
@copilot review

slaskawi:
@greptileai review

slaskawi:
@copilot review

slaskawi:
@codex review

slaskawi:
@codex please review the latest commit b02e3062.

slaskawi:
@copilot review

slaskawi:
@codex review

slaskawi:
@greptileai review

tests/legacy/e2e/zarf_test.go
```diff
@@ -90,3 +90,14 @@ func TestZarfFeatureBootstrap(t *testing.T) {
 		})
 	}
 }
+
+func TestZarfToolsIgnoreCLIFeatures(t *testing.T) {
+	for _, features := range []string{"values=true", "NextMode=true,values=true"} {
```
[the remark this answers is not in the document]
slaskawi: Fixed in f164b895. Legacy E2E now covers only values=true; the NextMode=true binary scenario moved to tests/integration/cmd/zarf.

internal/mode/mode.go
```diff
@@ -155,6 +155,9 @@ func knownFeature(name string) bool {
 func addZarfFeatures(args []string, features FeatureSet) []string {
 	// Mode resolution consumes the shared --features flag before Cobra runs. Restore every
 	// feature other than NextMode after the zarf command so Zarf still receives its features.
+	if rootZarfToolsCommandIndex(args) >= 0 {
+		return args
```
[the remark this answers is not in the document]
slaskawi: Fixed in f164b895. The bypass is now limited to recognized vendored Zarf tools; non-vendor tools retain Zarf feature forwarding. Added coverage for the non-vendor path.

internal/mode/mode.go
```diff
@@ -170,6 +173,26 @@ func addZarfFeatures(args []string, features FeatureSet) []string {
...
+	if index < 0 {
+		return false
+	}
+	for i := index + 1; i < len(args); i++ {
+		if args[i] == "tools" || args[i] == "t" {
+			if i+1 >= len(args) {
+				return false
+			}
+			switch args[i+1] {
+			case "kubectl", "k", "syft", "sbom", "s", "k9s", "monitor", "m", "wait", "w", "crane", "registry", "r", "helm", "h", "yq":
```
slaskawi: Unfortunately Zarf doesn't expose any way to figure out the tools. 

@brandtkeller Would it be acceptable to expose something like `VendorCommands()` through Zarf? If so - I'd be happy to create and Issue and a PR for it.

brandtkeller: An issue describing the core problem would be great.

mjnagel (commented):
In testing this does appear to fix the `kubectl` regression I experienced. I still think there's a broader concern though that is worth addressing or clearly deferring/opening a new issue for... `--features` / `CLI_FEATURES` still mixes UDS-owned features (for example, NextMode) with Zarf-owned features (for example, values) in one set. I'm not sure if there's any particular issues with this today but this could potentially cause problems in the future if:
- CLI or Zarf perform any validation around feature flags (that the selected flag actually exists)
- CLI/Zarf have overlapping feature flags (same name - this might be okay if they are meant to do the same thing)

Even putting these potential issues aside, I think it could be a bit confusing that feature flags meant for the CLI get passed to Zarf and vice-versa?

Edit: also (as I know you flagged) there's some clunkiness around the way this PR filters out features from the tool commands. I wonder whether this could be cleaner if the distinction between Zarf and CLI features was more explicitly designed/accounted for?

internal/mode/mode.go
```diff
@@ -170,6 +173,26 @@ func addZarfFeatures(args []string, features FeatureSet) []string {
 	return args
 }
 
+func zarfVendorTool(args []string) bool {
+	index := rootZarfToolsCommandIndex(args)
+	if index < 0 {
+		return false
```
chance-coleman: Can we handle `__complete` and `__completeNoDesc` here too? Zarf treats these as vendor commands, but this check does not.

internal/mode/mode.go
joelmccoy: can we just assume if it starts with tools we cane assume it is a tools operation? do we really need to check the sub tools here or am i missing something?

mjnagel: There's a few sub-tools that are "zarf owned" rather than vendored, and technically accept the feature flags (ex: `zarf tools update-creds`, `zarf tools download-init`, ...). I'm not sure any strictly need/use the feature flags today but that could change?

joelmccoy:
> In testing this does appear to fix the `kubectl` regression I experienced. I still think there's a broader concern though that is worth addressing or clearly deferring/opening a new issue for... `--features` / `CLI_FEATURES` still mixes UDS-owned features (for example, NextMode) with Zarf-owned features (for example, values) in one set. I'm not sure if there's any particular issues with this today but this could potentially cause problems in the future if:
> 
> * CLI or Zarf perform any validation around feature flags (that the selected flag actually exists)
> * CLI/Zarf have overlapping feature flags (same name - this might be okay if they are meant to do the same thing)
> 
> Even putting these potential issues aside, I think it could be a bit confusing that feature flags meant for the CLI get passed to Zarf and vice-versa?
> 
> Edit: also (as I know you flagged) there's some clunkiness around the way this PR filters out features from the tool commands. I wonder whether this could be cleaner if the distinction between Zarf and CLI features was more explicitly designed/accounted for?

After seeing this I wish we would have taken the approach of separating the features. Like using `--uds-features` for cli features and `--features` for zarf features. I'd be happy to make this split as it is much more cleaner.  This would be a breaking change but I think a worthwhile one for longevity here.

Created a ticket here: https://linear.app/defense-unicorns/issue/CLI-323/separate-ownership-of-uds-cli-and-zarf-feature-flags

internal/mode/mode.go
slaskawi: Implemented completion routing to match Zarf’s vendor detection, with coverage for both Cobra completion commands in [mode.go](https://github.com/defenseunicorns/uds-cli/blob/sebastianlaskawiec/cli-317/internal/mode/mode.go#L283).

internal/mode/mode.go
slaskawi: Micah is right. Feature ownership now follows argument position, so flags after either Zarf entrypoint pass through unchanged and the vendored-tool allowlist is gone in [mode.go](https://github.com/defenseunicorns/uds-cli/blob/sebastianlaskawiec/cli-317/internal/mode/mode.go#L60).

slaskawi:
Implemented position-sensitive ownership in [mode.go](https://github.com/defenseunicorns/uds-cli/blob/sebastianlaskawiec/cli-317/internal/mode/mode.go#L60): root `--features`/`CLI_FEATURES` configure UDS, while everything after either Zarf entrypoint passes to Zarf unchanged.

tests/integration/cmd/zarf/zarf_integration_test.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"os/exec"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/defenseunicorns/uds-cli/tests/testutil"
+)
+
+func TestZarfToolsIgnoreCLIFeatures_NextMode(t *testing.T) {
```
joelmccoy, line 17: wouldn't we want to test the failing commands from before here? I feel like this integration test doesn't add any value here from current state... we should have a bunch of integration tests for the failing command at a minimum

```console
uds --features=NextMode zarf --features=values=true zarf tools kubectl version --client
     ERROR:  duplicate --features option
```

Should the above work?  We should be able to pass in features for both uds and zarf correct?

I think we should take a step back and understand what the intended behavior for our feature flags is. Maybe start with an ADR on how we will handle this.

I definitely think we need to at a minimum to support a way to pass feature flags into UDS CLI AND zarf in the same command without causing breaking commands paths (like the one that was stated in this issue).

docs/reference/next-mode.mdx
```diff
@@ -99,6 +99,11 @@ uds --features=NextMode=true version
 CLI_FEATURES=NextMode=true uds version
 ```
 
+`CLI_FEATURES` and `--features` configure UDS only when they appear before a
+Zarf command. Arguments after `uds zarf`/`uds z` or `uds tools zarf`/`uds
```
chance-coleman, line 103: Should `uds z` or `uds tools z` work with next? I dont think we have that alias setup in next mode, at least for me `CLI_FEATURES=NextMode=true uds z ...` does not work, maybe im missing something.

slaskawi: Good catch. Next mode recognized  during feature resolution but did not register it with Cobra; I added the aliases and coverage in [the Zarf command wiring](https://github.com/defenseunicorns/uds-cli/blob/sebastianlaskawiec/cli-317/internal/cli/zarf/zarf.go#L81).

slaskawi: Good catch. Next mode recognized `z` during feature resolution but did not register it with Cobra; I added the aliases and coverage in [the Zarf command wiring](https://github.com/defenseunicorns/uds-cli/blob/sebastianlaskawiec/cli-317/internal/cli/zarf/zarf.go#L81).

joelmccoy (approved):
LGTM - converted this to a breaking change in the pr title, but I think this small change is worth while

chance-coleman (approved):
🚀 looks good

PR #1527, opened 2026-08-28
fix: gitleaks shall ignore uds-docs

bradclawsie:
## Description

After running `uds run dev-docs`, there is a directory `uds-docs` left behind (that is in `.gitignore`) that will be scanned by `gitleaks` via the pre-commit hook invoked by `hk`. Commits cannot take place since `gitleaks` will find objectionable matter in `uds-docs`. 

This change causes `gitleaks` to skip the generated (and `.gitignore`'d) `uds-docs` path.
...

## Related Issue

Fixes #CLI-321

## Type of change

## Checklist before merging

This was tested by running `uds run dev-docs` and then trying to commit this PR change itself.

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -7,7 +7,3 @@ useDefault = true
 [[allowlists]]
 description = "vendored Go dependencies"
 paths = ['''^vendor/''']
-
-[[allowlists]]
-description = "local generated documentation workspace"
-paths = ['''^uds-docs/''']
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -7,3 +7,7 @@ useDefault = true
 [[allowlists]]
 description = "vendored Go dependencies"
 paths = ['''^vendor/''']
+
+[[allowlists]]
+description = "local generated documentation workspace"
+paths = ['''^uds-docs/''']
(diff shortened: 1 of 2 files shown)


joelmccoy:
The more clean fix would be to have the hk gitleaks take in files: https://github.com/defenseunicorns/uds-cli/blob/main/hk.pkl#L21

something like this?
```
check = "gitleaks dir --redact --no-banner --exit-code 1 {{files}}"
```

bradclawsie:
> The more clean fix would be to have the hk gitleaks take in files: https://github.com/defenseunicorns/uds-cli/blob/main/hk.pkl#L21
> 
> something like this?
> 
> ```
> check = "gitleaks dir --redact --no-banner --exit-code 1 {{files}}"
> ```

I'm leaning on Codex here, this is my first contact with `gitleaks`. It responds:

```
gitleaks dir accepts a single path. With multiple arguments, Gitleaks 8.30.1 scans
  the first and silently ignores the rest, so:

  gitleaks dir ... {{files}}

  could miss secrets in every changed file after the first.
```

Suggested alternative in https://github.com/defenseunicorns/uds-cli/pull/1527/changes/e3e10721ba4c111868c0af1876c146966a8f1dc3 . Let me know what you think.

joelmccoy (approved):

PR #1528, opened 2026-08-28
feat(next)!: operator monitor command

eoghanriley:
## Description

Next implementation of `uds monitor pepr`

## Related Issue

Fixes CLI-285

## Type of change

## Checklist before merging


internal/operator/monitor_test.go
```diff
@@ -0,0 +1,367 @@
...
+			var out bytes.Buffer
+			if tt.out == nil {
+				tt.out = &out
+			}
+			var errOut bytes.Buffer
+			streams := iostreams.New(nil, tt.out, &errOut)
+			err := Monitor(context.Background(), streams, MonitorOptions{source: tt.source, NoColor: true})
+			if tt.wantErr != "" {
+				require.ErrorContains(t, err, tt.wantErr)
+				assert.ErrorIs(t, err, tt.wantTarget)
```
chance-coleman: The required checks / Lint job fails because `assert.ErrorIs` violates testifylint’s require-error rule.

internal/operator/monitor.go
```diff
@@ -0,0 +1,561 @@
...
+			}
+			if opts.Since > 0 {
+				seconds := int64(math.Ceil(opts.Since.Seconds()))
+				logOpts.SinceSeconds = &seconds
+			}
+
+			opened := false
+			warnedEnded := false
+			for {
+				logStream, err := source.StreamLogs(ctx, target.pod, logOpts)
```
chance-coleman: When the log connection ends, the command asks for the same logs again from the beginning of the selected range. Events that were already printed can appear again after reconnecting.

internal/operator/monitor.go
```diff
@@ -0,0 +1,561 @@
...
+			return fmt.Errorf("%w: %w", ErrCreateKubernetesClient, err)
+		}
+		source = kubernetesPodLogSource{client: client}
+	}
+
+	pods, err := source.ListPods(ctx)
+	if err != nil {
+		return fmt.Errorf("%w: %w", ErrListPeprPods, err)
+	}
+	targets := selectContainers(pods, opts.Stream)
```
chance-coleman, line 201: The command finds the pods only once when it starts. If a controller restarts, it usually gets a new pod name, but this code keeps trying the old name and never reads the new pod.

internal/operator/monitor.go
```diff
@@ -0,0 +1,561 @@
...
+				logOpts.SinceSeconds = &seconds
+			}
+
+			opened := false
+			warnedEnded := false
+			for {
+				logStream, err := source.StreamLogs(ctx, target.pod, logOpts)
+				if err != nil {
+					streamErr := fmt.Errorf("%w %s: %w", ErrStreamPodLogs, target.pod, err)
+					if !opened {
```
chance-coleman: If opening one log stream fails before any logs are read, this worker exits even with `--follow`. If another stream works, the command continues running but never tries the failed stream again, so that controller’s events are missing for the rest of the session.

internal/cli/core/operator_monitor.go
```diff
@@ -0,0 +1,123 @@
...
+
+// newOperatorMonitorCommand is the underlying implementation of NewOperatorMonitorCommand.
+// They are separate functions for the sake of testing, as being able to specify the monitorFunc allows tests
+// to skip connecting to kubernetes
+func newOperatorMonitorCommand(streams iostreams.IOStreams, monitor monitorFunc) *cobra.Command {
+	o := NewOperatorMonitorOptions(streams)
+	o.monitor = monitor
+
+	cmd := &cobra.Command{
+		Use:   "monitor [policies | operator | allowed | denied | failed | mutated]",
```
chance-coleman, line 53: The PR adds uds core operator monitor, but the user guide and command reference still show only uds monitor pepr

internal/operator/monitor_test.go
eoghanriley: Fixed

internal/operator/monitor.go
eoghanriley: fixed, added timestamp tracking

internal/cli/core/operator_monitor.go
eoghanriley: I am probably looking in the wrong places, but I haven't been able to find next documentation that refers to `monitor pepr`. When running `uds --features=NextMode=true core operator` the usage output is as shown below
```
Usage:
  uds core operator [command]

Available Commands:
  monitor     Observe UDS Core operator activity
  ```
  
  For legacy usage the command is still `uds monitor pepr`

File: internal/operator/monitor.go (hunk 4 above)
Comment (eoghanriley) [reply internal/operator/monitor.go]: Should be fixed as part of the partial rewrite fixing looking for new pods. I also added a test to cover this.

File: internal/operator/monitor.go (hunk 3 above)
Comment (eoghanriley) [reply internal/operator/monitor.go:201]: Should be fixed now, added followMonitor which supervisors the following workers.

File: internal/cli/core/operator_monitor.go (hunk 5 above)
Comment (chance-coleman) [reply internal/cli/core/operator_monitor.go:53]: Sorry, this is my fault, poorly worded and placed comment... I was referring to older legacy docs but with our split subdir for next docs I really dont think this is something i should've commented on. Im going to resolve this. I think what you have is great!

**Review by chance-coleman — DISMISSED** [review]
nice work, happy with this. I'd probably appreciate one more set of eyes on this.

File: internal/cli/core/operator_monitor.go [hunk 6]
```diff
@@ -0,0 +1,124 @@
...
+		Short: "Observe UDS Core operator activity",
+		Long:  "Stream UDS Core operator activity and related Pepr policy decisions from the cluster.",
+		Example: `  # Stream all operator and policy events
+  uds core operator monitor
```
joelmccoy, line 57: Could we mark this as [alpha]? kind of like we do [here](https://github.com/defenseunicorns/uds-cli/blob/ebe08bc3d08356ea16100b0dbddd98903e0ef15d/pkg/legacy/config/lang/lang.go#L80) or how zarf does this for all it's new big commands ([ex](https://github.com/zarf-dev/zarf/blob/fb3a53b276ed927fa5e279a826ee366243f9d9b7/src/config/lang/english.go#L179)).

I think any new big feature we should mark as alpha to have grace on making breaking changes here.  Although this was a port of existing functionality (with some good cleanup), I imagine the shape of this could change/break.  I think marking this as alpha in the command help (and hopefully eventually in command reference sheet) would be very helpful for us.

joelmccoy (commented):
Generally LGTM - great job porting and removing bubblecharm dep. One comment for consideration, i would recommend marking this feature as alpha

internal/cli/core/operator_monitor.go
eoghanriley: Added

joelmccoy (approved):

PR #1533, opened 2026-08-31
fix(legacy): preserve --set overrides after config reload

chance-coleman:
## Description

  Preserve parsed Legacy `--set` deploy overrides when `UDS_CONFIG` reloads bundle configuration.

  This prevents `uds deploy` from panicking when pflag restores changed string-to-string flags after config unmarshalling. Adds regression coverage for an empty `uds-config.yaml` and multiple deploy overrides.

  ## Related Issue

  Fixes CLI-326

  ## Type of change

  ## Steps to Validate

  1. Create an empty `uds-config.yaml`.
  2. Run:

     ```bash
     UDS_CONFIG=./uds-config.yaml \
     uds deploy ./bundle-does-not-need-to-exist.tar.zst --confirm --no-progress \
       --set dependency-operator.OBJECT_STORAGE_PROVIDER=seaweedfs \
       --set dependency-operator.CI=false

  3. Verify the command reaches bundle validation and does not panic.

  Validation run: `uds run test:unit`

  ## Checklist before merging

  - [x] Test added. Documentation and ADR updates are not needed for this regression fix.
  - [ ] [Contributor Guide](https://github.com/defenseunicorns/uds-core/blob/main/CONTRIBUTING.md) followed

### DIFF
--- a/internal/legacy/cli/root.go
+++ b/internal/legacy/cli/root.go
@@ -226,14 +226,13 @@ func loadViperConfig() error {
 		return err
 	}
 
-	// Preserve parsed --set values because pflag retains its changed state when flags are restored.
-	setVariables := bundleCfg.DeployOpts.SetVariables
+	hadSetVariables := bundleCfg.DeployOpts.SetVariables != nil
 	err = unmarshalAndValidateConfig(configFile, &bundleCfg)
 	if err != nil {
 		return err
 	}
-	if setVariables != nil {
-		bundleCfg.DeployOpts.SetVariables = setVariables
+	if hadSetVariables && bundleCfg.DeployOpts.SetVariables == nil {
+		bundleCfg.DeployOpts.SetVariables = map[string]string{}
 	}
 
 	// ensure the DeployOpts.Variables pkg vars are uppercase
--- a/internal/legacy/cli/root.go
+++ b/internal/legacy/cli/root.go
@@ -1,4 +1,4 @@
-// Copyright 2024 Defense Unicorns
+// Copyright 2024-2026 Defense Unicorns
 // SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
 
 package cmd
@@ -226,9 +226,14 @@ func loadViperConfig() error {
 		return err
 	}
 
+	// Preserve parsed --set values because pflag retains its changed state when flags are restored.
+	setVariables := bundleCfg.DeployOpts.SetVariables
 	err = unmarshalAndValidateConfig(configFile, &bundleCfg)
 	if err != nil {
 		return err
+	}
+	if setVariables != nil {
+		bundleCfg.DeployOpts.SetVariables = setVariables
 	}
 
 	// ensure the DeployOpts.Variables pkg vars are uppercase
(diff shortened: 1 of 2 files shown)


chance-coleman:
@greptileai @codex review these changes

internal/legacy/cli/root.go
```diff
@@ -226,10 +226,15 @@ func loadViperConfig() error {
 		return err
 	}
 
+	// Preserve parsed --set values because pflag retains its changed state when flags are restored.
+	setVariables := bundleCfg.DeployOpts.SetVariables
```
joelmccoy: I think this bug only exists when the provided uds config file is empty... I could not get this to fail for any real yaml file with content.

this approach causes a regression and not the most accurate fix. If `uds-config.yaml` contains `setVariables` and the user also passes `--set`, it preserves the pre-unmarshal CLI map and replaces the config-loaded `setVariables`, causing config values to be dropped instead of merged with CLI overrides....

A more targeted fix is to handle only the case that causes the panic: empty, blank, or comment-only YAML results in `DeployOpts.SetVariables` being set to nil after unmarshal. We should ensure the map is non-nil after unmarshal if it was non-nil before, allowing the existing `restoreChangedFlags()` function to merge the changed `--set` values as intended.

Here’s a suggested implementation:

```go
hadSetVariables := bundleCfg.DeployOpts.SetVariables != nil

err = unmarshalAndValidateConfig(configFile, &bundleCfg)
if err != nil {
   return err
}

if hadSetVariables && bundleCfg.DeployOpts.SetVariables == nil {
   bundleCfg.DeployOpts.SetVariables = map[string]string{}
}
```

Here is an integration test that we should add and will also prove the regression introduced with this first change:

```go
   func TestConfigurationMergesDeploySetFlagsWithConfigSetVariables(t *testing.T) {
      configPath := filepath.Join(t.TempDir(), "uds-config.yaml")
      contents := []byte(`setVariables:
     from_config: yes
     cli: config
   `)
      if err := os.WriteFile(configPath, contents, 0o600); err != nil {
         t.Fatal(err)
      }
      t.Setenv("UDS_CONFIG", configPath)

      root := NewRootCommand()
      deploy, _, err := root.Find([]string{"deploy"})
      if err != nil {
         t.Fatal(err)
      }
      if err := deploy.Flags().Set("set", "cli=flag"); err != nil {
         t.Fatal(err)
      }
      if err := readViperConfig(); err != nil {
         t.Fatal(err)
      }
      overrides := captureChangedFlags(deploy)
      if err := loadViperConfig(); err != nil {
         t.Fatal(err)
      }
      if err := restoreChangedFlags(overrides); err != nil {
         t.Fatal(err)
      }

      want := map[string]string{
         "from_config": "yes",
         "cli":         "flag",
      }
      for key, value := range want {
         if got := bundleCfg.DeployOpts.SetVariables[key]; got != value {
            t.Fatalf("SetVariables[%q] = %q, want %q", key, got, value)
         }
      }
      if len(bundleCfg.DeployOpts.SetVariables) != len(want) {
         t.Fatalf("SetVariables = %#v, want %#v", bundleCfg.DeployOpts.SetVariables, want)
      }
   }
```

joelmccoy (changes requested):

internal/legacy/cli/root.go
chance-coleman: The original change replaced config-loaded `setVariables` with the parsed CLI map. I replaced it with the targeted nil-map guard and added coverage for config values merged with `--set`, with CLI values taking precedence. The existing empty-config regression test remains to cover the original panic path.

joelmccoy (approved):

PR #1535, opened 2026-09-01
chore: temporary pin of renovate k8s version package to zarf

bradclawsie:
## Description

Renovate wants to promote `k8s` packages to a version incompatible with those used by Zarf. See https://github.com/defenseunicorns/uds-cli/pull/1529 . The version update in these packages causes build failures.

These changes to `renovate.json` will allow renovate to function normally and skip these packages. When Zarf makes changes with respect to these libraries, we will see new build failures due to the same api incompatibility described above (we will be pinned on the old version, rather than Zarf), and we will be able to remove this from `renovate.json`.

I understand this change is not popular with everyone, I welcome discussion.

## Related Issue

Relates to # CLI-325

## Type of change

## Checklist before merging


jasonwashburn:
if merged, title/merge commit should be changed from `fix: ` to `chore: ` . Though we don't have release-please configured here _yet_ , `fix:` type commits end up in the "Bug Fixes" section of the automated changelog rather than something like "Miscellaneous" or "CI"

joelmccoy (commented):
Not a fan of this approach, this config becomes stale and will need to clean it up immediately after the zarf bump. And you will need to add this for EVERY k8s api version that is not compatible with upstream zarf.

Additionally if we leave this config in and we are able to bump these deps without a zarf bump in the future, we lose renovate prs for that and are blind on updates to these deps

bradclawsie:
> Not a fan of this approach, this config becomes stale and will need to clean it up immediately after the zarf bump. And you will need to add this for EVERY k8s api version that is not compatible with upstream zarf.
> 
> Additionally if we leave this config in and we are able to bump these deps without a zarf bump in the future, we lose renovate prs for that and are blind on updates to these deps

Fair points! I'll close it on this. Thanks Joel!

PR #1541, opened 2026-09-03
chore(next): adr about the implementation of uds tools

eoghanriley:
## Description

ADR discussing the future implementation, purpose and usage of `uds tools` in next mode.

## Related Issue

Relates to CLI-162

## Type of change

## Checklist before merging


bradclawsie (commented):
This makes sense to me. `tools` just feels like a superfluous placeholder that aids the implementers instead of the users. I'll be happy to approve once others have had a chance to "vote" on this since it is a big change.

docs/adr/0025-tools-command-structure.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+As of today, the command `uds tools` ships only one tool - Zarf.
+Using Zarf-vendored tools requires a long invocation such as `uds tools zarf tools kubectl`.
+However, Zarf is also exposed as `uds zarf`, which is required when invoking certain tool Zarf vendors, such as kubectl or helm.
```
chance-coleman: Could we update this example? `uds tools zarf tools kubectl` works in Next mode, but the command path is unnecessarily duplicated. Could we describe `uds zarf tools kubectl` as the preferred path instead?

eoghanriley: Good point with clarifying, it seems like the exact command `uds tools zarf tools kubectl` runs but adding on any subcommands like `get pods` causes cobra to error out.

eoghanriley: Clarified it a bit, I tried to convey that point without getting into technical details. Let me know if you think it could be clearer

docs/adr/0025-tools-command-structure.md
```diff
@@ -0,0 +1,56 @@
...
+Date: 2026-09-03
+
+## Status
+
+Accepted
+
+## Context
+
+As of today, the command `uds tools` ships only one tool - Zarf.
+The current recommented path to use Zarf-vendored tools requires a long invocation such as `uds tools zarf tools `.
```
chance-coleman:
```suggestion
The current recommended path to use Zarf-vendored tools requires a long invocation such as `uds tools zarf tools `.
```
nit: typo

docs/adr/0025-tools-command-structure.md
chance-coleman: yeah i think this is good 👍🏼 thanks for updating

docs/adr/0025-tools-command-structure.md
```diff
@@ -0,0 +1,56 @@
...
+As of today, the command `uds tools` ships only one tool - Zarf.
+The current recommended path to use Zarf-vendored tools requires a long invocation such as `uds tools zarf tools `.
+However, Zarf is also exposed as `uds zarf`, which is required when invoking certain commands in tools Zarf vendors, such as kubectl or monitor.
+
+## Decision
+
+Remove `uds tools zarf` in favor of just having `uds zarf`.
+Tools will still exist as a subcommand to expose tools and utilities to improve DevX when interacting with the UDS ecosystem.
+
+Potential examples of tools that would fit this namespace include:
```
joelmccoy: is there going to be a subset of tools we have in this initial implementation? i believe we can start to curate this list from what we have from zarf include some of the examples below.  Also is there criteria for what we should put in uds tools (besides excluding things that aren't in the uds ecosystem)? Like, should we port over all of zarfs current tools? or are we having some criteria for deciding what we should include in the subset of `uds tools`.

Right now zarf has a lot of tools:

```console
  archiver      Compresses/Decompresses generic archives, including Zarf packages
  clear-cache   Clears the configured git and image cache directory
  download-init Downloads the init package for the current Zarf version into the specified directory
  gen-key       Generates a cosign public/private keypair that can be used to sign packages
  gen-pki       Generates a Certificate Authority and PKI chain of trust for the given host
  get-creds     Displays a table of credentials for deployed Zarf services. Pass a service key to get a single credential
  helm          The Helm package manager for Kubernetes.
  kubectl       kubectl controls the Kubernetes cluster manager
  monitor       Launches a terminal UI to monitor the connected cluster using K9s.
  registry      Tools for working with container registries using go-containertools
  sbom          Generates a Software Bill of Materials (SBOM) for the given package
  trusted-root  Tools for working with Sigstore trusted roots
  update-creds  Updates the credentials for deployed Zarf services (deprecated; prefer the per-service subcommands)
  wait-for      Waits for a given Kubernetes resource to be ready
  yq            yq is a lightweight and portable command-line data file processor.
```

Maybe it makes sense  to port all of these? or maybe just a few. I think I lean towards curating this list a bit more and maybe explicitly listing out the things we want in `uds tools` right now

joelmccoy (commented):
Generally like this approach! Have one comment that I would maybe like to see thought through a bit more in this decision

docs/adr/0025-tools-command-structure.md
eoghanriley: I think kubectl and gen-key are the 2 big ones we should definitely port over since they are in our how to guides. In Chance's how-to-guides he has users execute kubectl through `uds zarf tools`. In the PR for signing guides Chance has users generate their signing key through zarf with `uds zarf tools gen-key`. I think helm could and monitor could also be useful. I think it could be cool if we did tofu as well, but that might be harder since their cmd structure is under internal/ so if we went down that rabbit hole it would probably just be invoking the binary.

eoghanriley: Criteria wise I think we should view it as if a tool is needed or provides a large amount of QOL to a user interacting with the UDS ecosystem. Although it should be counterbalanced by the amount of work it can cause to maintain it and keep it stable, and what we can already expect to be on maintenance systems. For example, one could make the argument that off of our current guidance, we shouldn't bundle opentofu as we recommend using the uds provider if they already manage their infrastructure with it. It may be worthwhile to speak with delivery on this to see if we need to include anymore criteria with this for compliance sake outside of our existing criteria, or if this could impose any paperwork headaches on them or mission herors.

joelmccoy: I think it is worth spelling out a couple things in this ADR based on this:

1. Update to list the explicit tools we will support in uds tools (I agree with `gen-key` and `kubectl`, and would maybe consider `yq`)
2. Add some words that we will currate a list of tools to add in `uds tools` based on improvements we expect in UX/Qol, but will not encompass all existing underlying vendored tools

One thing on tofu: We don't currently vendor that in (and I don't think we have plans to at the moment).  That would be another decision altogether if we wanted to vendor that in.  Maybe worth considering down the line

eoghanriley: Updated

eoghanriley:
I bumped the ADR to 26 as there are currently 2 ADRs that are numbered 24 on main.

chance-coleman (approved):

joelmccoy (approved):

PR #1545, opened 2026-09-08
feat(next)!: split remove into 2 commands

eoghanriley:
## Description

This PR splits `uds bundle remove` into two commands, following the split that was done with deploy. `uds bundle remove` handles artifacts while `uds bundle dev remove` handle hcl files and directories.

## Related Issue

Fixes CLI-320

## Type of change

## Checklist before merging


internal/cli/bundle/remove.go
```diff
@@ -149,22 +146,16 @@ func (o *RemoveOptions) Run(ctx context.Context) error {
...
-			return fmt.Errorf("%w %q: %w", ErrParseBundle, o.BundlePath, err)
-		}
-		parsedBundle = inspection.Bundle
-	} else {
-		parsedBundle, err = bundleinternal.NewHCLParser(o.Config.Options.Architecture, s).ParseBundleFile(ctx, bundlePath)
-		if err != nil {
-			return fmt.Errorf("%w %q: %w", ErrParseBundle, bundlePath, err)
-		}
+
+	inspection, err := artifact.InspectBundleDefinition(ctx, artifact.InspectOptions{
```
chance-coleman: This change makes `uds bundle remove` artifact-only, but ADR-0021 still classifies it as a trusted source-based operation. I still get confused here but i think? we write a new ADR to supersede the old one.

tests/testutil/cluster.go
```diff
@@ -428,10 +429,12 @@ func bundleRemoveCommand(ctx context.Context, udsPath, bundlePath string, extraA
 		if info.IsDir() {
 			target = "."
 			workingDir = bundlePath
+			args = append(args, "dev")
```
chance-coleman, line 432: The helper adds dev only when the input is a directory. Passing a direct `bundle.uds.hcl` path leaves the command as `uds bundle remove <file>`, which now rejects HCL input. Current cluster coverage uses a directory, so it does not catch this path. I think if we route a file named `bundle.uds.hcl` through `uds bundle dev remove` we can catch this behavior.

internal/cli/bundle/remove.go
chance-coleman, line 150: This change makes uds bundle remove artifact-only, but ADR 0021 still lists it as a trusted source-based operation. remove now calls `artifact.InspectBundleDefinition` for `.tar.zst` and `OCI` inputs. Should we add an ADR that supersedes this classification and keeps `uds bundle dev remove` as the trusted local-HCL workflow?

chance-coleman (commented):
Looking good, two smaller comments and i'd be happy to approve.

internal/cli/bundle/remove.go
eoghanriley: Is editing ADRs a standard practice? I feel like in this case if it is standard practice it may be fitting since its just updating a command path. If not I can put together an ADR to handle this change.

tests/testutil/cluster.go
eoghanriley: should be fixed now

internal/cli/bundle/remove.go
chance-coleman: Yeah i honestly am not sure, i feel like i've seen us do both previously. I would agree with you that i would typically just prefer to update ADRs with new information if its not a completely different direction.

eoghanriley: Updated ADR 21 to use dev remove

chance-coleman (dismissed):

internal/cli/bundle/dev_remove.go
```diff
@@ -0,0 +1,197 @@
...
+	if _, err := fmt.Fprintln(o.ErrOut(), bundleDefinitionRemoveDiagnostic); err != nil {
+		return fmt.Errorf("%w for bundle definition diagnostic: %w", ErrWriteDefinitionNotice, err)
+	}
+
+	parsedBundle, err := bundleinternal.NewHCLParser(o.Config.Options.Architecture, s).ParseBundleFile(ctx, bundlePath)
+	if err != nil {
+		return fmt.Errorf("%w %q: %w", ErrParseBundle, bundlePath, err)
+	}
+
+	if err = parsedBundle.Validate(); err != nil {
```
joelmccoy: `DevRemoveOptions.Run` now duplicates most of `RemoveOptions.Run` after loading the source-specific bundle. It validates the parsed bundle, checks `--packages`, runs dependency checks, prompts the user, builds `bundle.RemoveOptions`, calls `bundle.Remove`, and prints the result.

Could we extract the shared logic into a helper function that takes the parsed bundle, source path, verification settings, and skip-signature flag? Then, `remove` and `dev remove` would only differ in how they load the bundle and select the verification policy.

Simple cleanup structure:

```go
func runRemove(
   ctx context.Context,
   streams iostreams.IOStreams,
   printer printer.ResourcePrinter,
   cfg *bundle.UDSBundleConfig,
   sourcePath string,
   parsedBundle *spec.UDSBundle,
   packages []string,
   force bool,
   prompt bool,
   verification bundle.VerificationPolicy,
   skipSignatureVerification bool,
) error {
   // shared validate packages, dependency check, prompt, bundle.Remove, print
}
```

Then:
- `RemoveOptions.Run` handles artifact inspection and verification policy, then calls the helper.
- `DevRemoveOptions.Run` prints the bundle-definition warning, parses HCL, and passes an empty policy with `skipSignatureVerification` set to true.

joelmccoy (dismissed):
LGTM - i added a comment to reduce some code duplication, you can take it or leave it

internal/cli/bundle/dev_remove.go
eoghanriley: Extracted

joelmccoy (approved):

PR #1553, opened 2026-09-11
fix(ci): stabilize Zarf registry for legacy bundle E2E tests

bradclawsie:
## Description

The e2e-bundle CI test keeps failing, but then works fine with a manual re-run. This is just annoying. This is an attempt to fix.

After Zarf init, the E2E test now:

  1. Keeps the Zarf registry at one replica.
  2. Sets the registry deployment to one replica.
  3. Waits for the registry to finish starting.

  This keeps the registry from restarting while the test is pushing images.
...

## Related Issue

Relates to CLI-339

## Type of change

## Checklist before merging


bradclawsie:
@greptileai review

bradclawsie:
@codex review

tests/legacy/e2e/main_test.go
```diff
@@ -114,6 +114,35 @@ func deployZarfInit(t *testing.T) {
 		// Deploy
 		runCmd(t, fmt.Sprintf("deploy %s --confirm -l=debug", bundlePath))
 	}
+
+	stabilizeZarfRegistry(t)
```
chance-coleman: Could we pass `--set REGISTRY_HPA_MAX=1` when deploying the init bundle and keep this stabilization only as a one-time fallback for existing init deployments? `deployZarfInit` runs for many tests, so the current unconditional call repeats three Kubernetes commands per test and adds avoidable CI time.

bradclawsie: should be addressed in https://github.com/defenseunicorns/uds-cli/pull/1553/commits/82b4b1ba152fd47495d99b07e10d3610d8b54bf3

tests/legacy/e2e/main_test.go
```diff
@@ -91,7 +93,12 @@ func doAllTheThings(m *testing.M) (int, error) {
 // deployZarfInit deploys Zarf init (from a bundle!) if it hasn't already been deployed.
 func deployZarfInit(t *testing.T) {
 	t.Helper()
-	if !zarfInitDeployed() {
+	zarfInitOnce.Do(func() {
```
chance-coleman: `TestDevDeploy` deletes the Zarf state and webhook, but `zarfInitOnce` remains consumed. So when i run this locally `legacy:e2e`, it runs in one process, later tests skip initialization and fail with secrets "zarf-state" not found. Probably should reset/invalidate this guard when `removeZarfInit` runs, or use a check that allows initialization to run again.

bradclawsie: should be fixed in https://github.com/defenseunicorns/uds-cli/pull/1553/commits/7de39b5ba85c3f2f9412ca25093debc4c495cd23 , but since this PR doesn't even seem to be permanently fixing CI flakiness, I will probably close it. I'll watch what happens with these changes.

bradclawsie:
I was looking for some quick tweaks to CI configs; instead ended up with confusing source changes that didn't address the flakiness

PR #1563, opened 2026-09-17
fix: narrow error check and message to only unsigned bundles

bradclawsie:
## Description

Original issue (CLI-314) requested simplified language for an error message regarding bundle signing.

Previous error message and policy checking merged here: https://github.com/defenseunicorns/uds-cli/pull/1560 was expanded too far, but there is a preference for a much simpler version.

In addition to the verbose messaging, the previous PR had expanded checks on policy state. This PR reduces the check to _only_ check if the bundle is unsigned, and will not associate the new abbreviated message with other policy errors or states.

New checks and flow just to be clear:

`internal/cli/bundle/deploy.go` 179-180 - when `--skip-signature-verification` is absent, in `policyForArtifactDeploy`, get the verification policy, then:

1. Valid key/keyless policy → return it and continue verification.
2. Invalid supplied policy (bad/empty key, incomplete keyless config, etc.) →
     return associated error (not the new abbreviated string).
3. No policy at all → **if unsigned**, return abbreviated message to re-run with `--skip-signature-verification`.

...

## Related Issue

Fixes # CLI-314

## Type of change

## Checklist before merging


internal/cli/bundle/deploy.go
```diff
@@ -129,7 +129,7 @@ func (o *DeployOptions) policyForArtifactDeploy() (bundle.VerificationPolicy, er
 		return policy, err
 	}
 
-	return bundle.VerificationPolicy{}, errors.New("bundle signature verification is enabled but no public key or keyless verification policy is configured; configure one to verify this bundle, or if the bundle is unsigned, re-run with --skip-signature-verification")
+	return bundle.VerificationPolicy{}, errors.New("bundle is not signed, if you wish to deploy this unsigned bundle, re-run with --skip-signature-verification")
```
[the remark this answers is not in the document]
bradclawsie: Sorry, there is a preference for the simplified messaging.

internal/cli/bundle/deploy.go
```diff
@@ -116,20 +119,46 @@ func (o *DeployOptions) Validate() error {
...
 	}
 
-	return bundle.VerificationPolicy{}, errors.New("bundle signature verification is enabled but no public key or keyless verification policy is configured; configure one to verify this bundle, or if the bundle is unsigned, re-run with --skip-signature-verification")
+	isUnsigned := o.isUnsigned
+	if isUnsigned == nil {
+		isUnsigned = artifactIsUnsigned
+	}
+	unsigned, detectErr := isUnsigned(ctx, o.BundlePath, o.Config)
+	if detectErr != nil || !unsigned {
+		return policy, err
```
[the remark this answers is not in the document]
bradclawsie: Intentional: if we can’t determine whether the bundle is unsigned, we keep the existing missing-policy error instead of changing the failure based on a failed signature lookup.

internal/cli/bundle/deploy.go
```diff
@@ -116,20 +119,46 @@ func (o *DeployOptions) Validate() error {
...
 		return policy, err
 	}
 
-	return bundle.VerificationPolicy{}, errors.New("bundle signature verification is enabled but no public key or keyless verification policy is configured; configure one to verify this bundle, or if the bundle is unsigned, re-run with --skip-signature-verification")
+	isUnsigned := o.isUnsigned
+	if isUnsigned == nil {
+		isUnsigned = artifactIsUnsigned
+	}
+	unsigned, detectErr := isUnsigned(ctx, o.BundlePath, o.Config)
+	if detectErr != nil || !unsigned {
```
joelmccoy: I believe we should separate these cases. If `isUnsigned` fails, returning the original missing-policy error conceals the actual artifact or OCI inspection failure. For instance, a corrupt local archive or an unreachable registry would appear to be a verification-policy configuration issue.

Could this be:

```go
if detectErr != nil {
    return policy, fmt.Errorf("checking bundle signature: %w", detectErr)
}
if !unsigned {
    return policy, err
}
```

internal/cli/bundle/deploy.go
```diff
@@ -116,20 +119,46 @@ func (o *DeployOptions) Validate() error {
 		return err
 	}
 	if !o.Verification.SkipSignatureVerification {
-		if _, err := o.policyForArtifactDeploy(); err != nil {
+		policy, err := o.Verification.policy()
+		if err != nil && !isMissingVerificationPolicy(policy, err, o.Verification.PublicKey) {
```
joelmccoy: This helper feels a bit indirect because callers must pass pieces of `VerifyOptions` into it. Since it essentially answers the question, "Did the user provide any verification policy?" could it be placed closer to `VerifyOptions`, perhaps as a method like `o.Verification.isMissingPolicy(policy, err)` or something similar?

This change would make the deployment flow more straightforward:

```go
policy, err := o.Verification.policy()
if err == nil {
    return policy, nil
}
if !o.Verification.isMissingPolicy(policy, err) {
    return policy, err
}
```

joelmccoy (commented):
Functionally good error message flow here. Two small comments as suggestions

internal/cli/bundle/deploy.go
bradclawsie: Agreed. I had previously contradicted the automatic review here that cited the same issue, in retrospect this is more clear.

Change in https://github.com/defenseunicorns/uds-cli/pull/1563/commits/6590749abf350981f891dbda9b5cf2feb50cb612

internal/cli/bundle/deploy.go
bradclawsie: Thank you, this is cleaner. Change made in https://github.com/defenseunicorns/uds-cli/pull/1563/commits/6590749abf350981f891dbda9b5cf2feb50cb612

joelmccoy (approved):

PR #1564, opened 2026-09-17
feat(next): override config vars via cli

eoghanriley:
## Description

This PR adds the `--set` flag to allow users to override config variables at deploy time.

You cannot reference hcl variables declared in defaults such as `--set keycloak_domain=domain` where domain is a variable set in defaults.

## Related Issue

Relates to CLI-315

## Type of change

## Checklist before merging


internal/cli/bundle/config_resolver.go
```diff
@@ -281,6 +281,15 @@ func mergeVariables(base, overrides bundle.Variables) bundle.Variables {
...
 
+func applySetVariables(config *bundle.UDSBundleConfig, entries []string) error {
+	variables, err := bundleinternal.ParseSetVariables(entries)
+	if err != nil {
+		return err
+	}
+	config.Variables = mergeVariables(config.Variables, fromInternalVariables(variables))
+	return nil
+}
+
```
chance-coleman, line 295: When no `--set` flags are provided, this still creates an empty variables map. That causes every values file to be templated and can break existing deployments. I think if we return early when `entries` is empty, that should avoid this edge case.

docs/reference/next/config-uds-hcl.mdx
```diff
@@ -136,9 +136,21 @@ The two configuration categories resolve independently:
...
+
+Both `bundle deploy` and `bundle dev deploy` accept repeatable `--set key=value` flags. Plain values are strings, while booleans, numbers, HCL lists, and HCL objects retain their types. Quote ambiguous values to force strings:
+
+```bash
+CLI_FEATURES=NextMode=true uds bundle dev deploy ./my-bundle \
+  --set domain=example.com \
+  --set replicas=3 \
+  --set enabled=true \
+  --set 'ports=[8080, 8443]' \
+  --set 'release=\"3\"' \
```
chance-coleman: maybe its just my machine 🤔 this line caused the example to not work for me. 
```suggestion
  --set 'release="3"' \
```

chance-coleman (commented):
looking good, couple small comments. [One doc link](https://github.com/defenseunicorns/uds-cli/blob/main/docs/how-to-guides/migrate-legacy-to-next.mdx?plain=1#L195) not in the changed file list that needs to be updated says `--set` doesnt exist on Next.

docs/reference/next/config-uds-hcl.mdx
eoghanriley: Good catch, I meant to put the single quotes around the database var and not release

internal/cli/bundle/config_resolver.go
eoghanriley: Fixed

docs/reference/next/config-uds-hcl.mdx
eoghanriley: Fixed, I ended up just moving the single quotes down to the database var. I don't think that doc has to be all encompassing with the different methods to escape strings, so I didn't add another example in there.

chance-coleman (approved):

slaskawi (approved):

